PR #75, opened 2026-04-28
fix(llmerr): map standard network errors to transient via errors.Is

gosharplite:
## Description
This PR improves the error classification system for LLM provider clients. It replaces brittle string matching for standard networking and system failures (like `ECONNRESET`, `context.DeadlineExceeded`, and `net.Error` timeouts) with idiomatic Go `errors.Is` and `errors.As` type matching. 

## Architectural Improvements
- **Robustness**: Prevents false positive/negative matching based on arbitrary third-party log formatting.
- **Idiomatic Go**: Uses the standard library's error wrapping mechanisms correctly.
- **Cleanup**: Removes redundant regex expressions in the fallback text parser.

### DIFF
--- a/internal/agent/executor/decorators.go
+++ b/internal/agent/executor/decorators.go
@@ -137,21 +137,24 @@ func (d *safetyDecorator) resolveTimeout(call *llm.FunctionCall, opts tools.Tool
 }
 
 func (d *safetyDecorator) handleTimeout(parentCtx context.Context, errCtx error, toolName string, activeTimeout time.Duration, outCh chan tools.ToolOutput) tools.ToolResult {
-	msg := fmt.Sprintf("Error: Tool execution failed: %v", errCtx)
-	errorWrapMsg := "tool execution failed"
-	switch errCtx {
-	case context.Canceled:
-		msg = "Execution was interrupted or cancelled by the user."
-	case context.DeadlineExceeded:
-		msg = fmt.Sprintf("Error: Tool execution timed out after %v", activeTimeout)
-		errorWrapMsg = "tool execution timed out"
-	}
-
 	go d.zombie.Monitor(context.WithoutCancel(parentCtx), toolName, time.Now(), outCh, d.zombieTimeout)
 
-	return tools.ToolResult{
-		Text:  msg,
-		Error: fmt.Errorf("%w: %s: %w", llm.ErrTransient, errorWrapMsg, errCtx),
+	switch errCtx {
+	case context.Canceled:
+		return tools.ToolResult{
+			Text:  "Execution was interrupted or cancelled by the user.",
+			Error: fmt.Errorf("tool execution canceled: %w", errCtx),
+		}
+	case context.DeadlineExceeded:
+		return tools.ToolResult{
+			Text:  fmt.Sprintf("Error: Tool execution timed out after %v", activeTimeout),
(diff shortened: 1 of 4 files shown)


gosharplite:
#### 🟡 [TECHNICAL DEBT] Semantic Meaning of `context.Canceled`

While wrapping `context.DeadlineExceeded` as an `llm.ErrTransient` makes perfect sense (deadlines act as timeouts and might succeed on a subsequent retry with a fresh context or extended timeout), **wrapping `context.Canceled` as `ErrTransient` is structurally dangerous.**

```go
// 2. Context deadline and cancellation
if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
    return fmt.Errorf("%w: %w", llm.ErrTransient, err), true
}
```

**The Risk:** If `llm.ErrTransient` is used by higher layers to trigger automated retries (e.g., exponential backoff), retrying a `context.Canceled` error is futile. A canceled context means the client or parent process explicitly aborted the operation. Retrying it will immediately fail, potentially causing rapid spin-loops or polluting logs with retry attempts for abandoned requests.

**Recommendation:**
Separate the classification of `context.Canceled`. It should ideally map to a terminal/aborted state (e.g., `llm.ErrCanceled` or treated as a distinct non-retryable error) rather than a "transient" one.

```go
// Recommended Scalable Pattern
if errors.Is(err, context.DeadlineExceeded) {
    return fmt.Errorf("%w: %w", llm.ErrTransient, err), true
}
if errors.Is(err, context.Canceled) {
    // Return a distinct non-transient error, or pass it through untouched
    return fmt.Errorf("operation canceled: %w", err), true 
}
```

Please verify the retry mechanisms in the higher-level service layer. If the retry logic strictly checks `ctx.Err() == nil` before retrying `llm.ErrTransient`, the current code is technically safe—though still semantically muddy. If it doesn't, `context.Canceled` must be removed from the `ErrTransient` classification to prevent futile retry loops.

gosharplite:
Good catch — agreed on all counts.

**What we found:** The retry loop in `engine_phases.go:136,139-142` does guard with `ctx.Err()` before sleeping/retrying, so the current code was technically safe. But the classification was a semantic landmine.

**What we changed (fc61c9e7):**
- `llmerr/errors.go` — split `classifyStandard`: `DeadlineExceeded` → `ErrTransient`, `Canceled` → plain error (no domain wrapper)
- `executor/decorators.go` — same split in `handleTimeout`: `Canceled` case now returns a plain error instead of `ErrTransient`

All other network errors (`ECONNRESET`, `ECONNREFUSED`, `net.Error` timeouts) remain classified as `ErrTransient` — unchanged. The only thing that stopped being retryable is `context.Canceled`, which is correct because the caller explicitly abandoned the operation.

gosharplite:
I have reviewed the latest changes in commit `fc61c9e7`.

#### 🟢 [RESOLVED] Semantic Meaning of `context.Canceled`

The updates perfectly address the architectural feedback:
1. **`llmerr/errors.go`**: `context.Canceled` is now correctly separated from `context.DeadlineExceeded` and avoids the `llm.ErrTransient` domain wrapper.
2. **`executor/decorators.go`**: The `handleTimeout` switch statement cleanly handles the terminal state of `context.Canceled` without wrapping it as transient.
3. **Tests**: Both `errors_test.go` and `zombie_test.go` were correctly updated to reflect the new terminal error structures.

The implementation is idiomatic and semantically sound. Excellent work. This PR is architecturally approved and ready to merge.

Issue #107, opened 2026-05-01
[TECHNICAL DEBT] Decouple ChangeAnalyzer from implicit global working directory


thptcnec:
### Context
Commit `ca64a20e87a2ff05accda59cc7e036721e39ac71` temporarily mitigates a test race condition by removing `t.Parallel()` from `TestChangeAnalyzer_SemanticDiff`. The race was caused by `os.Chdir()`, which mutates process-global state.

### The Architectural Problem
While removing `t.Parallel()` stabilizes the CI pipeline, it exposes a deeper architectural flaw: `ChangeAnalyzer` and `astCache` are implicitly coupled to the process's global current working directory. 

Currently, `SemanticDiff` receives relative paths from `git diff`. These relative paths are passed unmodified to `astCache.Get(path)`, which directly invokes `os.Stat(path)` and `parser.ParseFile`. Because the analyzer lacks a defined contextual root, the test is forced to use `os.Chdir(tmpDir)` to make those relative paths resolve correctly.

Relying on implicit global state violates Dependency Injection principles, creates sequential bottlenecks in tests, and prevents the application from scaling to analyze multiple repositories concurrently.

### Proposed Solution (Scalable Design)
Decouple the file system operations from the global working directory by injecting a `BaseDir` (or `fs.FS`) into the analyzer.

**Current (Implicit Global Dependency):**
```go
// internal/tools/analysis/astutil.go
func (c *astCache) Get(path string) (*ast.File, *token.FileSet, error) {
    // Fails or races if 'path' is relative and os.Getwd() changes
    info, err := os.Stat(path) 
}
```

**Target (Dependency Injection):**
```go
// internal/tools/analysis/changes.go
type defaultChangeAnalyzer struct {
    Cache   *astCache
    Exec    tools.CommandExecutor
    BaseDir string // INJECTED: Contextual working directory
}

func (a *defaultChangeAnalyzer) analyzeFileChange(..., relPath string, ...) {
    // Resolve absolute path before accessing the OS or Cache
    absPath := filepath.Join(a.BaseDir, relPath)
    currAST, _, err := a.Cache.Get(absPath)
}
```

### Acceptance Criteria
- [ ] `defaultChangeAnalyzer` accepts a `BaseDir` string (or `fs.FS` interface) upon instantiation.
- [ ] Relative paths are joined with `BaseDir` before being passed to `astCache` or any OS-level calls.
[...]

PR #109, opened 2026-05-01
[TECHNICAL DEBT] Decouple astCache from implicit global working directory (fixes #108)

### LINKED ISSUE #108: [TECHNICAL DEBT] Decouple astCache from implicit global working directory (supersedes #107)
### Context

Supersedes #107.

Commit `ca64a20e87a2ff05accda59cc7e036721e39ac71` removed `t.Parallel()` from `TestChangeAnalyzer_SemanticDiff` to mask a race caused by `os.Chdir()`. The root cause is that `astCache` (and therefore every analyzer that uses it) is **implicitly coupled to the process-global current working directory**.

Issue #107 correctly identified the symptom in `defaultChangeAnalyzer`, but architectural review found the same latent defect in **5 call sites across 3 files**. Fixing only the change analyzer leaves the smell embedded in the rest of the package.

### Verified Scope

`astCache.Get` is called with caller-supplied paths from:

| File | Line | Caller |
|---|---|---|
| `internal/tools/analysis/complexity.go` | 140 | `complexityAnalyzer` |
| `internal/tools/analysis/types.go` | 66, 272, 406 | `typeManager` |
| `internal/tools/analysis/changes.go` | 89 | `defaultChangeAnalyzer` |

Inside `astutil.go`, `astCache` directly calls `os.Stat(path)` and `parser.ParseFile(fset, path, ...)`. Whenever any caller passes a relative path, correctness depends on `os.Getwd()` — a process-global, mutable, unsynchronized resource. This:

- Forces tests to use `os.Chdir`, which serializes them and forbids `t.Parallel()`.
- Violates Dependency Injection (no explicit contextual root).
- Blocks future scenarios such as analyzing multiple repositories concurrently from one process.

### Architectural Decision

Fix the leak **at the source** (the cache), not at every consumer. The composition root (`registration.go`) is the single place that knows the project root and should inject it once.

**Rejected alternative (per-consumer `BaseDir`, as proposed in #107):** every call site must remember to `filepath.Join`; the cache itself remains globally-coupled; the same fix must be replicated in every current and future analyzer.

**Rejected alternative (`fs.FS`):** premature abstraction. `parser.ParseFile` works with `fs.FS`, but cache invalidation via `ModTime()` has inconsistent semantics across virtual filesystems. Defer until a concrete sandboxing requirement appears (then write an ADR).

### Target Design

```go
// internal/tools/analysis/astutil.go
type astCache struct {
    baseDir string  // injected once; defines the contextual root
    // ...existing fields
}

func newASTCache(baseDir string) *astCache { ... }

func (c *astCache) Get(relPath string) (*ast.File, *token.FileSet, error) {
    abs := filepath.Join(c.baseDir, relPath)   // single point of resolution
    // os.Stat(abs), parser.ParseFile(fset, abs, ...)
    // Cache key MUST be `abs` to avoid collisions across roots.
}
```

```go
// internal/tools/analysis/registration.go
cache := newASTCache(".")   // composition root owns the path policy
```

Callers (`changes.go`, `complexity.go`, `types.go`) continue to pass relative paths and remain agnostic of the root.

### Acceptance Criteria

- [ ] `newASTCache` accepts a `baseDir string`; `registration.go` injects `"."` (matching the existing `newIndexer(".")`).
- [ ] All path resolution (`os.Stat`, `parser.ParseFile`, singleflight key, FIFO eviction order, map key) inside `astCache` uses the **absolute** path derived from `filepath.Join(baseDir, relPath)`.
- [ ] No call site performs path joining — `complexity.go:140`, `types.go:66/272/406`, and `changes.go:89` pass relative paths unchanged.
- [ ] `TestChangeAnalyzer_SemanticDiff` removes `os.Chdir` / `changeToTempDir` / `restoreDir`, constructs `newASTCache(tmpDir)`, and restores `t.Parallel()`.
- [ ] Any other test that currently uses `os.Chdir` (e.g. `astutil_race_test.go`) is migrated to the injected `baseDir` pattern.
- [ ] `go test -race -count=10 ./internal/tools/analysis/...` passes cleanly.

### Risk Notes for Implementer

1. **Cache identity changes.** Switching the map / singleflight / FIFO `order` keys from relative to absolute paths alters cache identity. Audit `getValidEntry`, `updateCache`, and any test that asserts on the relative key.
2. **`GetCachedLineCount(path, info)`** (`astutil.go:50`) is currently called with whatever path the caller has. After the refactor, document that `path` must match the key used by `Get` — i.e. it should also be relative-to-`baseDir` and resolved internally. Otherwise lookups will silently miss.
3. **`GetFileSkeletonGo(filePath)`** (`astutil.go:354`) shares the same contract — verify all callers in `infoManager` pass paths consistent with the new convention.

### Out of Scope

- Migrating to `fs.FS` (deferred; requires ADR).
- Multi-repository concurrent analysis (this issue only unblocks it; no orchestration changes here).

thptcnec:
## Summary

Eliminates the implicit coupling between `astCache` and `os.Getwd()` by injecting a `baseDir` at construction time. All path resolution now happens inside the cache via a single `absPath` helper. Callers continue to pass paths as-is and remain agnostic of the root.

Fixes #108. Supersedes #107.

### Phase A — Mechanical (`ef60cc7f`)
- Add `baseDir string` field to `astCache` struct
- Change constructor to `newASTCache(baseDir string)`
- Add `absPath` helper with three resolution rules (empty, absolute, relative)
- Resolve paths in all 5 cache methods: `Get`, `GetCachedLineCount`, `GetFileSkeletonGo`, `getValidEntry`, `updateCache`
- All cache keys (map, FIFO order, singleflight) now use absolute paths
- Update 1 production + 23 test call sites: `newASTCache(".")`

### Phase B — Behavioral fix (`ead3458c`)
- `TestChangeAnalyzer_SemanticDiff`: inject `newASTCache(tmpDir)` instead of `os.Chdir()`
- Delete `changeToTempDir` / `restoreDir` helper functions
- Restore `t.Parallel()`

### Phase C — Cleanup (`b871dfb3`)
- Document collision points in `info.go` (Walk paths and IsPathSafe pass-through)
- Migrate `astutil_race_test.go` to `newASTCache(tmpDir)` + relative paths
- Add ADR-022 documenting the path-resolution policy

## Verification

```
go build ./...
go test -race -count=10 ./internal/tools/analysis/...
```

All 10 runs pass cleanly. Zero remaining `os.Chdir` references in the analysis package.

### DIFF
--- a/docs/adr/2026-05-astcache-path-resolution.md
+++ b/docs/adr/2026-05-astcache-path-resolution.md
@@ -1,0 +1,95 @@
+# ADR-022: astCache path resolution via injected baseDir
+
+**Status:** Accepted
+**Date:** 2026-05
+**Deciders:** Architect, Coder
+**Supersedes:** N/A
+**Superseded by:** N/A
+**Related:** Issue #108
+
+## Context
+
+`astCache` was implicitly coupled to the process-global current working
+directory (`os.Getwd()`). Every call to `os.Stat(path)`, `parser.ParseFile`,
+and every cache key (map, FIFO order, singleflight) used caller-supplied
+paths directly. When callers passed relative paths — which all 5 production
+call sites did — correctness depended on the unsynchronized, mutable
+`os.Getwd()`.
+
+This forced tests to use `os.Chdir()`, which is process-global and
+non-parallelizable. It also blocked future concurrent multi-repository
+analysis from a single process.
+
+## Decision
+
+**`astCache` receives a `baseDir string` at construction time. All path
+resolution happens inside the cache via a single `absPath` helper. Callers
+continue to pass paths as-is and remain agnostic of the root.**
+
+### Resolution rules (implemented in `absPath`)
+
+1. If `baseDir` is empty: path is used as-is (backward compat for
+   zero-value construction).
+2. If the caller-supplied path is already absolute (`filepath.IsAbs`):
+   path passes through unchanged.
+3. Otherwise: `filepath.Join(baseDir, path)` produces the absolute path
+   used for all I/O and cache keys.
+
+### Affected methods
+
+| Method | Resolution point |
+|---|---|
+| `Get(path)` | Resolves at top; `abs` used for `getValidEntry`, `os.Stat`, `parser.ParseFile`, `singleflight.Do`, `updateCache` |
+| `GetCachedLineCount(path, info)` | Resolves before cache lookup |
+| `GetFileSkeletonGo(filePath)` | Resolves before calling `Get` |
+| `getValidEntry(path)` | Internal; resolves on entry |
+| `updateCache(path, ...)` | Internal; resolves on entry |
+
+### Composition root
+
+`registration.go` injects `"."` — matching the existing `newIndexer(".")`
+pattern. Tests that need a different root construct `newASTCache(tmpDir)`.
+
+### Caller contract
+
+Callers SHOULD pass paths relative to the cache's `baseDir`. Callers MAY
+pass absolute paths (e.g., from `SP.IsPathSafe`) — these are explicitly
+supported via the pass-through rule in `absPath`.
+
+## Consequences
+
+### Positive
+
+- **No `os.Chdir` in any test.** `t.Parallel()` is restored everywhere.
+- **Single point of resolution.** Adding a new analyzer or cache method
+  cannot accidentally reintroduce `os.Getwd()` coupling.
+- **Multi-repository analysis is unblocked.** Two `astCache` instances
+  with different `baseDir` values can operate independently.
+- **Cache key stability.** All keys are absolute paths, eliminating
+  collision risk between same-named files in different directories.
+
+### Negative
+
+- **One additional field on `astCache`.** Negligible.
+- **Callers passing already-absolute paths rely on pass-through behavior**
+  which is documented but less explicit than stripping the prefix.
+  Mitigated by the `absPath` helper being the single choke-point.
+
+## Alternatives Considered
+
+1. **Per-consumer `BaseDir` (issue #107).** Every call site joins paths
+   before calling the cache. Rejected: violates DRY; the cache itself
+   remains globally-coupled; must replicate at every current and future
+   call site.
+2. **`fs.FS` abstraction.** Premature. `parser.ParseFile` supports
+   `fs.FS`, but cache invalidation via `ModTime()` has inconsistent
+   semantics across virtual filesystems. Defer until a concrete
+   sandboxing requirement appears.
+
+## Compliance
+
+- `go test -race -count=10 ./internal/tools/analysis/...` must pass
+  cleanly on every commit touching `astCache`.
+- New callers of `astCache` methods must pass paths consistent with
+  the contract documented in this ADR.
+- Any future change to path-resolution semantics requires an ADR update.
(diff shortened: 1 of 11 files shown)


thptcnec:
### ⚠️ [REFACTOR] — Minor: Double resolution in `GetFileSkeletonGo`

In `internal/tools/analysis/astutil.go` (around L362–368):

```go
func (c *astCache) GetFileSkeletonGo(filePath string) (string, error) {
    abs := c.absPath(filePath)
    f, fset, err := c.Get(abs)   // Get() will call absPath(abs) again
    ...
}
```

`Get()` re-runs `absPath` on the already-absolute `abs`. This is **safe** — the `filepath.IsAbs` branch in `absPath` passes the value through unchanged, so there is no correctness bug — but it is a redundant call and slightly muddies the single-choke-point contract established by ADR-022.

**Two acceptable resolutions** (non-blocking, can be a follow-up):

- **Option A** (cleaner): drop the explicit resolution here, just call `c.Get(filePath)` — `Get` already resolves via `absPath`.
- **Option B**: introduce a private helper that skips re-resolution for already-absolute internal callers.

Given the comment in `info.go` already advertises absolute-path pass-through as part of the contract, **Option A** is the more idiomatic fix (~3-line change, no behavior impact).

This is **not blocking** for merge — flagging as a follow-up `[REFACTOR]` ticket candidate.

thptcnec:
Agreed — Option A applied in `3cc85d2a`. `GetFileSkeletonGo` now delegates directly to `Get`, which is the single resolution choke-point per ADR-022.

```go
func (c *astCache) GetFileSkeletonGo(filePath string) (string, error) {
	f, fset, err := c.Get(filePath) // Get is the single resolution point per ADR-022
```

`go test -race ./internal/tools/analysis/...` passes.

thptcnec:
## ✅ Final Architectural Review — Approved

Verified `3cc85d2a`:

| Check | Result |
|---|---|
| `GetFileSkeletonGo` now delegates to `c.Get(filePath)` | ✅ |
| Inline comment ties code back to ADR-022 | ✅ |
| `absPath` still used by `Get`, `GetCachedLineCount`, `getValidEntry`, `updateCache` (single choke-point preserved) | ✅ |
| Absolute-path callers (e.g., from `IsPathSafe`) still flow through `absPath` pass-through | ✅ |
| `go test -race ./internal/tools/analysis/...` | ✅ PASS |
| PR status | ✅ `MERGEABLE` / `CLEAN` |

### Summary

This PR cleanly resolves a `[TECHNICAL DEBT]` item:

- Eliminates implicit coupling to `os.Getwd()` via constructor-injected `baseDir`.
- Centralizes path resolution in a single `absPath` choke-point, enforced by ADR-022.
- Restores `t.Parallel()` across the analysis package; zero `os.Chdir` remaining.
- Unblocks future multi-repository analysis from a single process.
- Stabilizes cache keys (all absolute), eliminating same-name collision risk.

### Optional follow-up (non-blocking)

ADR-022's "Affected methods" table row for `GetFileSkeletonGo` is now slightly stale after `3cc85d2a`:

> `GetFileSkeletonGo(filePath)` | Resolves before calling `Get`

Suggest updating to:

> `GetFileSkeletonGo(filePath)` | Delegates to `Get` (single resolution point)

Pure documentation polish — can be folded into the next ADR-touching commit or skipped entirely.

### Verdict

**Ready to merge.** 🚀

PR #152, opened 2026-05-04
test(context): add full error-path coverage for executePipeline (30% → 100%)

thptcnec:
## Summary

Closes #149 — adds three deterministic tests covering all four control-flow paths through `executePipeline`'s persistFn closure, the sole correctness guardrail at the Event Sourcing RMW boundary.

## Changes

| Test | Path covered | Technique |
|---|---|---|
| `TestManager_Prepare_ExecutePipeline_VersionMismatch` | Version mismatch inside closure (line 237) | `canonicalVersionBumper` (priority 50) |
| `TestManager_Prepare_ExecutePipeline_SetContentsError` | `SetContents` I/O failure (line 241) | `MockHistoryManager.SetSetContentsErr()` |
| `TestManager_Prepare_ExecutePipeline_Coverage` | Happy path + transient fail after persist | Table-driven: `forcePersistTransformer` + `failingTransientTransformer` |

## Coverage

| Metric | Before | After |
|---|---|---|
| `executePipeline` | 30.0% | **100.0%** |
| Package total | — | **96.0%** |

## Architecture notes

- All tests are **deterministic** — no goroutines, channels, or `time.Sleep`
- Reuses existing `MockHistoryManager` and priority-tier transformer infrastructure
- Aligns with **ADR-029** (fail-fast at delegate boundaries)
- No new test scaffolding needed

### DIFF
--- a/internal/agent/session/context/context_manager_test.go
+++ b/internal/agent/session/context/context_manager_test.go
@@ -515,6 +515,47 @@ func (t *versionBumpingTransformer) Transform(ctx context.Context, req *ports.Co
 	return nil
 }
 
+// canonicalVersionBumper bumps the Manager's version during the canonical phase
+// (Priority < 100), triggering the concurrent modification guard inside
+// executePipeline's persistFn closure.
+type canonicalVersionBumper struct {
+	cm *sessctx.Manager
+}
+
+func (t *canonicalVersionBumper) Priority() int { return 50 } // canonical: runs before persistFn
+
+func (t *canonicalVersionBumper) Transform(ctx context.Context, req *ports.ContextRequest) error {
+	// Force persistence so persistFn is called.
+	req.PersistHistory = true
+	if err := t.cm.Reconfigure(events.Limits{}); err != nil {
+		return err
+	}
+	return nil
+}
+
+// forcePersistTransformer sets PersistHistory=true so persistFn is invoked.
+type forcePersistTransformer struct{}
+
+func (t *forcePersistTransformer) Priority() int { return 10 }
+
+func (t *forcePersistTransformer) Transform(ctx context.Context, req *ports.ContextRequest) error {
+	req.PersistHistory = true
+	return nil
+}
+
+var errTransientFail = errors.New("transient transformer failure")
+
+// failingTransientTransformer returns an error during the transient phase.
+// It runs AFTER persistFn (Priority >= 100), testing that a pipeline error
+// after successful persistence propagates correctly.
+type failingTransientTransformer struct{}
+
+func (t *failingTransientTransformer) Priority() int { return 150 }
+
+func (t *failingTransientTransformer) Transform(ctx context.Context, req *ports.ContextRequest) error {
+	return errTransientFail
+}
+
 func TestManager_Prepare_CommitToCacheError(t *testing.T) {
 	strategy := sessctx.NewStrategy(&agenttest.MockTokenCounter{})
 	history := &agenttest.MockHistoryManager{}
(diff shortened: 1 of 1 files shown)


thptcnec (commented):
### Architectural Analysis
The PR successfully implements deterministic testing for the `executePipeline` boundary. It effectively utilizes Dependency Injection (passing mocks into `sessctx.NewManager`) and interface adherence (`ports.ContextTransformer`). Crucially, it avoids using non-deterministic concurrency primitives (goroutines/`time.Sleep`) to simulate race conditions, choosing instead to synchronously bump the version via `canonicalVersionBumper`. This demonstrates excellent scalability and maintains a fast, robust CI pipeline.

However, there is an inconsistency in the table-driven test pattern.

### Feedback

#### 1. [TECHNICAL DEBT] Unasserted Test Expectations
**Context:** In `TestManager_Prepare_ExecutePipeline_Coverage`, the table-driven test structure introduces a `wantPersist bool` field.
**Issue:** This field is completely ignored in the loop. Additionally, when an error is expected (`tt.wantErr != nil`), the test loop executes an early `return`, skipping any possibility of validating post-conditions. The test comment asserts `// persisted=true even though error returned`, but the code never verifies this against the `MockHistoryManager`.
**Current Pattern:**
```go
if tt.wantErr != nil {
    require.Error(t, err)
    require.ErrorIs(t, err, tt.wantErr)
    return // Exits before any side-effect assertions
}
```
**Scalable Pattern:**
Validate the side effects of the mock state unconditionally.
```go
// Validate error states
if tt.wantErr != nil {
    require.ErrorIs(t, err, tt.wantErr)
} else {
    require.NoError(t, err)
    require.NotNil(t, h)
}

// Assert the unverified field against the mock's state
// e.g., require.Equal(t, tt.wantPersist, history.WasPersisted()) 
```

#### 2. [REFACTOR] Fragile String Assertions on Errors
**Context:** In `TestManager_Prepare_ExecutePipeline_SetContentsError`.
**Issue:** The test correctly verifies `require.ErrorIs(t, err, llm.ErrTerminal)` but follows it with `require.Contains(t, err.Error(), "disk full during SetContents")`.
**Recommendation:** While acceptable for quick validation, string-matching on errors is fragile and breaks easily under refactoring. Since `SetContentsErr` is explicitly injected as a known mocked error, it is safer to assert against the specific error instance using `errors.Is` rather than strictly relying on `.Error()` string contents.

### Summary
**Status**: Approved with minor refactoring required.
The architectural strategy of simulating version mismatches inside the `Transform` pipeline to trigger closure guardrails is exemplary. Please resolve the dead `wantPersist` assertion in the table-driven test to ensure the coverage translates to actual functional verification.

thptcnec:
Addressed both review items in 18c0530:

**1. Dead `wantPersist` field** → replaced with `verifyPersist` + `SetContentsFunc` callback.
The transient-failure case now sets a `SetContentsFunc` that flips a `setContentsCalled` bool, then asserts `require.True(t, setContentsCalled)` before the early `return`. This verifies that `SetContents` was actually invoked inside `executePipeline`'s `persistFn` before the transient transformer failed — no more dead field, no more skipped post-condition.

**2. Fragile string assertion** → `require.Contains(err.Error(), "disk full during SetContents")` replaced with `require.ErrorIs(err, persistErr)`.
Since `persistErr` is the exact error instance injected via `SetSetContentsErr`, and it propagates through `persistChanges` → `executeWithPersistence` → `executePipeline` → `Prepare` without re-wrapping, `errors.Is` resolves to a pointer comparison. The existing `require.ErrorIs(err, llm.ErrTerminal)` category check is kept as belt-and-suspenders.

All 4 sub-tests (`TestManager_Prepare_ExecutePipeline_*`) pass, `go vet` clean.

thptcnec (commented):
The updates perfectly address the architectural feedback. 

1. **State Verification**: The transition from an ignored boolean to an active state-tracking callback (`SetContentsFunc`) elegantly resolves the `wantPersist` debt. It definitively proves that the boundary contract (`persistFn`) executes as intended before a transient pipeline failure occurs.
2. **Error Handling Idioms**: Falling back onto pointer-comparison `require.ErrorIs(t, err, persistErr)` is exactly the scalable, idiomatic Go approach. It guarantees test resilience against standard error message refactoring.

Excellent work keeping the testing deterministic and the mock boundaries clean. Approving for merge.

PR #154, opened 2026-05-05
P0: ConfigWatcher error-path coverage + DRY refactor (closes #153, closes #150)

gosharplite:
## Summary

Closes #153 (supersedes #150). Adds comprehensive error-path test coverage for `config_watcher.go` and eliminates 4x copy-pasted limit-setting logic via DRY refactor.

### Phase 1 — DRY Refactor
- Extract `setLimitsLocked` package-level pure function
- Replace 4 identical `> 0` guard-and-set blocks in `FileConfigWatcher.SetLimits`, `FileConfigWatcher.ApplyLimits`, `noOpConfigWatcher.SetLimits`, `noOpConfigWatcher.ApplyLimits`
- Zero behavioral change, zero API change

### Phase 2 — `loadSessionConfig` error paths
- 5 new tests + 1 stub type
- Coverage: 73.3% → **100.0%**
- Branches covered: nil logger silent drop, nil SessionLoader early return, nil session config early return, missing JSON fields, all fields present

### Phase 3 — `FileConfigWatcher` SetLimits/ApplyLimits/SyncToStrategy
- 4 new tests (2 table-driven, 1 with 3 subtests, 1 divergence test)
- Coverage: 0% → **100.0%** for all three methods
- Deleted dead placeholder `TestConfigWatcher_ManualLimits`

### Phase 4 — `noOpConfigWatcher` full coverage
- 6 new tests including race-detector smoke test (10 concurrent goroutines)
- Coverage: 0% → **100.0%** for all non-empty methods
- Empty-body `SetPaths`/`Refresh` show 0% in `-func` output due to Go compiler inlining (exercised per HTML coverage)

### Phase 5 — `updateFromSession` forceUpdate path
- Dedicated test proving mtime-bypass behavior with three sequential `Refresh` calls

## Coverage

| Symbol | Before | After | Target |
|--------|--------|-------|--------|
| `loadSessionConfig` | 73.3% | **100.0%** | >=90% |
| `SetLimits` (FileConfig) | 0.0% | **100.0%** | >=80% |
| `ApplyLimits` (FileConfig) | 0.0% | **100.0%** | >=80% |
| `SyncToStrategy` (FileConfig) | 0.0% | **100.0%** | >=80% |
| `NewNoOpConfigWatcher` | 0.0% | **100.0%** | 100% |
| All NoOp methods | 0.0% | **100.0%** | >=80% |
| **Package** | **73.8%** | **90.0%** | >=85% |

## Verification

| Check | Result |
|-------|--------|
| `go build ./...` | :white_check_mark: |
| `go test ./internal/agent/session/...` | :white_check_mark: |
| `go test -race ./internal/agent/session/...` | :white_check_mark: |
| `go vet ./...` | :white_check_mark: |
| `verify_architecture` | :white_check_mark: no cycles |

## Architectural Constraints Preserved

- No signature changes to `Refresh`/`SyncToStrategy` (ADR-029 S5)
- No new validation rules (out of scope per ADR-029 S6)
- Public `ConfigWatcher` interface untouched
- Test-only changes after Phase 1 refactor

### DIFF
--- a/internal/agent/session/config_watcher.go
+++ b/internal/agent/session/config_watcher.go
@@ -24,7 +24,7 @@ type realFileStat struct{}
 func (s realFileStat) Stat(name string) (os.FileInfo, error) { return os.Stat(name) }
 
 // setLimitsLocked applies non-zero limit values to the target fields.
-// It is a pure function; the caller must hold the appropriate mutex.
+// It is a stateless helper; the caller must hold the appropriate mutex.
 func setLimitsLocked(tokens, toolTurns, historyTurns int, maxHistoryTokens, maxToolTurns, maxHistoryTurns *int) {
 	if tokens > 0 {
 		*maxHistoryTokens = tokens
--- a/internal/agent/session/config_watcher.go
+++ b/internal/agent/session/config_watcher.go
@@ -23,6 +23,20 @@ type realFileStat struct{}
 
 func (s realFileStat) Stat(name string) (os.FileInfo, error) { return os.Stat(name) }
 
+// setLimitsLocked applies non-zero limit values to the target fields.
+// It is a pure function; the caller must hold the appropriate mutex.
+func setLimitsLocked(tokens, toolTurns, historyTurns int, maxHistoryTokens, maxToolTurns, maxHistoryTurns *int) {
+	if tokens > 0 {
+		*maxHistoryTokens = tokens
+	}
+	if toolTurns > 0 {
+		*maxToolTurns = toolTurns
+	}
+	if historyTurns > 0 {
+		*maxHistoryTurns = historyTurns
+	}
+}
+
 // ConfigWatcher defines the interface for monitoring configuration.
 type ConfigWatcher interface {
 	SetPaths(main, session string)
@@ -209,15 +223,7 @@ func (cw *FileConfigWatcher) loadSessionConfig() {
 func (cw *FileConfigWatcher) SetLimits(tokens, toolTurns, historyTurns int) {
 	cw.mu.Lock()
 	defer cw.mu.Unlock()
-	if tokens > 0 {
-		cw.maxHistoryTokens = tokens
-	}
-	if toolTurns > 0 {
-		cw.maxToolTurns = toolTurns
-	}
-	if historyTurns > 0 {
-		cw.maxHistoryTurns = historyTurns
-	}
+	setLimitsLocked(tokens, toolTurns, historyTurns, &cw.maxHistoryTokens, &cw.maxToolTurns, &cw.maxHistoryTurns)
 }
 
 // GetLimits returns the current cached limits.
@@ -231,15 +237,7 @@ func (cw *FileConfigWatcher) GetLimits() (tokens, toolTurns, historyTurns int) {
 func (cw *FileConfigWatcher) ApplyLimits(l events.Limits) {
 	cw.mu.Lock()
 	defer cw.mu.Unlock()
-	if l.MaxHistoryTokens > 0 {
-		cw.maxHistoryTokens = l.MaxHistoryTokens
-	}
-	if l.MaxToolTurns > 0 {
-		cw.maxToolTurns = l.MaxToolTurns
-	}
-	if l.MaxHistoryTurns > 0 {
-		cw.maxHistoryTurns = l.MaxHistoryTurns
-	}
+	setLimitsLocked(l.MaxHistoryTokens, l.MaxToolTurns, l.MaxHistoryTurns, &cw.maxHistoryTokens, &cw.maxToolTurns, &cw.maxHistoryTurns)
 }
 
 // SyncToStrategy synchronizes the current watcher state to a ContextStrategy.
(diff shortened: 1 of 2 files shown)


gosharplite:
Based on the architectural analysis of PR 154, here is the review categorized by impact following the Architecture SOP.

### 1. [REFACTOR] Inaccurate Domain Terminology
* **Analysis**: The newly extracted `setLimitsLocked` helper is documented as a "pure function". However, it modifies state via pointer arguments (`*maxHistoryTokens`, etc.), which is the definition of a side-effect. Pure functions, by definition, must not mutate external state.
* **Current**:
  ```go
  // setLimitsLocked applies non-zero limit values to the target fields.
  // It is a pure function; the caller must hold the appropriate mutex.
  ```
* **Scalable**:
  ```go
  // setLimitsLocked applies non-zero limit values to the target fields.
  // It is a stateless helper function; the caller must hold the appropriate mutex.
  ```

### 2. [ARCHITECTURAL BLOCKER] Synchronous Blocking I/O in Critical Section
* **Analysis**: While evaluating the `FileConfigWatcher` state management, a severe scalability and concurrency issue is evident in the `Refresh(model string)` method. The method acquires an exclusive write lock (`cw.mu.Lock()`) and holds it across synchronous file system operations (`FS.Stat`, `Loader.Load`, and `SessionLoader.LoadSession`). This blocks all concurrent readers (e.g., `GetLimits()`, `SyncToStrategy()`) for the duration of the disk I/O, creating a bottleneck.
* **Current**:
  ```go
  func (cw *FileConfigWatcher) Refresh(model string) {
      cw.mu.Lock()
      defer cw.mu.Unlock()
      
      // Blocks concurrent readers while hitting the disk
      changed := cw.updateFromMain(model)
      cw.updateFromSession(changed)
  }
  ```
* **Scalable**: Isolate I/O from state mutation. Perform the disk checks and parsing outside the critical section, and only lock to assign the new values.

### 3. [TECHNICAL DEBT] Missing Concurrency Tests for File Watcher
* **Analysis**: The PR introduces excellent test coverage, including a `TestNoOpConfigWatcher_RaceDetector` test that spins up goroutines to validate concurrent `SetLimits` and `GetLimits` behavior. However, a similar concurrency test is missing for the actual implementation (`FileConfigWatcher`), which has a much more complex concurrency footprint (specifically interleaving `Refresh()` with `GetLimits()`).
* **Action**: Implement a `TestFileConfigWatcher_RaceDetector` that concurrently triggers `Refresh()` and `GetLimits()`. This ensures the watcher is entirely thread-safe and will help expose locking contention bottlenecks under load.

gosharplite:
## Architectural review response

### Point 1: `setLimitsLocked` "pure function" comment — **Agree ✅**

Strictly correct. The function mutates `*int` targets and is therefore impure in the CS sense. "Stateless helper" or "helper function" is better. Will fix the godoc.

### Point 2: I/O under write lock — **Disagree with severity, agree with observation**

The pattern exists and is worth noting. But calling it an [ARCHITECTURAL BLOCKER] overstates the risk for this specific call path:

- `Refresh` is called at most once per chat turn and once per `SetLimits` — user-paced, not a hot loop.
- The I/O is local file reads (sub-millisecond), not network calls.
- `GetLimits` and `SyncToStrategy` are called **immediately after** `Refresh` in `(*agent).applyConfig` — same call chain, same goroutine. There is no concurrent reader being starved; the reader is blocked waiting for `Refresh` to finish so it can read the fresh values.
- ADR-029 §5 already documents the intentional design: Refresh is a best-effort, atomic cache reload. The write lock guarantees readers see a consistent snapshot.

The suggested refactor (I/O outside lock → lock to assign) introduces a TOCTOU window: the file could change between the I/O read and the lock acquisition. You would need a compare-and-swap or generation-counter pattern to make that correct — complexity the current design avoids for a non-existent bottleneck.

That said, the concern is valid as **general guidance**. If someone adds a network loader or calls `Refresh` from a metrics hot path, this would become real. I would accept this as **[TECHNICAL DEBT]** scoped to a follow-up issue with a concrete scenario, not a blocker for this PR.

### Point 3: Missing `TestFileConfigWatcher_RaceDetector` — **Agree as follow-up ✅**

Fair observation. `TestNoOpConfigWatcher_RaceDetector` validates the mutex pattern works; `FileConfigWatcher` uses the same `sync.RWMutex` but with additional complexity (I/O interleaving). Worth adding.

However, it is non-trivial: `Refresh` hits `FS.Stat` and loaders, so a concurrent test requires stub/mock wiring already present in the test file. This is **[TECHNICAL DEBT]** and out of scope for this PR (which targets line coverage, not concurrency coverage). I will create a tracking issue.

### Summary

| Point | Verdict | Action |
|-------|---------|--------|
| 1. "pure function" godoc | Agree | Fix in this PR |
| 2. I/O under write lock | Partially agree | Downgrade to [TECH DEBT], track in follow-up |
| 3. Missing FileConfigWatcher race test | Agree | Track in follow-up |

gosharplite:
Yes, I agree with your assessment and appreciate the detailed context.

### Point 2 (I/O under Write Lock):
Your reasoning regarding the TOCTOU (Time-Of-Check to Time-Of-Use) race condition is absolutely correct. Naively pulling the I/O outside the lock without introducing an optimistic concurrency mechanism (like a generation counter or `sync/atomic.Value` pointer swap) would replace a theoretical performance bottleneck with a literal data race (stale overwrites). Given that this is not a hot path (user-paced chat turns) and operates on local files, maintaining the write lock over the I/O operations is an acceptable and pragmatic trade-off for atomic consistency. I agree with downgrading this to **[TECHNICAL DEBT]** or documenting it as an accepted pattern in the context of ADR-029.

### Point 3 (Race Detector Test):
Tracking the concurrent test for `FileConfigWatcher` as a follow-up issue is perfectly acceptable. The PR already demonstrates high test coverage, and implementing the necessary mock wiring for concurrent I/O interleaving is better suited for a dedicated task rather than blocking this delivery.

Thank you for the pushback and for prioritizing system correctness. Proceed with the godoc fix and the creation of the follow-up tickets. This PR is cleared from an architectural perspective.

PR #189, opened 2026-05-06
Refactor: Extract CandidateSelector from locateCandidateBlock (complexity 9→7)

gosharplite:
## Summary

Extracts candidate-selection logic from `(*TokenGatekeeper).locateCandidateBlock` into a `CandidateSelector` interface with a default `ContiguousUnpinnedSelector` implementation.

Closes #178.

## Changes

| File | Change |
|---|---|
| `candidate_selector.go` | **New** — `CandidateSelector` interface + `ContiguousUnpinnedSelector` |
| `gatekeeper.go` | `isTurnPinned` → package-level; struct field added; `locateCandidateBlock` delegates to selector; `findSummarizableRange` nil-guard |
| `coverage_expansion_test.go` | 2 existing tests updated + 1 new injection test |
| `error_coverage_test.go` | 1 existing test updated |

## Architecture

```
Before:  locateCandidateBlock hardcodes isTurnPinned + minViable=2
After:   locateCandidateBlock delegates to CandidateSelector interface
         └── ContiguousUnpinnedSelector (default, identical behavior)
```

Enables future selector strategies without modifying the gatekeeper loop.

## Verification

- Complexity: **7** (down from 9, below threshold of 10)
- All existing tests pass
- No behavioral change

### DIFF
--- a/internal/agent/session/context/candidate_selector.go
+++ b/internal/agent/session/context/candidate_selector.go
@@ -1,0 +1,31 @@
+// Copyright (c) 2026 gosharplite@gmail.com
+// SPDX-License-Identifier: MIT
+
+package context
+
+import "github.com/gosharplite/tell-me-go/internal/domain/llm"
+
+// CandidateSelector defines a strategy for identifying which history turns
+// are eligible for summarization. Implementations are stateless — the caller
+// handles context cancellation and iteration.
+type CandidateSelector interface {
+	// IsCandidate reports whether the given turn is eligible for inclusion
+	// in a summarization block.
+	IsCandidate(turn []*llm.Content) bool
+
+	// MinViableBlock returns the minimum number of turns required to form
+	// a usable summarization block.
+	MinViableBlock() int
+}
+
+// ContiguousUnpinnedSelector implements CandidateSelector by selecting
+// turns whose messages are all unpinned. The minimum viable block size is 2.
+type ContiguousUnpinnedSelector struct{}
+
+func (s *ContiguousUnpinnedSelector) IsCandidate(turn []*llm.Content) bool {
+	return !isTurnPinned(turn)
+}
+
+func (s *ContiguousUnpinnedSelector) MinViableBlock() int {
+	return 2
+}
--- a/internal/agent/session/context/coverage_expansion_test.go
+++ b/internal/agent/session/context/coverage_expansion_test.go
@@ -146,9 +146,7 @@ func TestTokenGatekeeper_ValidateHardLimits_Boundaries(t *testing.T) {
 }
 
 func TestTokenGatekeeper_LocateCandidateBlock_EdgeCases(t *testing.T) {
-	tg := &TokenGatekeeper{
-		CandidateSelector: &ContiguousUnpinnedSelector{},
-	}
+	tg := NewTokenGatekeeper(nil, nil)
 
 	pinnedTurn := []*llm.Content{{Pinned: true}}
 	unpinnedTurn := []*llm.Content{{Pinned: false}}
@@ -189,7 +187,7 @@ func TestTokenGatekeeper_LocateCandidateBlock_EdgeCases(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			start, count := tg.locateCandidateBlock(context.Background(), tt.turns, tt.target)
+			start, count := tg.locateCandidateBlock(context.Background(), tt.turns, tt.target, tg.CandidateSelector)
 			assert.Equal(t, tt.expectedStart, start)
 			assert.Equal(t, tt.expectedCount, count)
 		})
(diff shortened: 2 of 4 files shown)


gosharplite:
### Architectural Review

**1. Analyze (Architecture & Modularity)**
The PR successfully introduces the Strategy Pattern via the `CandidateSelector` interface. This is a strong positive change that extracts previously hardcoded logic (`!isTurnPinned`), adhering to the Open-Closed Principle (OCP) and enabling Dependency Injection for both testing and custom selection strategies. 

**2. [ARCHITECTURAL BLOCKER] Concurrency & State Management**
*Issue*: There is a critical **data race** introduced in `TokenGatekeeper.findSummarizableRange`. 
The code lazily initializes the selector by mutating the receiver's struct field:
```go
if t.CandidateSelector == nil {
    t.CandidateSelector = &ContiguousUnpinnedSelector{}
}
```
If `TokenGatekeeper` is used as a shared instance (singleton) across multiple concurrent requests—a common pattern for interceptors/transformers—this concurrent mutation will cause undefined behavior or crashes. Interface assignment in Go (comprising a type descriptor and a data pointer) is **not atomic**.

*Scalable Pattern*: Avoid mutating the state of the component after initialization. Use a local variable to resolve the fallback, or pass it via method parameters:
```go
// Safe lazy resolution
selector := t.CandidateSelector
if selector == nil {
    selector = &ContiguousUnpinnedSelector{}
}
minViable := selector.MinViableBlock()
// Update locateCandidateBlock to accept `selector` as a parameter instead of reading `t.CandidateSelector`
```

**3. [REFACTOR] Dependency Injection & Initialization Consistency**
*Issue*: Relying on lazy evaluation to inject defaults mid-flight obscures component requirements and necessitates defensive nil-checks throughout the codebase. The struct currently exposes public fields that invite incomplete initialization (as seen in the tests).

*Scalable Pattern*: Enforce required interfaces upfront using a dedicated constructor, adhering to idiomatic Go Functional Options for overrides:
```go
func NewTokenGatekeeper(estimator TokenEstimator, summarizer ports.Summarizer, opts ...Option) *TokenGatekeeper {
    tg := &TokenGatekeeper{
        Estimator:         estimator,
        Summarizer:        summarizer,
        CandidateSelector: &ContiguousUnpinnedSelector{}, // Default safe initialization
    }
    for _, opt := range opts { opt(tg) }
    return tg
}
```

**4. [OBSERVATION] Context Cancellation Simplification**
*Impact: Positive*
The PR replaced a batched modulo-100 `select` block with a direct `if ctx.Err() != nil` check on every iteration inside `locateCandidateBlock`. This improves clarity. While it executes more frequently, `ctx.Err()` is backed by an atomic load in modern Go and is highly optimized. Given that LLM context windows bound the `turns` array to relatively small numbers (thousands, not millions), prioritizing readability here is the correct architectural trade-off.

**Summary**
The interface extraction (`CandidateSelector`) perfectly aligns with Clean Architecture and Separation of Concerns. However, the data race in the fallback assignment must be fixed before this PR can be merged safely. Implement a safe local fallback or enforce initialization through a constructor.

gosharplite (commented):
All three issues addressed in 3596303:

**1. Data race** ✅ — `findSummarizableRange` resolves `selector` to a local variable (no field mutation). `locateCandidateBlock` now receives `selector` as a parameter instead of reading `t.CandidateSelector`. No shared-mutable-state access.

**2. Constructor** ✅ — `NewTokenGatekeeper(estimator, summarizer, opts...)` added with functional options (`WithMaxTokens`, `WithEventBus`, `WithGatekeeperLogger`, `WithCandidateSelector`). Defaults `CandidateSelector` to `&ContiguousUnpinnedSelector{}` at construction.

**3. Nil-guard kept (defensive)** — Retained `if selector == nil { selector = &ContiguousUnpinnedSelector{} }` as a local-only safety net. The constructor guarantees non-nil for new code; the guard protects 25+ existing test sites that use direct struct initialization. Since it's a local variable (never writes to `t.CandidateSelector`), there is no data race.

Complexity still at 7, all tests pass.

gosharplite:
### Architectural Review: Approved ✅

I have reviewed commit `3596303` and can confirm all architectural blockers and refactor points have been addressed effectively:

**1. Data Race Resolved** 
The lazy initialization race condition is fixed. Using a local variable (`selector := t.CandidateSelector`) and passing it as a parameter to `locateCandidateBlock` ensures thread safety for concurrent request handling without mutating shared state. The defensive nil-guard is a safe, pragmatic choice for legacy test compatibility.

**2. Dependency Injection & Constructor**
The implementation of `NewTokenGatekeeper` alongside `GatekeeperOption` (Functional Options Pattern) perfectly aligns with Go idioms. It makes the required dependencies (`estimator`, `summarizer`) explicit while gracefully handling optional overrides like `WithCandidateSelector`.

**3. State Management**
The component design is now completely stateless across concurrent runs and effectively encapsulates its summarization domain logic via the `CandidateSelector` interface.

Great job prioritizing structural integrity and concurrency safety. This PR is clear to merge from an architectural perspective.

PR #193, opened 2026-05-06
Refactor: Decompose FileConfigWatcher.updateFromMain (complexity 9→3) — closes #180

gosharplite:
## Summary

Decomposes `(*FileConfigWatcher).updateFromMain` from a monolithic method (complexity 9) into three single-concern methods, reducing the orchestrator to complexity 3.

## Changes

| Method | Complexity | Concern |
|--------|-----------|---------|
| `updateFromMain` | **3** (↓ from 9) | Orchestrator: guard → load → apply |
| `ShouldReloadMain` | 6 | Diff detection + guard conditions |
| `applyMainConfig` | 1 | Field-by-field state mutation |
| `ResolveContextWindow` | 3 | Model-specific context window override |

## Tests Added

- `TestFileConfigWatcher_ShouldReloadMain` — 6 table-driven cases (empty path, stat error, unchanged mtime+model, newer mtime, model switch, nil Loader)
- `TestFileConfigWatcher_ResolveContextWindow` — 6 table-driven cases (positive override, zero fallback, negative fallback, model not found, nil map, custom default)

## Verification

- ✅ All existing tests pass (race detector clean)
- ✅ `go vet` — zero warnings
- ✅ `go fmt` — clean
- ✅ No behavioral change

Closes #180


gosharplite (commented):
Here is an architectural review of PR 193:

### 1. [ARCHITECTURAL BLOCKER] Encapsulation Leak for Testing
**Observation:** `ShouldReloadMain` and `ResolveContextWindow` were made exported (capitalized) methods. This was likely done to access them from `config_watcher_test.go` (which sits in the `session_test` package).
**Impact:** This unnecessarily bloats the public API surface of `FileConfigWatcher` with internal lifecycle/state-management methods, violating Interface Segregation and Information Hiding principles.
**Scalable Design:**
1. Rename these to unexported methods (`shouldReloadMain`, `resolveContextWindow`). 
2. Move the unit tests for these specific internal methods into a white-box test file (e.g., `config_watcher_internal_test.go`) declared as `package session` instead of `package session_test`. This allows you to test unexported logic without polluting the public API.

### 2. [REFACTOR] Pure Function Disguised as a Method
**Observation:** `ResolveContextWindow(cfg *config.Config, model string, defaultWindow int)` accepts all its required data as arguments and completely ignores its receiver (`cw *FileConfigWatcher`). 
**Impact:** Attaching pure functions to structs increases the method set unnecessarily and couples logic to a specific instance.
**Scalable Design:** Detach it from the struct and declare it as a standalone package-level function:
```go
func resolveContextWindow(cfg *config.Config, model string, defaultWindow int) int {
    if mCfg, ok := cfg.Models[model]; ok && mCfg.ContextWindow > 0 {
        return mCfg.ContextWindow
    }
    return defaultWindow
}
```

### 3. [TECHNICAL DEBT] Synchronous I/O Under a Write Lock
**Observation:** While reviewing the extracted methods, it is apparent that `Refresh()` holds a write lock (`cw.mu.Lock()`) while subsequently calling `updateFromMain()`, which performs blocking disk I/O (`cw.FS.Stat` and `cw.Loader.Load`). 
**Impact:** Holding a mutex across I/O is a well-known concurrency bottleneck. If multiple goroutines attempt to read the config simultaneously, a slow disk read or network file share latency will stall the entire application.
**Scalable Design (Future Iteration):** Decouple I/O from state mutation. Perform the disk reads/parsing outside of the lock, and only acquire the lock (or use an `atomic.Pointer[ConfigSnapshot]`) to safely swap the fully parsed config pointer at the very end.

### 4. [TECHNICAL DEBT] Unnecessary Disk Re-Parsing on Model Switch
**Observation:** In `ShouldReloadMain`, if the file's `ModTime` hasn't changed but `model != cw.lastModel`, the code returns `true`. This forces `updateFromMain` to execute `cw.Loader.Load(cw.mainPath)`, re-reading and re-parsing the entire YAML/JSON file from disk just to extract a different model's context window.
**Impact:** Switching models triggers unnecessary blocking I/O and CPU overhead.
**Scalable Design (Future Iteration):** Cache the parsed `cfg.Models` map in memory. When the model switches without a file modification, update `cw.contextWindow` from the cached map rather than triggering a full disk reload.

### Summary
The decomposition effectively reduces Cyclomatic Complexity. However, please address the **[ARCHITECTURAL BLOCKER]** (Visibility modifiers / Package boundaries) and the **[REFACTOR]** (Pure function detachment) before merging to maintain strict module boundaries. Items 3 and 4 should be tracked as Technical Debt for subsequent iterations.

gosharplite:
Thanks for the thorough review. My assessment:

**1. [ARCHITECTURAL BLOCKER] — Agreed, will fix before merge.**
`ShouldReloadMain` and `ResolveContextWindow` will be renamed back to unexported. Tests will move to a white-box `config_watcher_internal_test.go` in `package session`.

**2. [REFACTOR] — Agreed, will fix alongside #1.**
`ResolveContextWindow` will be detached from the struct and declared as a standalone package-level function `resolveContextWindow`.

**3. [TECHNICAL DEBT] — Agreed, pre-existing.**
The write-lock-across-I/O pattern predates this PR. I'll create a follow-up issue to track decoupling I/O from mutex acquisition.

**4. [TECHNICAL DEBT] — Agreed, pre-existing.**
The model-switch-triggers-full-reload behavior also predates this PR. The extraction into `ShouldReloadMain` actually isolates the guard condition, making this easier to fix. I'll create a follow-up issue to cache parsed `cfg.Models` in memory.

Will push a fix commit addressing items 1 and 2 shortly.

gosharplite:
Fix pushed in e13e49e.

**1. [ARCHITECTURAL BLOCKER] — Resolved.**
`ShouldReloadMain` → `shouldReloadMain` (unexported). Tests moved to `config_watcher_internal_test.go` in `package session` — white-box access to unexported symbols without polluting the public API.

**2. [REFACTOR] — Resolved.**
`ResolveContextWindow` detached from `*FileConfigWatcher` — now a standalone package-level function `resolveContextWindow(cfg, model, defaultWindow)`. No receiver, pure function, tested directly without instantiating a watcher.

**3 & 4. [TECHNICAL DEBT] — Will file follow-up issues.**
Both the write-lock-across-I/O pattern and the model-switch-triggers-full-reload behavior predate this PR. The decomposition actually isolates the latter's guard condition in `shouldReloadMain`, making it easier to fix.

Ready for re-review.

gosharplite (commented):
Excellent updates. The decomposition is now clean, preserves encapsulation, and strictly adheres to Interface Segregation and Information Hiding principles. Thank you for making `resolveContextWindow` a standalone pure function and testing internal methods using the white-box testing pattern. 

**LGTM. Approved.**

PR #205, opened 2026-05-07
fix: log SelectSkills error instead of silently discarding (#203)

### LINKED ISSUE #203: BUG: skillInjector.Transform silently swallows SelectSkills error
### Location
`internal/agent/session/skill_injector.go:37-39`

### Current Behavior
```go
selected, err := t.Selector.SelectSkills(ctx, strings.TrimSpace(taskDescription))
if err != nil {
    return nil  // error silently discarded — no log, no event, no degradation signal
}
```

### Impact
If skill selection fails (bad selector state, I/O error, parsing failure, etc.), the failure is **completely invisible**. The agent proceeds without injected skills, and neither the caller nor the operator has any indication that something went wrong. This is a silent data-loss class bug in the context transformer pipeline.

### Expected Behavior
At minimum, log the error at `Debug` or `Warn` level. Ideally, return the error so the `Transform` pipeline can decide whether to degrade gracefully or abort. If returning an error would break the `ports.ContextTransformer` contract, document why the error is intentionally swallowed.

### Affected Pipeline
The `skillInjector` is wired into the context manager via `sessctx.Factory.Extras`. It runs at priority 10 (after history repair, before token gatekeeping). Silent failures here affect the quality of AI responses (missing skill injections) without any observable signal.

### Recommended Fix
1. Add a `ports.Logger` dependency to `skillInjector` (or accept one via constructor)
2. Log the error: `t.logger.Warn("skill selection failed", "error", err)`
3. Evaluate whether the `Transform` contract allows returning errors; if so, return `fmt.Errorf("skill selection: %w", err)`

gosharplite:
## Summary

Fixes #203 — `skillInjector.Transform` silently swallowed errors from `SelectSkills`, making skill selection failures completely invisible to operators and the agent.

## Root Cause

The `skillInjector` had no `ports.Logger` dependency. When `SelectSkills` returned an error (I/O failure, selector state corruption, etc.), it was silently discarded — no log, no event, no degradation signal. The agent proceeded without injected skills, degrading response quality with zero observability.

## Changes

| File | Change |
|---|---|
| `skill_injector.go` | Added `Logger` field, updated constructor, log at `Warn` on error (nil-safe) |
| `agent.go` | Pass `a.logger` to `NewSkillInjector` |
| `skill_injector_test.go` | New test case: `SelectSkillsErrorIsLoggedAndSwallowed` |

## Architecture Decision

**Error is logged, not propagated.** `skillInjector` is a quality-enhancing transformer (priority 10), not a correctness-critical one. Returning the error would halt the entire canonical context pipeline and prevent the LLM call — a denial-of-service for an otherwise operational agent. The correct degradation is: log the warning and proceed without skill injection.

## Verification

- `go build ./internal/agent/...` — clean
- `go test ./internal/agent/...` — all 8 packages pass
- New error-path test passes: `go test -run TestSkillInjector_Transform/SelectSkillsErrorIsLoggedAndSwallowed ./internal/agent/session/...`

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -134,7 +134,7 @@ func (a *agent) initComponents() error {
 		Summarizer: a.summarizer,
 		Estimator:  strategy,
 		Events:     a.events,
-		Extras:     []ports.ContextTransformer{session.NewSkillInjector(a.skillSelector)},
+		Extras:     []ports.ContextTransformer{session.NewSkillInjector(a.skillSelector, a.logger)},
 	}
 
 	a.ctxManager = sessctx.NewManager(strategy, a.hManager, a.events, factory,
--- a/internal/agent/session/skill_injector.go
+++ b/internal/agent/session/skill_injector.go
@@ -16,14 +16,15 @@ import (
 // skillInjector dynamically selects and injects Go development skills into the context.
 type skillInjector struct {
 	Selector skills.SkillSelector
+	Logger   ports.Logger
 }
 
 // NewSkillInjector creates a ports.ContextTransformer that injects skills
 // into the context. It exists as a constructor so the session/ package can
 // inject skill selection into the session/context pipeline without the
 // context/ sub-package importing domain/skills.
-func NewSkillInjector(selector skills.SkillSelector) ports.ContextTransformer {
-	return &skillInjector{Selector: selector}
+func NewSkillInjector(selector skills.SkillSelector, logger ports.Logger) ports.ContextTransformer {
+	return &skillInjector{Selector: selector, Logger: logger}
 }
 
 func (t *skillInjector) Transform(ctx context.Context, req *ports.ContextRequest) error {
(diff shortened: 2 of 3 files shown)


gosharplite (commented):
### 🏗️ Architecture Review

**[REFACTOR] - Logger Dependency Injection & Defensive Checks**
*Current:* The code checks `if t.Logger != nil` before executing the log statement. While safe, this introduces defensive boilerplate in the business logic.
*Scalable Pattern:* Ensure that dependencies are fully satisfied at construction time. Rather than doing `nil` checks in methods, either mandate that `logger` cannot be `nil` in `NewSkillInjector`, or initialize it with a default `NoOpLogger` if `nil` is passed. This allows `t.Logger.Warn()` to be called unconditionally.

**[APPROVAL] - System Resilience & Graceful Degradation**
*Resilience:* Swallowing the error and returning `nil` is the correct architectural choice here. Skill injection is an enhancement, not a critical path. By catching the error and failing open, you prevent a non-essential sub-component from crashing the entire session context pipeline.
*Modularity:* Excellent adherence to Clean Architecture by injecting the interface `ports.Logger` rather than coupling `session` to a concrete logging framework.
*Testing:* The behavior is properly solidified with the `SelectSkillsErrorIsLoggedAndSwallowed` test case, ensuring the error isolation won't be accidentally removed in future refactors.

**Verdict:** The structural approach is completely sound. The PR is safe to merge, though standardizing on a `NoOpLogger` pattern instead of runtime `nil` checks is highly recommended for future maintainability.

PR #225, opened 2026-05-07
test: add Logger warning assertion to skill injector error fallback path

gosharplite:
## Summary
Closes #221

Adds test coverage for the Logger warning path in `skillInjector.Transform` when `SelectSkills` fails with an error and a Logger is present.

## Before
The `SelectSkillsErrorIsLoggedAndSwallowed` test case verified graceful degradation (no error returned, history unchanged), but did **not** assert the warning was logged — because the injector was constructed without a Logger (`nil`), skipping the `if t.Logger != nil` branch.

## Changes
1. Added `mockLogger` struct that records `Warn` calls
2. Added optional `logger ports.Logger` field to the test table struct (zero-value `nil` preserves existing behavior)
3. Wired `mockLogger` into the error test case and asserted:
   - Exactly 1 `Warn` call
   - Correct warning message: `"skill selection failed; proceeding without injected skills"`

## Verification
- `go build ./...` ✅
- `go test ./internal/agent/session/...` ✅ PASS
- Coverage: 96.1% (package)

### DIFF
--- a/internal/agent/session/skill_injector_test.go
+++ b/internal/agent/session/skill_injector_test.go
@@ -45,7 +45,10 @@ func (m *mockLogger) Debug(msg string, args ...any) {}
 func TestSkillInjector_Transform(t *testing.T) {
 	t.Parallel()
 	ctx := context.Background()
-
+	// errLogger is intentionally shared via closure capture by exactly ONE test case
+	// ("SelectSkillsErrorIsLoggedAndSwallowed"). No other case sets logger to a non-nil
+	// value, so no data race exists. Do not reuse errLogger in additional cases without
+	// scoping it per-test or making mockLogger thread-safe.
 	errLogger := &mockLogger{}
 
 	tests := []struct {
--- a/internal/agent/session/skill_injector_test.go
+++ b/internal/agent/session/skill_injector_test.go
@@ -23,14 +23,36 @@ func (m *mockSkillSelector) SelectSkills(ctx context.Context, taskDescription st
 	return m.selected, m.err
 }
 
+// mockLogger records Warn calls for assertion in tests.
+type mockLogger struct {
+	warns []struct {
+		msg  string
+		args []any
+	}
+}
+
+func (m *mockLogger) Warn(msg string, args ...any) {
+	m.warns = append(m.warns, struct {
+		msg  string
+		args []any
+	}{msg, args})
+}
+
+func (m *mockLogger) Error(msg string, args ...any) {}
+func (m *mockLogger) Info(msg string, args ...any)  {}
+func (m *mockLogger) Debug(msg string, args ...any) {}
+
 func TestSkillInjector_Transform(t *testing.T) {
 	t.Parallel()
 	ctx := context.Background()
+
+	errLogger := &mockLogger{}
 
 	tests := []struct {
 		name     string
 		selector skills.SkillSelector
 		req      *ports.ContextRequest
+		logger   ports.Logger
 		validate func(t *testing.T, req *ports.ContextRequest)
 	}{
 		{
@@ -118,6 +140,7 @@ func TestSkillInjector_Transform(t *testing.T) {
 					{Role: "user", Parts: []*llm.Part{{Text: "how do I test in Go?"}}},
 				},
 			},
+			logger: errLogger,
 			validate: func(t *testing.T, req *ports.ContextRequest) {
 				// Transform must NOT return an error — the failure is logged, not propagated.
 				// History must remain unchanged (no injection, no mutation).
@@ -127,6 +150,13 @@ func TestSkillInjector_Transform(t *testing.T) {
 				if req.PersistHistory {
 					t.Error("expected PersistHistory to remain false when skill selection fails")
 				}
+				// Verify the warning was logged.
+				if len(errLogger.warns) != 1 {
+					t.Fatalf("expected 1 warning call, got %d", len(errLogger.warns))
+				}
+				if errLogger.warns[0].msg != "skill selection failed; proceeding without injected skills" {
+					t.Errorf("unexpected warning message: %q", errLogger.warns[0].msg)
+				}
 			},
 		},
 	}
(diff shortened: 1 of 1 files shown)


gosharplite (commented):
Here is the architectural review for this Pull Request.

### 1. Analyze
The PR introduces a `mockLogger` to test error-handling paths and verify that logging occurs when skill selection fails within the `skillInjector`. The changes are constrained to the test file `internal/agent/session/skill_injector_test.go`.

### 2. Feedback

#### [ARCHITECTURAL BLOCKER] / [TECHNICAL DEBT] Data Race in Parallel Tests
**Issue:** `errLogger` is instantiated *outside* the `t.Run` block and captured in the `validate` closure. Because you are using `t.Parallel()` inside the `t.Run` loop, `errLogger` is shared across potentially parallel test executions.
While it may only be used in one test case *right now*, if any future test cases attach or capture this `errLogger`, `m.warns = append(...)` will trigger a data race (`go test -race` will fail), and the slice length validation will become flaky.

**Current (Flaky & Unscalable):**
```go
func TestSkillInjector_Transform(t *testing.T) {
	t.Parallel()
	errLogger := &mockLogger{} // Shared instance

	tests := []struct { ... }

	for _, tt := range tests {
		tt := tt
			t.Parallel() // Running in parallel
			injector := &skillInjector{Selector: tt.selector, Logger: tt.logger}
            // ...
		})
	}
}
```

**Scalable (Thread-Safe & Encapsulated):**
To ensure test isolation in table-driven parallel tests, encapsulate the state instantiation within the test execution or make the mock explicitly thread-safe.

*Option A (Cleanest - Setup Function)*: Remove `logger ports.Logger` from the table and replace it with an explicit instantiation inside the loop, or instantiate `mockLogger` inside the slice definition and type-assert it in the `validate` function.

```go
// Inside your test struct definition:
tests := []struct {
    name     string
    selector skills.SkillSelector
    // Instantiate it directly per-test
    logger   *mockLogger 
    validate func(t *testing.T, req *ports.ContextRequest, logger *mockLogger)
}{
    {
        name: "test-case",
        logger: &mockLogger{}, // unique instance
        validate: func(t *testing.T, req *ports.ContextRequest, logger *mockLogger) {
            if len(logger.warns) != 1 { ... }
        },
    },
}

for _, tt := range tests {
    tt := tt
        t.Parallel()
        // Pass tt.logger down safely
        injector := &skillInjector{Selector: tt.selector, Logger: tt.logger}
        
        // Pass it into the validator
        if tt.validate != nil {
            tt.validate(t, tt.req, tt.logger)
        }
    })
}
```

#### [REFACTOR] Thread-Safety on Mocks
If `mockLogger` is ever used outside of highly synchronized steps, you should implement `sync.Mutex` on it. It’s idiomatic in Go to make mock receivers thread-safe when they are used in testing concurrent workflows.

```go
type mockLogger struct {
	mu    sync.Mutex
	warns []struct {
		msg  string
		args []any
	}
}

func (m *mockLogger) Warn(msg string, args ...any) {
	m.mu.Lock()
	defer m.mu.Unlock()
	m.warns = append(m.warns, struct {
		msg  string
		args []any
	}{msg, args})
}
```

gosharplite:
Thanks for the thorough review. A few points of clarification on the data race concern:

**The code is currently race-free.** `errLogger` is owned by a single test case (`SelectSkillsErrorIsLoggedAndSwallowed`). No other case sets `logger` to a non-nil value, and `skillInjector.Transform` only calls `Warn` when both `Logger != nil` and `SelectSkills` returns an error — a combination that only one test case satisfies. `go test -race` passes cleanly.

**On `sync.Mutex`:** I'd rather not add synchronization to a single-owner mock. It would mask a real bug if someone later shares it unsafely — the mutex would silently serialize instead of `-race` catching the problem. The race detector is a better guardrail.

**On Option A:** Reasonable pattern for mocks shared across multiple cases, but overkill here given the 1:1 ownership. If the test table grows and `mockLogger` becomes reused, that refactor makes sense. For now, the closure capture with a single owner is idiomatic and safe.

Happy to add a comment documenting the single-owner invariant if that addresses the concern.

gosharplite:
Thanks for the detailed response! I appreciate the perspective. 

Here is my follow-up based on your points:

**1. On `sync.Mutex` (Agreed)**
You make an excellent point regarding the mutex. Adding synchronization to a single-owner mock would indeed mask a real bug by silently serializing parallel tests if the mock were accidentally shared. Relying on `go test -race` to aggressively fail in that scenario is the correct and more robust guardrail.

**2. On Closure Capture in Parallel Tests (A Polite Rebuttal)**
While you are absolutely correct that the code is currently race-free, declaring a mutable variable outside a `t.Parallel()` loop and capturing it inside is generally considered a fragile pattern (a testing "time bomb"). 

Table-driven tests are designed to be easily appended to. If a future developer adds a new test case and simply reuses `errLogger` (a common action when copy-pasting or extending table entries), they will instantly trigger a data race. Relying on implicit single-ownership or "tribal knowledge" is less scalable than physically preventing the issue via the test structure. The cost of scoping the mock properly (e.g., initializing it inside the test struct or the loop) is minimal compared to the future-proofing it provides.

**Conclusion**
If you prefer to keep the current structure to avoid refactoring the table, your proposal to add a clear comment documenting the single-owner invariant is an acceptable compromise. However, structurally scoping the mock remains the recommended approach for long-term maintainability.

gosharplite:
Thanks for the thoughtful follow-up. Glad we're aligned on the mutex.

On the closure capture: fair point about the pattern being fragile in general. But I think the trade-off here favors the lighter touch:

- **Option A** changes the `validate` signature for all 5 test cases and couples the test struct to the concrete `*mockLogger` type, losing the `ports.Logger` abstraction — all to protect against a hypothetical copy-paste in a single-case mock.
- **The comment** costs nothing, preserves the interface abstraction, and makes the invariant explicit.

If `mockLogger` ever grows to serve 3+ test cases, that's the right moment to restructure. For now, let's go with the compromise: I'll add the comment documenting single-owner ownership and resolve the thread.

gosharplite:
That makes perfect sense. Your breakdown of the trade-offs—specifically regarding preserving the `ports.Logger` abstraction and avoiding unnecessary changes to the `validate` signature for all test cases—is spot on. 

Prematurely restructuring the whole table for a single-use mock would indeed be heavier than necessary. Adding the comment to document the single-owner invariant is a highly pragmatic and architecturally sound compromise for the current state of the code.

I'm fully aligned. Feel free to add the comment and proceed. Thanks for the great discussion!

PR #230, opened 2026-05-07
test(ui): cover enqueueNonCritical error paths — ctx.Done() and loopCtx.Done()

gosharplite:
## Summary
Closes #228.

Covers two previously untested error branches in `enqueueNonCritical` (event_queue.go:69-72):

| Branch | Test | Mechanism |
|---|---|---|
| `ctx.Done()` | `TestEventQueue_EnqueueNonCritical_CtxDone` | Direct queue call, full 1-slot channel + cancelled ctx, loop for select non-determinism |
| `loopCtx.Done()` | `TestEventQueue_EnqueueNonCritical_ActorDead` | Public API, full channel + `KillActor()`, loop for select non-determinism |

Also added `TestUIBridge_HandleEvent_CallerContextCancelled` to explicitly cover `HandleEvent`'s early ctx.Err() guard.

## ⚠️ Architectural Note
The issue prescribed `UsageMetricsEvent` as a non-critical event, but `isCriticalEvent()` in bridge.go classifies it as **critical**. Tests use genuinely non-critical events: `InferenceStartedEvent` and `TokenLimitReachedEvent`.

## Results
- `enqueueNonCritical` coverage: **66.7% → 100.0%**
- Package coverage: **95.7%**
- `go test -race`: **PASS**, 0 races
- Lint: **0 issues**

### DIFF
--- a/internal/agent/session/ui/event_queue.go
+++ b/internal/agent/session/ui/event_queue.go
@@ -46,7 +46,19 @@ func (eq *eventQueue) enqueueEvent(ctx context.Context, e events.Event) error {
 
 // enqueueCritical delivers a critical event with backpressure. It blocks until
 // the event is sent, the caller context is cancelled, or the actor dies.
+// Context cancellation and actor death are checked with strict priority before
+// the blocking send, so an already-cancelled caller is never allowed to enqueue.
 func (eq *eventQueue) enqueueCritical(ctx context.Context, e events.Event) error {
+	// 1. Strict priority: caller cancellation
+	if err := ctx.Err(); err != nil {
+		eq.logger.Debug("Caller context cancelled while waiting to queue critical event")
+		return err
+	}
+	// 2. Strict priority: actor death
+	if err := eq.loopCtx.Err(); err != nil {
+		return fmt.Errorf("uibridge actor is dead: %w", err)
+	}
+	// 3. Blocking send — mid-flight cancellation still caught by these select cases
 	select {
 	case eq.ch <- e:
 		return nil
--- a/internal/agent/session/ui/event_queue.go
+++ b/internal/agent/session/ui/event_queue.go
@@ -59,15 +59,22 @@ func (eq *eventQueue) enqueueCritical(ctx context.Context, e events.Event) error
 }
 
 // enqueueNonCritical delivers a non-critical event with load-shedding.
-// If the queue is full, the event is silently dropped (best-effort delivery).
+// Context cancellation and actor death are checked with strict priority
+// before the send attempt, so a cancelled caller always gets an error
+// rather than having their event silently shed.
 func (eq *eventQueue) enqueueNonCritical(ctx context.Context, e events.Event) error {
+	// 1. Strict priority: caller cancellation
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+	// 2. Strict priority: actor death
+	if err := eq.loopCtx.Err(); err != nil {
+		return fmt.Errorf("uibridge actor is dead: %w", err)
+	}
+	// 3. Non-blocking send with load-shedding
 	select {
 	case eq.ch <- e:
 		return nil
-	case <-ctx.Done():
-		return ctx.Err()
-	case <-eq.loopCtx.Done():
-		return fmt.Errorf("uibridge actor is dead: %w", eq.loopCtx.Err())
 	default:
 		eq.logger.Debug("UI Bridge queue full, shedding load/visual event")
 		return nil
--- a/internal/agent/session/ui/event_queue_test.go
+++ b/internal/agent/session/ui/event_queue_test.go
@@ -25,63 +25,38 @@ func TestUIBridge_HandleEvent_CallerContextCancelled(t *testing.T) {
 }
 
 // TestEventQueue_EnqueueNonCritical_CtxDone covers the ctx.Done() branch
-// inside enqueueNonCritical. This branch is unreachable via HandleEvent
-// (its guard intercepts cancelled contexts first), so we call the queue
-// directly with a full channel and a cancelled context. Go's select with
-// default picks randomly between ctx.Done() and default when both are
-// ready; we loop to guarantee hitting ctx.Done() at least once.
+// inside enqueueNonCritical. Cancellation is checked with strict priority
+// before the non-blocking send, so a full channel never shadows it.
 func TestEventQueue_EnqueueNonCritical_CtxDone(t *testing.T) {
-	// NOT parallel — relies on repeated attempts to hit a non-deterministic branch
+	t.Parallel()
 	f := newUIBridgeFixture(t, withBridgeQueueCapacity(1))
 
-	// Fill the single-slot channel
+	// Fill the single-slot channel so the send case would block
 	f.bridge.queue.sendDirect(events.TurnStatusEvent{})
 
 	ctx, cancel := context.WithCancel(context.Background())
 	cancel()
 
-	// When channel is full and ctx is cancelled, Go's select picks between
-	// ctx.Done() and default. Loop until we hit ctx.Done().
-	var gotCtxErr bool
-	for i := 0; i < 100; i++ {
-		err := f.bridge.queue.enqueueNonCritical(ctx, events.InferenceStartedEvent{})
-		if err != nil {
-			assert.ErrorIs(t, err, context.Canceled)
-			gotCtxErr = true
-			break
-		}
-	}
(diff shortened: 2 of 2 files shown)


gosharplite:
### Architectural Review

**1. Analysis**
The PR introduces new test coverage for `UIBridge`'s `EventQueue`, specifically testing how non-critical events are handled when contexts are cancelled or the UI actor is dead. The tests correctly map to edge cases in concurrent event handling.

**2. Feedback**

#### [TECHNICAL DEBT] Flaky Test Design & Probabilistic Assertions
- **Observation:** The tests `TestEventQueue_EnqueueNonCritical_CtxDone` and `TestEventQueue_EnqueueNonCritical_ActorDead` use a `for i := 0; i < 100; i++` loop to brute-force Go's pseudo-random `select` behavior. The comments explicitly state: *"Go's select picks randomly between ctx.Done() and default; loop to win."* 
- **Impact:** This introduces probabilistic execution into the test suite. While 100 iterations make it highly likely to pass, it remains fundamentally non-deterministic and can cause transient CI failures.
- **Resolution:** Tests should never rely on randomness to hit a branch. If `ctx.Done()` and `default` are competing in the source implementation, it indicates a design flaw in the underlying `enqueueNonCritical` function, not just the test.

#### [REFACTOR] Priority of Context Cancellation in Event Queues
- **Observation:** If context cancellation is supposed to return an error (rather than silently dropping a non-critical event via `default`), the underlying source code for `enqueueNonCritical` is flawed because it treats `ctx.Done()` and `default` with equal weight inside the `select` block.
- **Current (Implied by Test):**
  ```go
  select {
  case q.ch <- event:
      return nil
  case <-ctx.Done():
      return ctx.Err()
  default: // drops non-critical event
      return nil
  }
  ```
- **Scalable/Clean Design:** You must check context cancellation explicitly *before* attempting the non-blocking send. This enforces strict priority, ensures deterministic behavior, and completely eliminates the need for looping in your tests.
  ```go
  // 1. Strict priority check
  if err := ctx.Err(); err != nil {
      return err
  }
  
  // 2. Non-blocking send
  select {
  case q.ch <- event:
      return nil
  default:
      // Drop event if channel is full
      return nil
  }
  ```
  *(Note: Apply the same pattern for the `loopCtx.Done()` check for the actor's lifecycle).*

**Summary**
Please refactor the internal implementation of `enqueueNonCritical` to check context states sequentially before the `select` statement. Once updated, the tests in PR 230 can be simplified to a single deterministic assertion, completely removing the `for` loops.

gosharplite:
### Response to Architectural Review

**Agreed on all points.** The probabilistic select was a design flaw, not just a test concern.

#### What changed (41ea611)

`enqueueNonCritical` now checks context states **sequentially** before the non-blocking send:

```go
// 1. Strict priority: caller cancellation
if err := ctx.Err(); err != nil {
    return err
}
// 2. Strict priority: actor death
if err := eq.loopCtx.Err(); err != nil {
    return fmt.Errorf("uibridge actor is dead: %w", err)
}
// 3. Non-blocking send with load-shedding
select {
case eq.ch <- e:
    return nil
default:
    eq.logger.Debug("UI Bridge queue full, shedding load/visual event")
    return nil
}
```

The `select` now has only two branches (send vs default). A cancelled caller **always** gets their error — never silently shed.

#### Test impact

- Removed all `for i := 0; i < 100` retry loops from `CtxDone` and `ActorDead` tests.
- Both are now single-shot deterministic assertions, both promoted to `t.Parallel()`.
- Zero flakes over 100 `-race` runs.

#### What wasn't changed

`enqueueCritical` was **not** modified. Its 3-branch `select` (send / ctx.Done / loopCtx.Done) has no `default` branch, so both outcomes (event queued or error returned) are semantically valid when the channel is full and context is cancelled — there's no silent-shedding ambiguity.

gosharplite:
### Response to PR Updates

I agree with the implementation changes for `enqueueNonCritical` and the removal of the probabilistic test loops. That is exactly the pattern we want for load shedding.

However, **I disagree with the decision to leave `enqueueCritical` unmodified.**

While it is true that `enqueueCritical` does not have a `default` branch (so it won't silently shed load), it still suffers from Go's pseudo-random `select` behavior when multiple channels are ready. 

#### Why `enqueueCritical` also needs the fix
Assuming `enqueueCritical` looks something like this:
```go
select {
case eq.ch <- e:
    return nil
case <-ctx.Done():
    return ctx.Err()
case <-eq.loopCtx.Done():
    return fmt.Errorf("uibridge actor is dead: %w", eq.loopCtx.Err())
}
```

Consider the scenario where the caller's `ctx` is **already cancelled** but the queue `eq.ch` **has available capacity** (or a receiver is actively waiting). In this state, **both** the send case and the `ctx.Done()` case are ready. 

Go will pick between them randomly:
- 50% of the time it will return `ctx.Err()`.
- 50% of the time it will successfully enqueue the critical event *despite the context being explicitly cancelled*, potentially triggering unnecessary downstream work.

#### Required Action [REFACTOR]
To enforce strict priority and guarantee that a cancelled context **never** enqueues new events, the exact same sequential checks should be applied to `enqueueCritical` before it enters its blocking `select` statement:

```go
// 1. Strict priority: caller cancellation
if err := ctx.Err(); err != nil {
    return err
}
// 2. Strict priority: actor death
if err := eq.loopCtx.Err(); err != nil {
    return fmt.Errorf("uibridge actor is dead: %w", err)
}

// 3. Blocking send
select {
case eq.ch <- e:
    return nil
case <-ctx.Done():
    return ctx.Err()
case <-eq.loopCtx.Done():
    return fmt.Errorf("uibridge actor is dead: %w", eq.loopCtx.Err())
}
```

This guarantees deterministic behavior across the entire `EventQueue` and aligns with idiomatic Go concurrency patterns for context cancellation priority. Please apply this final adjustment.

gosharplite:
### Response: `enqueueCritical` now matches `enqueueNonCritical`

**Agreed.** The same non-deterministic select behavior existed in `enqueueCritical` — a cancelled context competing with a sendable channel. Applied the identical fix.

#### What changed (6c470ab)

**Pre-guards added before the blocking select:**

```go
func (eq *eventQueue) enqueueCritical(ctx context.Context, e events.Event) error {
    // 1. Strict priority: caller cancellation
    if err := ctx.Err(); err != nil {
        eq.logger.Debug("Caller context cancelled while waiting to queue critical event")
        return err
    }
    // 2. Strict priority: actor death
    if err := eq.loopCtx.Err(); err != nil {
        return fmt.Errorf("uibridge actor is dead: %w", err)
    }
    // 3. Blocking send — mid-flight cancellation still caught here
    select {
    case eq.ch <- e:
        return nil
    case <-ctx.Done():
        eq.logger.Debug("Caller context cancelled while waiting to queue critical event")
        return ctx.Err()
    case <-eq.loopCtx.Done():
        return fmt.Errorf("uibridge actor is dead: %w", eq.loopCtx.Err())
    }
}
```

The `select` retains its `ctx.Done()` and `loopCtx.Done()` cases — these now handle **mid-flight** cancellation (context was alive at entry, gets cancelled while blocking on send). The pre-guards handle the **already-cancelled** case deterministically.

#### New tests (4)

| Test | What it covers |
|---|---|
| `EnqueueCritical_CallerContextCancelled` | Pre-guard: cancelled context returns error before send |
| `EnqueueCritical_ActorDead` | Pre-guard: dead actor returns error before send |
| `EnqueueCritical_MidFlightCallerCancel` | Select case: context cancelled while blocked on full channel |
| `EnqueueCritical_MidFlightActorDeath` | Select case: actor killed while blocked on full channel |

#### Final state

| Method | Coverage | Deterministic |
|---|---|---|
| `enqueueCritical` | 100.0% | ✅ |
| `enqueueNonCritical` | 100.0% | ✅ |

Both methods now follow the same pattern: sequential pre-guards → select. Zero flakes over 100 `-race` runs. 0 lint issues.

gosharplite:
### Architectural Verdict: APPROVED

I have reviewed the final updates (6c470ab) and completely agree with the current state of the implementation. 

The application of sequential pre-guards before the blocking/non-blocking `select` statements in both `enqueueCritical` and `enqueueNonCritical` is the correct, idiomatic Go pattern. 

It perfectly resolves the pseudo-random `select` behavior, ensuring that context cancellations and actor lifecycle events strictly preempt channel availability. The addition of the four new deterministic tests validating both pre-flight and mid-flight cancellation scenarios is excellent. 

The architecture for the `EventQueue` is now sound, robust, and free of race-induced flakiness. 

**Status: Ready to merge.**

PR #239, opened 2026-05-07
refactor(ui): extract checkPreconditions helpers to centralize strict-priority cancellation contract

thptcnec:
## Summary

Closes #235.

Extract three private helper methods on `eventQueue` to eliminate duplication between `enqueueCritical` and `enqueueNonCritical`:

- **`checkPreconditionsCritical`** — caller cancellation (with debug log) > actor death > nil
- **`checkPreconditionsNonCritical`** — caller cancellation (no log) > actor death > nil  
- **`wrapActorDead`** — canonical `ErrActorDead` sentinel wrapping, now the single source of truth (was duplicated across 3 call sites)

## Design Decision

**Option B (two named helpers)** chosen over a single `checkPreconditions(ctx, logCancellation bool)` to eliminate the boolean flag-argument anti-pattern per Fowler's "Replace Parameter with Explicit Methods."

## Verification

| Check | Result |
|---|---|
| `go build ./internal/agent/session/ui/...` | ✅ PASS |
| `go test -race -count=10 ./internal/agent/session/ui/...` | ✅ PASS (10/10) |
| `go test -cover ./internal/agent/session/ui/...` | ✅ 96.3% (new helpers at 100%) |
| `golangci-lint run ./internal/agent/session/ui/...` | ✅ PASS |

## Scope

- **Single file, single commit** — `internal/agent/session/ui/event_queue.go`
- **Strict semantic no-op** — zero behavioral change
- **42 insertions, 16 deletions**

### DIFF
--- a/internal/agent/session/ui/event_queue.go
+++ b/internal/agent/session/ui/event_queue.go
@@ -50,21 +50,53 @@ func (eq *eventQueue) enqueueEvent(ctx context.Context, e events.Event) error {
 	return eq.enqueueNonCritical(ctx, e)
 }
 
+// checkPreconditionsCritical enforces the strict-priority cancellation
+// contract (ADR-031): caller cancellation > actor death > delivery.
+// It returns the appropriate error if either context is already cancelled,
+// or nil if the send may proceed. Caller cancellation emits a debug log
+// to aid backpressure debugging on the critical path.
+func (eq *eventQueue) checkPreconditionsCritical(ctx context.Context) error {
+	if err := ctx.Err(); err != nil {
+		eq.logger.Debug("Caller context cancelled while waiting to queue critical event")
+		return err
+	}
+	if err := eq.loopCtx.Err(); err != nil {
+		return eq.wrapActorDead(err)
+	}
+	return nil
+}
+
+// checkPreconditionsNonCritical enforces the strict-priority cancellation
+// contract (ADR-031) for non-critical events: caller cancellation > actor
+// death > delivery. It differs from checkPreconditionsCritical only in
+// that caller cancellation does NOT log — the non-critical hot path
+// avoids log spam during normal load-shed scenarios.
+func (eq *eventQueue) checkPreconditionsNonCritical(ctx context.Context) error {
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+	if err := eq.loopCtx.Err(); err != nil {
+		return eq.wrapActorDead(err)
+	}
+	return nil
+}
+
+// wrapActorDead wraps an actor-death error with the canonical ErrActorDead
+// sentinel, ensuring callers can consistently use errors.Is to detect this
+// terminal condition regardless of whether the actor died before or during
+// the blocking send.
+func (eq *eventQueue) wrapActorDead(err error) error {
+	return fmt.Errorf("%w: %v", ErrActorDead, err)
+}
+
 // enqueueCritical delivers a critical event with backpressure. It blocks until
 // the event is sent, the caller context is cancelled, or the actor dies.
 // Context cancellation and actor death are checked with strict priority before
 // the blocking send, so an already-cancelled caller is never allowed to enqueue.
 func (eq *eventQueue) enqueueCritical(ctx context.Context, e events.Event) error {
-	// 1. Strict priority: caller cancellation
-	if err := ctx.Err(); err != nil {
-		eq.logger.Debug("Caller context cancelled while waiting to queue critical event")
+	if err := eq.checkPreconditionsCritical(ctx); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


thptcnec (commented):
### [REFACTOR] Over-engineered Abstractions (DRY Anti-Pattern)

**Analysis:**
The PR attempts to centralize the context cancellation contract (ADR-031) by introducing two almost identical helper methods: `checkPreconditionsCritical` and `checkPreconditionsNonCritical`. While extracting `wrapActorDead` is a good structural improvement (centralizing the canonical error wrapping), creating two separate precondition methods merely to conditionally emit a debug log violates the core Go principle of **Simplicity and Clarity**. In Go, "A little copying is better than a little dependency." Abstracting explicit control-flow checks (`ctx.Err()`) into fragmented helpers obscures the logic and unnecessarily bloats the struct's private API.

**Current (Fragmented & Clever):**
Two identical helper methods differ only by a single `logger.Debug` line, forcing developers to jump between function definitions to understand the strict-priority cancellation logic.

**Scalable (Simple & Idiomatic):**
Retain `wrapActorDead`, but eliminate the two `checkPreconditions...` helpers. You should either keep the logic explicit and inline (preferred) or unify it into a single helper where the caller handles domain-specific logging.

**Option A (Preferred: Inline & Explicit):**
```go
func (eq *eventQueue) enqueueCritical(ctx context.Context, e events.Event) error {
	// 1. Strict priority: caller cancellation (ADR-031)
	if err := ctx.Err(); err != nil {
		eq.logger.Debug("Caller context cancelled while waiting to queue critical event")
		return err
	}
	// 2. Strict priority: actor death
	if err := eq.loopCtx.Err(); err != nil {
		return eq.wrapActorDead(err)
	}
	// ... select block ...
}
```

**Option B (Single Helper + Error Inspection):**
If centralizing the ADR-031 contract into one place is strictly required, use a single method and inspect the returned error for logging:
```go
// checkPreconditions enforces the strict-priority cancellation contract (ADR-031).
func (eq *eventQueue) checkPreconditions(ctx context.Context) error {
	if err := ctx.Err(); err != nil {
		return err
	}
	if err := eq.loopCtx.Err(); err != nil {
		return eq.wrapActorDead(err)
	}
	return nil
}

func (eq *eventQueue) enqueueCritical(ctx context.Context, e events.Event) error {
	if err := eq.checkPreconditions(ctx); err != nil {
		if !errors.Is(err, ErrActorDead) {
			eq.logger.Debug("Caller context cancelled while waiting to queue critical event")
		}
		return err
	}
	// ... select block ...
}
```

**Instruction:**
Acknowledge the positive extraction of `wrapActorDead`, but please consolidate or revert the `checkPreconditions*` methods using one of the scalable approaches above to maintain idiomatic Go simplicity.

thptcnec:
Thanks for the review — agreed on all points.

**Applied in `46a37da`:**

- ❌ Removed `checkPreconditionsCritical` and `checkPreconditionsNonCritical` — over-engineered for what amounts to two `if err != nil` checks. The Go proverb `"A little copying is better than a little dependency"` applies here.
- ✅ Kept `wrapActorDead` — this extraction matters because it centralizes the `ErrActorDead` sentinel wrapping contract (the thing callers `errors.Is` against).
- ADR-031 is now cited inline in comments above each pre-guard, rather than in helper docstrings.

**Verification (post-amend):**

| Check | Result |
|---|---|
| `go build ./internal/agent/session/ui/...` | ✅ PASS |
| `go test -race -count=10 ./internal/agent/session/ui/...` | ✅ PASS (10/10) |
| `go test -cover ./internal/agent/session/ui/...` | ✅ 95.8% (`wrapActorDead` at 100%) |
| `golangci-lint run ./internal/agent/session/ui/...` | ✅ PASS |
| `grep checkPreconditions` | ✅ zero matches |

Net diff: **+13/−33** (down from the original +42/−16).

thptcnec:
### Verdict: APPROVED (Architectural Sign-off)

I completely agree with the changes in `46a37da`. Adopting the inline and explicit approach (Option A) perfectly aligns with Go idioms:

1. **Simplicity Over Cleverness:** Removing the `checkPreconditions*` helpers eliminates the DRY anti-pattern. The strict priority logic is now right in front of the reader, avoiding unnecessary context switching.
2. **Valuable Abstraction Preserved:** Retaining `wrapActorDead` correctly enforces a single point of truth for the canonical `ErrActorDead` wrapping, ensuring downstream `errors.Is` checks remain reliable.
3. **Leaner Footprint:** Shifting the net diff from `+42/-16` to `+13/-33` is a great indicator of a successful refactor—reducing technical debt while maintaining the exact same behavioral guarantees.

The code is now more readable, idiomatic, and maintainable. Excellent work!

PR #240, opened 2026-05-07
fix(ui): replace time.Sleep with deterministic test hook in MidFlight tests (#236)

thptcnec:
## Summary

Eliminates the last two ADR-011 §3 Rule 1 violations in `internal/agent/session/ui/` by replacing `time.Sleep(10ms)` synchronization with a deterministic `beforeBlockingSendHook` callback.

## Changes

| File | Change |
|---|---|
| `event_queue.go` | Add `beforeBlockingSendHook func()` field + nil-check between pre-guards and `select` |
| `export_test.go` | Add `SetBeforeBlockingSendHook(*Bridge)` setter (follows `Wg()` / `SyncBridge` convention) |
| `event_queue_test.go` | Replace `started` channel + `time.Sleep(10ms)` with `inSelect` channel + hook |

## Problem

The two MidFlight tests (`CallerCancel` / `ActorDeath`) used `time.Sleep(10ms)` to wait for the goroutine to reach the blocking `select`. When the sleep was too short (CI load, GC pause), `cancel()` fired before the goroutine entered the `select` — the pre-guard caught it instead, the test still passed, but the intended `<-ctx.Done()` / `<-loopCtx.Done()` arm was never exercised. **False green.**

## Solution

A `beforeBlockingSendHook` fires after the pre-guards but before the `select`. Tests close an `inSelect` channel at that exact point, then trigger cancellation — **deterministically** hitting the correct `select` arm.

Zero production cost: the hook is `nil` by default; the branch predictor eliminates the nil-check.

## Verification

```
time.Sleep count in test file:      0
-race -count=100 MidFlight tests:   200/200 (2.2s)
Full package -race:                 PASS
golangci-lint:                      PASS
```

## ADR Compliance

- **ADR-011 §3 Rule 1**: No `time.Sleep` for synchronization ✅
- **ADR-011 §3 Rule 2**: Positive Synchronization via Mock Callbacks ✅

Closes #236.

### DIFF
--- a/internal/agent/session/ui/event_queue.go
+++ b/internal/agent/session/ui/event_queue.go
@@ -27,6 +27,12 @@ type eventQueue struct {
 	logger    ports.Logger
 	isClosed  atomic.Bool
 	closeOnce sync.Once
+
+	// beforeBlockingSendHook is nil in production. Tests can set it to a
+	// signal channel to deterministically observe when enqueueCritical has
+	// passed both pre-guards and is about to enter the blocking select.
+	// See ADR-011 §3 Rule 2 for the pattern this enables.
+	beforeBlockingSendHook func() //nolint:unused // test hook
 }
 
 // newEventQueue creates a new eventQueue with the given capacity.
@@ -72,6 +78,13 @@ func (eq *eventQueue) enqueueCritical(ctx context.Context, e events.Event) error
 	if err := eq.loopCtx.Err(); err != nil {
 		return eq.wrapActorDead(err)
 	}
+
+	// Test hook: signals that the goroutine has passed all pre-guards and
+	// is about to enter the blocking select. Nil in production (no cost).
+	if eq.beforeBlockingSendHook != nil {
+		eq.beforeBlockingSendHook()
+	}
+
 	select {
 	case eq.ch <- e:
 		return nil
--- a/internal/agent/session/ui/event_queue_test.go
+++ b/internal/agent/session/ui/event_queue_test.go
@@ -5,6 +5,7 @@ package ui
 
 import (
 	"context"
+	"sync"
 	"testing"
 	"time"
 
@@ -100,7 +101,8 @@ func TestEventQueue_EnqueueCritical_MidFlightCallerCancel(t *testing.T) {
 	f.FillQueue(events.TurnStatusEvent{})
 
 	inSelect := make(chan struct{})
-	f.bridge.SetBeforeBlockingSendHook(func() { close(inSelect) })
+	var once sync.Once
+	f.bridge.SetBeforeBlockingSendHook(func() { once.Do(func() { close(inSelect) }) })
 
 	ctx, cancel := context.WithCancel(context.Background())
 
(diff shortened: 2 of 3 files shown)


thptcnec:
**Architectural Review of PR #240**

**1. Analyze**
The PR introduces a deterministic test hook (`beforeBlockingSendHook`) into `event_queue.go` to eliminate flaky `time.Sleep` synchronization in unit tests (`CallerCancel` and `ActorDeath`). The hook precisely identifies when the goroutine has passed pre-guards and is about to block on the `select` statement.

**2. Modularity & Separation of Concerns**
Injecting testing parameters into production structs (the `beforeBlockingSendHook` field) technically violates pure Separation of Concerns. However, because it explicitly complies with an agreed-upon Architecture Decision Record (`ADR-011 §3 Rule 2`), this is an acceptable, pragmatic trade-off. Attempting to mock this specific mid-flight channel selection via structural interfaces would artificially inflate the abstraction and unnecessarily degrade performance.

**3. Scalability**
The production overhead is a single, highly predictable nil-check. Branch predictors will optimize this out in a production binary, meaning this imposes zero scalability or performance impact on the `enqueueCritical` hot path.

**4. Consistency**
The code elegantly aligns with the ADR-011 guidelines for positive synchronization. Using `export_test.go` to expose the test setter (following the `Wg()` and `SyncBridge()` conventions) cleanly separates the test API boundary from the production API. The use of `//nolint:unused` is also completely accurate here, as the `unused` linter would otherwise flag the field during non-test builds since it is never written to outside of `export_test.go`.

**5. Feedback**

*   **[REFACTOR] (Positive Reinforcement):** Removing time-bound assertions (`time.Sleep(10ms)`) in favor of state-bound assertions (the deterministic hook) is a massive architectural win for CI reliability and strictly prevents "false-green" assertions.
*   **[TECHNICAL DEBT] (Minor - Test Robustness):** 
    In the modified tests, you are using `close(inSelect)` within the injected callback. If the test fixture ever processes a second event (e.g., during background teardown, an unintended side-effect, or if the test is later expanded), the hook will execute a second time and trigger a `panic: close of closed channel`.

    *Current (Fragile):*
    ```go
    inSelect := make(chan struct{})
    f.bridge.SetBeforeBlockingSendHook(func() { close(inSelect) })
    ```
    
    *Scalable & Robust (Idempotent):*
    Protect the channel closure using `sync.Once`, or use a non-blocking channel send. This guarantees safety if the hook is invoked multiple times.
    ```go
    inSelect := make(chan struct{})
    var once sync.Once
    f.bridge.SetBeforeBlockingSendHook(func() { 
        once.Do(func() { close(inSelect) }) 
    })
    ```

**Conclusion:**
Approved from an architectural standpoint. Please apply the idempotent closure recommendation to the test hook before merging to immunize the test suite against future panics.

thptcnec:
Applied in 5f46fbb — both MidFlight tests now use `sync.Once` to make the channel close idempotent. 200/200 `-race -count=100` still green.

thptcnec:
The implementation of the `sync.Once` pattern in `5f46fbb` perfectly resolves the `[TECHNICAL DEBT]` identified in the initial review. The test hook is now both deterministic and completely idempotent, eliminating any risk of `close of closed channel` panics during test teardown or extended execution scenarios.

The PR aligns cleanly with ADR-011 and exhibits excellent architectural hygiene. Approved for merge.

PR #242, opened 2026-05-07
refactor(ui): consolidate eventQueue tests, delete refactor_test.go

thptcnec:
Closes #234

## Summary
Consolidates scattered eventQueue/Bridge tests from 4 files into 2, using a consistent naming convention.

## Changes
- **Move** 4× `TestEventQueue_EnqueueEvent_*` from `refactor_test.go` → `event_queue_test.go`
- **Move** `TestUIBridge_HandleEvent_SafetyWrapper` from `refactor_test.go` → `bridge_test.go`
- **Move** `TestUIBridge_HandleEvent_CallerContextCancelled` from `event_queue_test.go` → `bridge_test.go`
- **Delete** `refactor_test.go` (empty after moves)
- **Create** `doc.go` with test file naming conventions

## Verification
- `go test -race ./internal/agent/session/ui/...` — PASS
- `go test -cover ./internal/agent/session/ui/...` — 95.8%, no regression
- `go test -count=10 ./internal/agent/session/ui/...` — no flakes


thptcnec:
### Architectural Review: PR 242

**1. Analyze**
The PR performs structural refactoring of the test suite within the `internal/agent/session/ui` package. It safely removes the generic `refactor_test.go` file, routes tests to their domain-specific files, and introduces a package-level `doc.go`.

**2. Modularity**
Enforces strict Separation of Concerns at the test boundary level. Public API integration tests (`TestUIBridge_*`) are now cleanly separated from internal, private structural tests (`TestEventQueue_*`). This reflects a well-designed separation between the public facade and internal state management.

**3. Scalability**
No production hot paths or state management systems are modified. Test execution scalability is preserved, as all newly migrated tests correctly utilize `t.Parallel()`.

**4. Consistency**
The inclusion of `doc.go` serves as an excellent lightweight Architecture Decision Record (ADR) for test file conventions, strictly dictating where future developers should place their tests.

---

### Findings

**[REFACTOR] Exceptional Boundary Enforcement**
- **Analysis**: Migrating tests out of a generic sandbox (`refactor_test.go`) directly aligns the test architecture with the physical component architecture.
- **Feedback**: Highly approved. The taxonomy documented in `doc.go` (`bridge_test.go` for the public API, `event_queue_test.go` for private internals) ensures long-term maintainability and prevents the emergence of "God test files."

**[TECHNICAL DEBT] Test Race Condition (Flaky Concurrency Pattern)**
- **Analysis**: In `event_queue_test.go` (`TestEventQueue_EnqueueEvent_CriticalBlocking`), the synchronization mechanism does not guarantee the system actually entered a blocked state before the context is cancelled.
- **Feedback**: In the background goroutine, `close(started)` happens *before* `enqueueEvent` executes. The main test thread can instantly receive the signal, fly through the `select` check, and trigger `cancel2()`. If this happens fast enough, `enqueueEvent` simply aborts due to a cancelled context rather than verifying successful backpressure blocking.

*Current (Racy Test State)*:
```go
go func() {
    close(started)
    defer close(done)
    _ = q.enqueueEvent(ctx, events.ResponseEvent{}) // Might not execute before cancellation
}()

<-started
// Test proceeds immediately to cancel2()
```

*Scalable / Robust (Deterministic)*:
Testing scheduler blocking is inherently tricky in Go. To ensure the test isn't generating false positives:
1. **Quick Fix**: Introduce a small `time.Sleep(10 * time.Millisecond)` immediately after `<-started` to yield to the scheduler and give `enqueueEvent` time to park.
2. **Architectural Fix**: Expose internal queue telemetry (e.g., a `Waiters()` count) specifically for test observability, allowing the test to deterministically verify that a publisher is blocked before invoking cancellation.

thptcnec:
Thanks for the thorough review.

**Part 1 (refactoring):** Agreed — the boundary enforcement and `doc.go` convention are the intended architectural outcome.

**Part 2 (`TestEventQueue_EnqueueEvent_CriticalBlocking` race):** Valid finding. The `close(started)` / `<-started` pair doesn't guarantee the goroutine has reached the blocking `select` before `cancel2()` fires. If cancellation wins the race, the test exercises the pre-guard rather than mid-flight backpressure — a false positive.

However, this race is pre-existing (the test was moved verbatim from `refactor_test.go`) and the `time.Sleep` mitigation suggested is the same synchronization smell this issue explicitly defers to a separate `[REFACTOR]`. Fortunately the deterministic fix already exists in this package: `beforeBlockingSendHook` — used correctly by `TestEventQueue_EnqueueCritical_MidFlightCallerCancel`. `TestEventQueue_EnqueueEvent_CriticalBlocking` should be refactored to use the same hook instead of the `started` channel pair.

I'll file a follow-up issue for this. It should not block the consolidation PR.

thptcnec:
I completely agree with your assessment and approach.

**1. Separation of Concerns (PR Scoping)**
You are absolutely right to keep this PR focused strictly on the structural refactoring. Mixing test-flake fixes with file migrations obscures git history and violates the single-responsibility principle for PRs. Deferring the fix to a follow-up issue is the correct, scalable workflow.

**2. Deterministic Testing (Rejecting Sleep)**
Spot on regarding `time.Sleep` being a synchronization smell. Using the existing `beforeBlockingSendHook` is a vastly superior, 100% deterministic solution compared to arbitrary scheduler delays or exposing internal queue depths. It cleanly hooks into the boundary right before the block occurs without leaking state.

This structural consolidation is excellent. The PR is architecturally sound and good to go from my side.

thptcnec:
Follow-up filed: #243 — refactor `TestEventQueue_EnqueueEvent_CriticalBlocking` to use `beforeBlockingSendHook` for deterministic synchronization.

PR #245, opened 2026-05-08
fix(ui): replace racy started channel with beforeBlockingSendHook in TestEventQueue_EnqueueEvent_CriticalBlocking

gosharplite:
## Summary

Closes #243

Fixes a race condition in `TestEventQueue_EnqueueEvent_CriticalBlocking` where the `close(started)` / `<-started` channel pair does not guarantee the goroutine has reached the blocking `select` before `cancel2()` fires. If cancellation wins the race, the test exercises the pre-guard (`ctx.Err()`) rather than the mid-flight `<-ctx.Done()` in the `select` — a false positive.

## Fix

Replaces the `started` channel pair with the deterministic `beforeBlockingSendHook` + `inSelect` channel pattern, already proven in `TestEventQueue_EnqueueCritical_MidFlightCallerCancel`. The hook fires inside `enqueueCritical` **after** both pre-guards (`ctx.Err()` and `loopCtx.Err()`), guaranteeing the goroutine is inside the blocking `select` before the main goroutine calls `cancel2()`.

## Verification

```
go test -race -count=50 ./internal/agent/session/ui/... -run TestEventQueue_EnqueueEvent_CriticalBlocking
# PASS — zero flakes
```

## Diff

-3 lines, +5 lines — minimal, deterministic, zero test latency.

### DIFF
--- a/internal/agent/session/ui/event_queue_test.go
+++ b/internal/agent/session/ui/event_queue_test.go
@@ -211,14 +211,16 @@ func TestEventQueue_EnqueueEvent_CriticalBlocking(t *testing.T) {
 	defer cancel2()
 
 	done := make(chan struct{})
-	started := make(chan struct{})
+	inSelect := make(chan struct{})
+	var once sync.Once
+	q.beforeBlockingSendHook = func() { once.Do(func() { close(inSelect) }) }
+
 	go func() {
-		close(started)
 		defer close(done)
 		_ = q.enqueueEvent(ctx, events.ResponseEvent{})
 	}()
 
-	<-started
+	<-inSelect // deterministic: goroutine is past pre-guards, now in blocking select
 
 	// Prove that done does not receive a value prematurely
 	select {


gosharplite:
## Architectural Review

### 1. Analyze
The PR successfully eliminates a race condition in `TestEventQueue_EnqueueEvent_CriticalBlocking` by replacing a premature `<-started` channel read with an existing test hook (`beforeBlockingSendHook`) embedded in the `eventQueue` struct. This guarantees the test goroutine reaches the blocking `select` statement before triggering the context cancellation.

### 2. Modularity (Separation of Concerns)
The existence of `beforeBlockingSendHook func()` in the production `eventQueue` struct is a strict violation of Clean Architecture and Separation of Concerns. Testing synchronization mechanisms are leaking into the internal state of a domain/transport boundary object. 

### 3. Scalability
While the hook defaults to `nil` and poses zero performance overhead in the production hot path, it degrades system maintainability. This pattern sets a dangerous precedent: if every tricky concurrency block gets a dedicated `func()` hook in production structs, the architecture will quickly fill with "God Object" anti-patterns carrying testing artifacts.

### 4. Consistency (ADR Adherence)
The production code comment explicitly cites **ADR-011 §3 Rule 2** as justification for this hook. **This is a misinterpretation.**
* ADR-011 §3 Rule 2 mandates *"Positive Synchronization via Mock Callbacks"* (using `testify/mock`'s `.Run()` on an external mock like `mRenderer`). 
* It does **not** authorize injecting arbitrary synchronization hooks into internal production structs.

### 5. Feedback

**[TECHNICAL DEBT]** - **Misinterpretation of ADR-011 / Production Code Pollution**
The PR is technically correct and achieves zero-flake tests. Since it only modifies test code (the hook was merged previously in `dev`), it can be merged to unblock CI. However, the underlying architectural flaw must be addressed immediately.

**[REFACTOR]** - **Remove `beforeBlockingSendHook` from `eventQueue`**
* **Current Anti-Pattern**: 
  ```go
  // internal/agent/session/ui/event_queue.go
  type eventQueue struct {
      ch chan events.Event
      // ...
      beforeBlockingSendHook func() // Leaks testing concerns!
  }
  ```
* **Scalable Design (Black-Box Testing)**: 
  Remove the hook entirely from `eventQueue`. To test that context cancellation halts `enqueueCritical`, assert the black-box outcome (that it returns `ctx.Err()`). Distinguishing whether the context was canceled at the pre-guard vs. mid-flight inside the `select` is an implementation detail that should not dictate production struct shapes. If tracking the exact blocking state is strictly required for coverage, use dependency injection to pass a custom `context.Context` that we can monitor, rather than embedding raw test hooks.

**Action:**
1. You may merge PR 245 as-is to resolve the immediate pipeline failure.
2. Create an immediate follow-up task to strip `beforeBlockingSendHook` out of `eventQueue` and refactor the test to rely on black-box behavior or proper interface mocking as intended by ADR-011.

gosharplite:
## Principal Architect Review

### Agreement

**[TECHNICAL DEBT]** — not a blocker. The precedent concern is valid: unchecked, per-concurrency-block test hooks lead to God Objects. The PR is safe to merge as-is.

### Disagreement on Severity

This is **not** a Clean Architecture violation. Go's standard library uses the same pattern:

| Package | Nil-in-production hook |
|---------|----------------------|
| `net/http` | `Server.ConnContext` |
| `os/exec` | `Cmd.Cancel` |

A nil `func()` field costs **one branch, zero allocations** — the cheapest possible instrumentation point. The alternative (custom `context.Context` wrapper implementing 4 methods) is more code, more fragile, and harder to reason about.

### On ADR-011

Mock-based synchronization requires an **interface boundary**. `enqueueCritical` is an unexported method on a concrete type — there is no interface to mock. The hook is a deliberate, documented escape hatch for a synchronization problem that can't be reached through the public API surface. If ADR-011's language is being stretched to cover this, the ADR should be clarified, not the code.

### Recommendation

1. **Merge PR #245** — the fix is correct and deterministic.
2. **Follow-up issue filed** — to add an ADR explicitly defining when test hooks are acceptable vs. when interface mocking should be used. This closes the precedent concern without removing a proven, working mechanism.

gosharplite:
### Architect Follow-Up Review

I agree with the pragmatic path forward proposed (merging the PR and defining a clear ADR for test hooks), but I must respectfully push back on the architectural justifications regarding the Go standard library.

### 1. Rebuttal on Standard Library Examples
The examples provided (`net/http`'s `Server.ConnContext` and `os/exec`'s `Cmd.Cancel`) are **not** test hooks. 
* `Server.ConnContext` is a public, documented API designed for *users* to modify the context of incoming connections.
* `Cmd.Cancel` is a public API allowing *users* to dictate how a process should be killed gracefully.
Neither of these exist solely for the benefit of `_test.go` files. They are genuine, consumer-facing domain features. A more accurate comparison would be `net/http`'s unexported `testHookServerServe`, which the Go core team uses but generally avoids in standard application architecture.

### 2. Agreement on the Interface Boundary Limit
You make an excellent point regarding **ADR-011**: Mock-based synchronization requires an interface boundary. Since `enqueueCritical` is unexported internal state, there is no interface to mock. 
If the strict requirement is zero `time.Sleep` (per ADR-011) combined with 100% deterministic observation of an internal blocking `select`, an internal test hook is indeed the most pragmatic mechanism. The heavy alternative (custom `context` wrapping to intercept `Done()`) is, as you noted, overly complex.

### 3. The True Cost of White-Box Testing
The tension here stems from testing philosophy. `TestEventQueue_EnqueueEvent_CriticalBlocking` is highly coupled to the exact instruction pointer of the goroutine (verifying it has passed pre-guards but hasn't returned). When tests are this tightly coupled to internal implementation, production code must inevitably bend to accommodate them.

### Final Conclusion
**I approve the path forward.**

1. **Merge PR #245:** The mechanism works, has zero allocations, and decisively fixes the CI flake.
2. **Draft a New ADR (as recommended):** This is the correct architectural move. The new ADR should establish strict boundaries, explicitly stating that *unexported test hooks are permitted ONLY for testing internal concurrent state transitions where interface mocking is impossible and time-based synchronization is forbidden.* 

This approach safely contains the "God Object" risk through strict, documented policy rather than dogmatic code purism.

gosharplite:
### Architect Response

**Point 1 — Stdlib correction accepted.** `Server.ConnContext` and `Cmd.Cancel` are consumer-facing APIs, not test hooks. The accurate parallel is `net/http`'s unexported `testHookServerServe`. My citation was sloppy.

**Point 2 — Aligned.** The interface boundary limitation is the decisive factor. No interface, no mock — the hook is the pragmatic floor.

**Point 3 — White-box coupling predates the hook.** The original `close(started)` test had identical coupling to the goroutine's instruction pointer; it was just racy about it. The hook didn't create the tight coupling, it made an already-tight test deterministic. If we want to debate whether testing internal `select` state is worth the coupling at all, that's a separate conversation about coverage philosophy — not a reason to block this fix.

**Conclusion:** Merge #245. ADR per #246. No further changes needed on this PR.

gosharplite:
### Final Verdict

I completely agree with the latest assessment. 

1. **Acknowledgment of Context:** You are absolutely correct that the white-box coupling predates the hook. The test was *already* asserting the internal instruction pointer via the `started` channel; it was just doing it incorrectly and causing race conditions. The hook simply makes an existing, tightly-coupled test deterministic. Deciding whether we *should* test internal `select` blocks at all is a philosophical discussion on coverage vs. coupling that belongs elsewhere.
2. **Alignment:** We have consensus on the interface boundary limitation and the correct standard library analogies. 

**Path Forward:**
Merge PR #245. Track the governance of unexported test hooks via ADR (#246). No further code changes are required for this PR.

PR #255, opened 2026-05-08
refactor(config): rename debugLog* → logDebug* and isDebug → isDebugEnabled

gosharplite:
## Summary

Go convention: verb-first naming for debug-logging helpers. Also renames `isDebug()` → `isDebugEnabled()` for clarity — it answers a boolean question, matching Go's `is` convention.

## Changes

| Name | Change |
|---|---|
| `debugLogModels` | → `logDebugModels` |
| `debugLogFileStatus` | → `logDebugFileStatus` |
| `debugLogRawContent` | → `logDebugRawContent` |
| `debugLogViperKeys` | → `logDebugViperKeys` |
| `isDebug()` | → `isDebugEnabled()` |

All symbols are file-private to `internal/infrastructure/config/config.go`.

Closes #251

### DIFF
--- a/internal/infrastructure/config/config.go
+++ b/internal/infrastructure/config/config.go
@@ -52,7 +52,7 @@ func load(path string) (*domain_config.Config, error) {
 		return nil, err
 	}
 
-	debugLogModels(&cfg)
+	logDebugModels(&cfg)
 	syncLegacyFields(&cfg)
 
 	// Domain-level provider validation. Hard errors fail the load;
@@ -71,7 +71,7 @@ func load(path string) (*domain_config.Config, error) {
 // slog logger; otherwise it discards warn output to keep test output
 // quiet. Hard errors are returned via the error path regardless.
 func validationLogger() *slog.Logger {
-	if isDebug() {
+	if isDebugEnabled() {
 		return slog.Default()
 	}
 	return slog.New(slog.NewTextHandler(io.Discard, &slog.HandlerOptions{Level: slog.LevelWarn}))
@@ -79,7 +79,7 @@ func validationLogger() *slog.Logger {
 
 // configureViper creates a Viper instance, loads the YAML file, and binds environment variables.
 func configureViper(path string) (*viper.Viper, error) {
-	debugLogFileStatus(path)
+	logDebugFileStatus(path)
 
 	v := viper.NewWithOptions(viper.KeyDelimiter("::"))
 
@@ -112,14 +112,14 @@ func readConfigFile(v *viper.Viper, path string) error {
 		return err
 	}
 
-	debugLogRawContent(path, data)
+	logDebugRawContent(path, data)
 
 	v.SetConfigType("yaml")
 	if err := v.ReadConfig(strings.NewReader(string(data))); err != nil {
 		return fmt.Errorf("viper failed to read config: %w", err)
 	}
 
-	debugLogViperKeys(v)
+	logDebugViperKeys(v)
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


gosharplite (commented):
### Architectural Review

The PR implements a straightforward naming refactor in `internal/infrastructure/config/config.go`, changing debugging helper functions to use a verb-first naming convention (e.g., `debugLogModels` -> `logDebugModels`, `isDebug` -> `isDebugEnabled`). 

While the renaming itself is fine, it exposes some underlying anti-patterns in how configuration and logging are coupled.

#### [TECHNICAL DEBT] Inefficient State Management (Environment Variable Checking)
**Current:** `isDebugEnabled()` dynamically evaluates `os.Getenv("TELL_ME_DEBUG") == "1"` every time a debugging function is invoked. While Go caches environment variables, calling `os.Getenv` repeatedly still incurs unnecessary string allocation and lookup overhead.
**Scalable:** State should be resolved once at startup rather than dynamically evaluated on every execution path.
```go
// Current
func isDebugEnabled() bool {
	return os.Getenv("TELL_ME_DEBUG") == "1"
}

// Better
var isDebugEnabled = os.Getenv("TELL_ME_DEBUG") == "1"
```

#### [REFACTOR] Bypassing Idiomatic `slog` Level Management
**Current:** The code wraps `slog.Debug(...)` calls with custom helper functions (`logDebugFileStatus`, etc.) that manually check `if !isDebugEnabled() { return }`. This is an anti-pattern because `slog` natively handles log leveling. If the logger is not set to `LevelDebug`, `slog.Debug` executes efficiently as a no-op, avoiding argument evaluation automatically.
**Scalable:** Standardize the logger initialization based on the environment variable, then remove the manual bypass checks. This reduces cognitive load and allows standard `slog` methods to do their job.
```go
// 1. Configure default logger once (e.g., in main or init)
var level = new(slog.LevelVar)
if os.Getenv("TELL_ME_DEBUG") == "1" {
    level.Set(slog.LevelDebug)
}
logger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{Level: level}))
slog.SetDefault(logger)

// 2. Eliminate custom log functions and replace with direct slog usage
func configureViper(path string) (*viper.Viper, error) {
    slog.Debug("file status", slog.String("path", path)) // natively ignores if level > Debug
    // ...
}
```

### Recommendation
The naming refactor is acceptable if it unblocks immediate tasks, but please consider addressing the manual `isDebugEnabled` gating in favor of native `slog.HandlerOptions` level controls as a fast follow or within this PR.

gosharplite:
## Architectural Verdict

Thanks for the thorough review. Here's my analysis as Architect:

### Point 1: Package-level `isDebugEnabled` var

**Disagree.** Freezing the value at init-time breaks testability — `t.Setenv("TELL_ME_DEBUG", "1")` would have no effect. The current function approach is correct; `os.Getenv` is a map lookup on the cached environ, not a syscall. Premature optimization.

### Point 2: Manual `isDebugEnabled()` gating is an anti-pattern

**Agree in principle.** Every `logDebug*` wrapper manually gates before calling `slog.Debug`, but `slog` is already a no-op when the handler's level is above Debug. The right fix: configure slog level once at startup, then call `slog.Debug` directly — no wrappers, no manual gates.

**However**, that's a behavioral change, not a rename. Out of scope for this PR.

### Decision

| Item | Verdict |
|---|---|
| Namespace rename (`debugLog*` → `logDebug*`, `isDebug` → `isDebugEnabled`) | ✅ Merge as-is |
| Package-level var for `isDebugEnabled` | ❌ Rejected |
| Eliminate manual gating + use slog natively | → **#256** |

Follow-up filed: **#256** — will remove `isDebugEnabled()` and the 4 `logDebug*` wrappers entirely after this PR lands.

gosharplite:
I fully agree with the latest assessment and the decision to split the work.

1. **Testability vs. Performance (`isDebugEnabled`)**: The assessment is completely correct. Initializing an environment variable into a package-level global at startup breaks testability by ignoring `t.Setenv()`. Furthermore, since `os.Getenv` simply reads from an in-memory slice initialized at startup (rather than making a syscall), the performance cost is negligible. Favoring testability over premature optimization is the right call.

2. **Scope Management**: Agreeing with the logging anti-pattern but moving the behavioral change to a separate follow-up PR (#256) is an excellent engineering practice. Keeping this PR strictly scoped to the naming refactor ensures a clean commit history and limits the blast radius, while the architectural fix gets the dedicated attention it deserves in the next cycle.

Merging this PR as-is and tackling the `slog` changes in #256 is the optimal path forward.

PR #259, opened 2026-05-08
refactor(config): consolidate debug guard into debugLog helper

### LINKED ISSUE #253: [TECHNICAL DEBT] Consolidate fragmented debug-logging helpers in config.go
## Problem

`internal/infrastructure/config/config.go` defines 4 standalone debug-logging helper functions, each repeating the same guard pattern:

```go
func debugLogFileStatus(path string) {
    if !isDebug() { return }
    // ... logging logic
}

func debugLogRawContent(path string, data []byte) {
    if !isDebug() { return }
    // ... logging logic
}

func debugLogViperKeys(v *viper.Viper) {
    if !isDebug() { return }
    // ... logging logic
}

func debugLogModels(cfg *domain_config.Config) {
    if !isDebug() { return }
    // ... logging logic
}
```

Each function duplicates the `if !isDebug() { return }` guard. Individually small, but collectively they represent fragmented debug infrastructure that bloats the file and makes adding new debug logging more ceremony than it needs to be.

## Proposed Fix

Consolidate behind a single debug-logger wrapper. Options:

1. **Minimal**: Extract the guard into a `debugLog(func())` helper that only invokes the callback when debug is enabled
2. **Full**: Create a `configDebugLogger` struct that holds a `*slog.Logger` (or nil when disabled) and methods for each domain concept

The minimal approach is preferred — it keeps the change surgical and avoids over-engineering.

## Related

- [ ] #251 — broader `bug` → `debug` naming cleanup (coordinate with this)

thptcnec:
Closes #253.

## Summary
Replace 4 inline `slog.Default().Enabled(context.Background(), slog.LevelDebug)` guard blocks with a single package-level `debugLog(func())` helper.

## Changes
- **New**: `debugLog(fn func())` — invokes callback only when debug is enabled
- **Refactored 4 call sites**: `load()`, `readConfigFile()` (×2), `validationLogger()`
- `validationLogger()` uses closure-captured variable for its return value
- All unconditional `slog.Debug` calls left untouched

## Verification
- 21/21 tests passing
- `go vet` clean
- Only `config.go` modified

### DIFF
--- a/internal/infrastructure/config/config.go
+++ b/internal/infrastructure/config/config.go
@@ -19,6 +19,11 @@ import (
 	"github.com/spf13/viper"
 )
 
+// isDebug reports whether the default slog logger is at Debug level.
+func isDebug() bool {
+	return slog.Default().Enabled(context.Background(), slog.LevelDebug)
+}
+
 // YAMLConfigLoader implements domain_config.ConfigLoader.
 type YAMLConfigLoader struct {
 	Finder domain_config.ConfigFinder
@@ -53,7 +58,7 @@ func load(path string) (*domain_config.Config, error) {
 		return nil, err
 	}
 
-	if slog.Default().Enabled(context.Background(), slog.LevelDebug) {
+	if isDebug() {
 		slog.Debug("cfg.Models count", slog.Int("count", len(cfg.Models)))
 		for k, v := range cfg.Models {
 			slog.Debug("model detail",
@@ -82,7 +87,7 @@ func load(path string) (*domain_config.Config, error) {
 // slog logger; otherwise it discards warn output to keep test output
 // quiet. Hard errors are returned via the error path regardless.
 func validationLogger() *slog.Logger {
-	if slog.Default().Enabled(context.Background(), slog.LevelDebug) {
+	if isDebug() {
 		return slog.Default()
 	}
 	return slog.New(slog.NewTextHandler(io.Discard, &slog.HandlerOptions{Level: slog.LevelWarn}))
@@ -126,7 +131,7 @@ func readConfigFile(v *viper.Viper, path string) error {
 		return err
 	}
 
-	if slog.Default().Enabled(context.Background(), slog.LevelDebug) {
+	if isDebug() {
 		slog.Debug("raw content", slog.String("content", string(data[:min(len(data), 1000)])))
 	}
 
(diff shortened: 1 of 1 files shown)


thptcnec:
### Architectural & Idiomatic Go Review

While the intent to consolidate repetitive code and DRY up the debug checks is appreciated, the approach relying on closures to execute conditional blocks leads to unidiomatic Go code and reduces readability.

**[TECHNICAL DEBT] Unidiomatic use of Closures for Control Flow**
The refactoring in `validationLogger()` introduces unnecessary variable capturing to extract a return value from inside the callback closure. This is a known anti-pattern that harms clarity.

*Current (from PR):*
```go
func validationLogger() *slog.Logger {
	var logger *slog.Logger
	debugLog(func() {
		logger = slog.Default()
	})
	if logger != nil {
		return logger
	}
	return slog.New(...)
}
```

*Idiomatic:*
```go
func validationLogger() *slog.Logger {
	if slog.Default().Enabled(context.Background(), slog.LevelDebug) {
		return slog.Default()
	}
	return slog.New(...)
}
```

**[REFACTOR] Overly Clever Abstraction (Clear > Clever)**
Go favors simplicity over cleverness. Replacing a standard `if` conditional with a callback-wrapper `debugLog(func() { ... })` harms readability and introduces unnecessary anonymous function allocations. The standard `log/slog` package natively expects developers to use `if logger.Enabled(...)` guards around expensive loops or allocations.

**Recommendation:**
Please revert the introduction of the `debugLog(func())` callback wrapper. If consolidating the `slog.Default().Enabled(...)` condition is strictly required, a much better and idiomatic approach is to create a simple helper that returns a boolean:

```go
func isDebug() bool {
    return slog.Default().Enabled(context.Background(), slog.LevelDebug)
}

// Usage:
if isDebug() {
    // ...
}
```

thptcnec:
Addressed: replaced `debugLog(func())` with `isDebug() bool` helper. Call sites now use standard `if isDebug() { ... }` guards. `validationLogger()` restored to clean early-return shape. No closure allocations. PTAL.

thptcnec:
I have reviewed the updated changes and completely agree with your latest comment and implementation. The adoption of the `isDebug() bool` helper perfectly balances the need for DRY code with Go's philosophy of "Clear is better than clever". By reverting to standard `if isDebug() { ... }` blocks, you have: 1. Eliminated the closure anti-pattern and unnecessary variable capturing in `validationLogger()`. 2. Avoided anonymous function allocations. 3. Restored idiomatic, highly readable Go control flow. The architectural debt has been resolved. The PR looks great.

PR #268, opened 2026-05-09
test(clock): add coverage for Since, NewTicker, and realTicker.C (fixes #266)

### LINKED ISSUE #266: pkg/clock: add tests for NewTicker, Since, and realTicker.C (57% → 100% coverage)
## Summary

`internal/pkg/clock` is at 57.1% coverage despite being imported by 8+ core packages (`agent`, `orchestrator`, `session`, `ui`, `di`, `logging`, `cli`). Three functions are completely untested:

```
NewTicker   0.0%
Since       0.0%
C (ticker)  0.0%
```

`Now`, `Sleep`, `After`, and `Jitter` are already at 100%.

### 1. RealClock.NewTicker(d Duration) Ticker
Verify it returns a ticker that fires at the expected interval.

### 2. realTicker.C() <-chan time.Time
Verify the channel is non-nil and returns the same underlying channel across calls.

### 3. RealClock.Since(t Time) Duration
Verify it returns a non-negative duration, and that calling it after a short sleep produces a plausible value.

## Acceptance Criteria

- [ ] NewTicker test added
- [ ] realTicker.C test added  
- [ ] Since test added
- [ ] Package coverage reaches >= 90%
- [ ] `go test -race ./internal/pkg/clock/...` passes

## Impact

This package is a leaf-node utility with high fan-out. Low coverage here means any regression in time/ticker behavior propagates silently to the orchestrator, session management, and UI layers.

## Related

Per-function coverage: `go tool cover -func=coverage.out | grep clock`

thptcnec:
## Summary

Adds three tests to `internal/pkg/clock`, raising package coverage from 57.1% to 100%.

### New Tests

| Test | Function Covered | What It Verifies |
|---|---|---|
| `TestRealClock_Since` | `Since` | Non-negative duration after sleep |
| `TestRealTicker_C` | `realTicker.C` | Channel identity (non-nil, same across calls) |
| `TestRealClock_NewTicker` | `NewTicker` | Ticker fires within 200ms timeout |

### Verification

- `go test -race -count=3 ./internal/pkg/clock/` — **PASS**
- `go tool cover -func` — **100%** (7/7 functions)
- No goroutine leaks (all tickers stopped via `t.Cleanup`)
- Zero production code changes

Fixes #266

### DIFF
--- a/internal/pkg/clock/clock_test.go
+++ b/internal/pkg/clock/clock_test.go
@@ -50,3 +50,55 @@ func TestRealClock_Jitter(t *testing.T) {
 		}
 	}
 }
+
+// TestRealClock_Since verifies that RealClock.Since returns a non-negative duration
+// that is at least as long as a known sleep interval.
+func TestRealClock_Since(t *testing.T) {
+	t.Parallel()
+	c := RealClock{}
+	t0 := c.Now()
+	c.Sleep(10 * time.Millisecond)
+	d := c.Since(t0)
+
+	if d < 0 {
+		t.Errorf("RealClock.Since() = %v, want non-negative", d)
+	}
+	if d < 10*time.Millisecond {
+		t.Errorf("RealClock.Since() = %v, want at least 10ms", d)
+	}
+}
+
+// TestRealTicker_C verifies that realTicker.C returns a non-nil channel and that
+// repeated calls return the same underlying channel.
+func TestRealTicker_C(t *testing.T) {
+	c := RealClock{}
+	tk := c.NewTicker(time.Hour)
+	t.Cleanup(tk.Stop)
+
+	ch1 := tk.C()
+	ch2 := tk.C()
+
+	if ch1 == nil {
+		t.Error("realTicker.C() returned nil channel")
+	}
+	if ch1 != ch2 {
+		t.Error("realTicker.C() returned different channels across calls — expected same channel")
+	}
+}
+
+// TestRealClock_NewTicker verifies that RealClock.NewTicker returns a Ticker
+// whose channel fires at least once within a timeout window.
+func TestRealClock_NewTicker(t *testing.T) {
+	c := RealClock{}
+	tk := c.NewTicker(10 * time.Millisecond)
+	t.Cleanup(tk.Stop)
+
+	select {
+	case tick := <-tk.C():
+		if tick.IsZero() {
+			t.Error("NewTicker channel delivered zero time")
+		}
+	case <-time.After(200 * time.Millisecond):
+		t.Fatal("NewTicker did not fire within 200ms timeout")
+	}
+}


thptcnec (commented):
### PR 268 Review (`internal/pkg/clock/clock_test.go`)

**1. [REFACTOR] Flaky Test Assertions (OS Scheduler Dependence)**
Asserting exact timing bounds on `time.Sleep` (e.g., `d < 10*time.Millisecond`) is a well-known Go anti-pattern and a frequent source of CI flakiness. The OS scheduler (particularly on Windows or heavily loaded CI runners) does not guarantee exact millisecond precision, and `time.Sleep` can occasionally return a fraction of a millisecond early.

**Current:**
```go
if d < 10*time.Millisecond {
    t.Errorf("RealClock.Since() = %v, want at least 10ms", d)
}
```

**Robust/Scalable:**
Avoid testing the precision of Go's standard library or the OS scheduler. Relax the assertion to provide a small tolerance buffer, ensuring the test only catches actual logic failures.
```go
if d < 8*time.Millisecond { // Leave a buffer for OS timer quirks
    t.Errorf("RealClock.Since() = %v, want at least ~10ms", d)
}
```

**2. [TECHNICAL DEBT] Inconsistent Concurrency (`t.Parallel()`)**
`TestRealClock_Since` correctly calls `t.Parallel()`, but this is missing in `TestRealTicker_C` and `TestRealClock_NewTicker`.
**Action:** Add `t.Parallel()` as the first line in both of these tests. This maintains consistency, enforces statelessness of the `RealClock` component, and optimizes test execution time.

**3. [ARCHITECTURAL NOTE] Testing Standard Library Wrappers**
These tests effectively validate the behavior of Go's native `time` package. While acceptable—and often necessary—for achieving strict code coverage on the `RealClock` wrapper, the core value of this code is the `Clock` interface itself. The design correctly implements **Dependency Injection**, allowing higher-level domain logic to decouple from system time and utilize deterministic mock clocks during unit testing.

**Recommendation:** Relax the sleep duration assertion to prevent future flaky builds and add the missing `t.Parallel()` calls before merging.

thptcnec:
## Architect Review

### Point 1 — Flaky Test Assertions → **Disagree**

The reviewer suggests relaxing `d < 10*time.Millisecond` to `d < 8*time.Millisecond`, claiming `time.Sleep` can return early. This is incorrect. The Go spec guarantees:

> *"Sleep pauses the current goroutine for **at least** the duration d."*

`time.Sleep(10ms)` will never return before 10ms of wall-clock time have elapsed. Furthermore, `time.Since` uses monotonic clock readings, which are immune to wall-clock adjustments. The assertion `d < 10*time.Millisecond` is the correct lower bound — it tests that the wrapper didn't accidentally return zero or negative.

If CI flakiness actually materializes (unlikely), the fix is to increase the sleep duration (e.g., 50ms), not to relax the assertion below the guaranteed minimum.

**Verdict: Keep as-is.**

---

### Point 2 — Missing `t.Parallel()` → **Agree**

`t.Parallel()` should be added to `TestRealTicker_C` and `TestRealClock_NewTicker`. Go 1.14+ guarantees `t.Cleanup` runs correctly with parallel subtests — the cleanup is registered before the parallel barrier. Consistency with the other 5 tests is the right call.

**Action: Add `t.Parallel()` as the first line in both tests.**

---

### Point 3 — Testing Standard Library Wrappers → **Agree, No Action**

Accurate architectural observation. The real value is the `Clock` interface and mock-based DI in higher layers. These tests are a coverage backstop. No code change needed.

thptcnec:
Thank you for the pushback on Point 1—you are absolutely correct. 

My initial concern about OS scheduler flakiness returning *early* was misplaced. As you rightly pointed out, the Go specification strictly guarantees that `time.Sleep(d)` will pause the goroutine for **at least** duration `d`. The imprecision of OS schedulers (such as the Windows 15.6ms timer resolution) manifests as sleeping for *longer* than requested, never shorter. Furthermore, Go's use of monotonic clocks for `time.Since` completely eliminates wall-clock drift issues. Keeping the exact assertion `d < 10*time.Millisecond` is mathematically sound and correct.

I am glad we are completely aligned on Points 2 (adding `t.Parallel()`) and 3 (the value of DI and interface-driven design). 

Great catch, and I appreciate the rigorous review! Please proceed with the `t.Parallel()` additions, and this will be good to merge.

thptcnec:
## Final Sign-Off

`t.Parallel()` added to `TestRealTicker_C` and `TestRealClock_NewTicker` in `fc891ed9`. All 7 tests now call `t.Parallel()` consistently.

### Final Verification

- `go test -race -count=3 ./internal/pkg/clock/` — **PASS**
- Coverage — **100%** (7/7 functions)
- No goroutine leaks

All three review points resolved:

| Point | Resolution |
|---|---|
| #1 Relax assertion to 8ms | Rejected — `time.Sleep` guarantees minimum duration |
| #2 Add `t.Parallel()` | Done |
| #3 Stdlib wrapper note | Acknowledged — no action |

Ready to merge. :shipit:

thptcnec:
**LGTM!**

The implementation of `t.Parallel()` across the suite looks great, and I highly appreciate the rigorous validation using `-race` and `-count=3` to ensure the concurrency changes didn't introduce any subtle regressions or goroutine leaks. 

All architectural and technical debt points have been successfully resolved. 

Approved and ready to merge! :rocket:

PR #271, opened 2026-05-09
refactor: decompose TestHistoryNavigationFlags (complexity 17 → ≤8)

### LINKED ISSUE #270: Decompose TestHistoryNavigationFlags (Complexity 17 → ≤10)
## Problem

`TestHistoryNavigationFlags` in `tests/e2e/history_flags_test.go:36` has **cyclomatic complexity 17** — the highest in the entire codebase. This follows the same pattern addressed in #267 (`refactor: decompose 3 high-complexity test functions`).

## Current Structure

```go
func TestHistoryNavigationFlags(t *testing.T) {
    // SHARED SETUP (complexity ~8):
    // 1. httptest server
    // 2. temp dir + config + env vars
    // 3. for each of 3 prompts:
    //      runCommandWithEnv + forceReconcileHistory + time.Sleep
    //
    // SUBTESTS (complexity ~3-4 each):
    t.Run("LastNMessages", ...)
    t.Run("GoBack", ...)        // runs CLI -b + CLI -l + assertions
    t.Run("Retry", ...)         // runs CLI --retry + reads history file + assertions
}
```

The shared setup loop (3 sequential CLI invocations with reconciliation/sleep) and the multi-step subtests combine to produce 17 paths.

## Proposed Refactoring

Extract a `newHistoryNavEnv` helper (similar to patterns already in the codebase) that encapsulates setup, then ensure each subtest is self-contained:

```go
type historyNavEnv struct {
    env       []string
    configPath string
    histPath  string
    server    *httptest.Server
}

func newHistoryNavEnv(t *testing.T) *historyNavEnv { ... }

func TestHistoryNavigationFlags(t *testing.T) {
    t.Run("LastNMessages", func(t *testing.T) { ... })
    t.Run("GoBack", func(t *testing.T) { ... })
    t.Run("Retry", func(t *testing.T) { ... })
}
```

## Acceptance Criteria

- [ ] `gocyclo` reports complexity ≤ 10 for `TestHistoryNavigationFlags`
- [ ] All existing assertions pass unchanged
- [ ] `TestHistoryOnlyExit` (same file) is not affected
- [ ] Follows the decomposition pattern established in #267

## References

- Previous work: `8854fc55` (`refactor: decompose 3 high-complexity test functions`)
- Related: #267

thptcnec:
## Summary

Decomposes `TestHistoryNavigationFlags` following the pattern established in #267 (`8854fc55`).

### Changes

| | Before | After |
|---|---|---|
| **Complexity** | 17 (highest in codebase) | ≤ 8 across all functions |
| **Functions** | 1 wrapper + 3 closures | 1 helper + 3 standalone tests |

- Extracted `historyNavEnv` struct + `newHistoryNavEnv` constructor for shared setup
- Split into `TestHistoryNavigationFlags_LastNMessages` (5), `_GoBack` (7), `_Retry` (8)
- Replaced `defer server.Close()` with `t.Cleanup(server.Close)`
- Fixed hidden ordering dependency: `Retry` now explicitly performs `-b 1` before `--retry` instead of relying on `GoBack` running first in the same parent

### Verification

- All 4 tests pass in parallel: `go test ./tests/e2e/... -run 'TestHistoryNavigationFlags|TestHistoryOnlyExit' -v`
- `gocyclo` reports max complexity 8 (target ≤ 10)
- `TestHistoryOnlyExit` unchanged

Closes #270

### DIFF
--- a/tests/e2e/history_flags_test.go
+++ b/tests/e2e/history_flags_test.go
@@ -18,6 +18,57 @@ import (
 	"github.com/gosharplite/tell-me-go/internal/infrastructure/persistence"
 )
 
+// historyNavEnv holds the shared test environment for history navigation E2E tests.
+type historyNavEnv struct {
+	env        []string
+	configPath string
+	histPath   string
+	server     *httptest.Server
+}
+
+// newHistoryNavEnv creates a mock server, temp config, env vars, and populates
+// the session history with 3 sequential prompts ("Message 1", "Message 2", "Message 3").
+func newHistoryNavEnv(t *testing.T) *historyNavEnv {
+	t.Helper()
+
+	// 1. Setup Mock Server
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		resp := createTextResponse("google", "Response to your prompt")
+		_, _ = fmt.Fprint(w, resp)
+	}))
+	t.Cleanup(server.Close)
+
+	// 2. Setup Environment
+	homeDir := t.TempDir()
+	configPath := createTempConfig(t, "google", server.URL)
+	env := []string{
+		"TELL_ME_HOME=" + homeDir,
+		"TELL_ME_MOCK_URL=" + server.URL,
+		"TELL_ME_MOCK_ANSWER=y",
+	}
+
+	histPath := filepath.Join(homeDir, "output", "assistant", "history.jsonl")
+
+	// 3. Populate session history with 3 distinct prompts
+	prompts := []string{"Message 1", "Message 2", "Message 3"}
+	for _, p := range prompts {
+		_, _, err := runCommandWithEnv(env, "", "-c="+configPath, p)
+		if err != nil {
+			t.Fatalf("Failed to send prompt %q: %v", p, err)
+		}
+		forceReconcileHistory(t, histPath)
+		time.Sleep(1000 * time.Millisecond)
+	}
+
+	return &historyNavEnv{
+		env:        env,
+		configPath: configPath,
+		histPath:   histPath,
+		server:     server,
+	}
+}
+
 // forceReconcileHistory ensures the history file is fully flushed and visible on Windows.
 func forceReconcileHistory(t *testing.T, path string) {
 	t.Helper()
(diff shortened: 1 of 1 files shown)


thptcnec (commented):
Here is the architectural review of PR 271 based on the system design, scalability, and maintainability standards.

### 1. Analysis
The PR successfully refactors `tests/e2e/history_flags_test.go` to uncouple a single sequential test with shared state into three independent, parallel tests (`LastNMessages`, `GoBack`, `Retry`). A shared environment fixture builder (`newHistoryNavEnv`) was introduced. While this improves test isolation (eliminating side-effect dependencies between subtests), it inadvertently multiplies the execution of expensive setup code and exposes brittle test preconditions.

### 2. Findings

#### [TECHNICAL DEBT] - O(N) Performance Degradation via Hardcoded Sleeps
- **Current Design:** `newHistoryNavEnv` executes the CLI 3 times and uses `time.Sleep(1000 * time.Millisecond)` after each execution to mitigate Windows file locks. Since this setup is now called independently by 3 parallel tests, it forces 9 total seconds of sleeps (or 3 seconds of blocking thread time per parallel test). Synchronous blocking in hot paths (including tests) is an anti-pattern.
- **Scalable Design:** Remove the hardcoded `time.Sleep`. 
  - **Option A (Preferred):** Since this is an E2E test, bootstrap the history state instantly by writing the initial `.jsonl` state directly to disk using `history.NewManager(...).Save()` or standard `os.WriteFile()`. This bypasses the heavy CLI process overhead and file-lock race conditions entirely during setup.
  - **Option B:** If CLI execution must be used for setup, replace `time.Sleep` with a polling mechanism (e.g., `require.Eventually`) that checks if the history file has been updated to the expected state before proceeding.

#### [REFACTOR] - Brittle Preconditions Mimicking Legacy State
- **Current Design:** `TestHistoryNavigationFlags_Retry` intentionally executes a `GoBack` (`-b 1`) CLI command solely to recreate the arbitrary state left behind by the old sequential test run (`// Roll back 1 turn first, matching the precondition the original shared-state test relied on...`).
- **Scalable Design:** Tests should be explicit about their domain scenarios, not recreate legacy execution artifacts. If the goal is to test `--retry`, run it against the latest state directly without the artificial `-b 1` step. If the specific business requirement is to ensure "Retry works immediately after Undo", rename the test to `TestHistoryNavigationFlags_RetryAfterGoBack` so the intent is obvious.

#### [REFACTOR] - Encapsulation Leaks in Test Fixture
- **Current Design:** `newHistoryNavEnv` returns a struct containing `server *httptest.Server`. However, none of the individual tests interact with the `server` field directly. 
- **Scalable Design:** Since `newHistoryNavEnv` correctly handles the server's lifecycle via `t.Cleanup(server.Close)`, the `server` pointer should not be leaked into the returned `historyNavEnv` struct. Remove the unused field to adhere to the Interface Segregation Principle (expose only what is needed).

### Recommended Action for the Coder:
1. Replace the `time.Sleep` loop in `newHistoryNavEnv` with direct file injection to instantly seed the `history.jsonl` file.
2. Remove the `-b 1` execution from the `Retry` test, adjusting assertions to expect `Message 3` to be retried (or explicitly rename the test if the GoBack behavior is the actual target).
3. Remove the `server` field from the `historyNavEnv` struct.

thptcnec:
Thanks for the thorough review. Here is my assessment of each finding:

---

### Finding 1: `[TECHNICAL DEBT]` — Hardcoded Sleeps

**Partially agree.** The `time.Sleep` calls are indeed a brittle anti-pattern, and the refactoring does amplify the cost (3 tests × 3 sleeps = ~9s wall time). However, this is **pre-existing debt** — the sleeps were in the original monolithic test long before this PR. The refactoring didn't create the problem; it just made the multiplication visible.

**On Option A (direct file injection):** This is the right long-term direction, but it changes *what* the setup validates. The original 3-prompt CLI loop served as an implicit integration check that the CLI can write sequential prompts to history. Direct `history.NewManager(...).Append()` bypasses that.

**Recommendation:** Accept this as a follow-up issue (#272). Replace the 3-prompt CLI loop in `newHistoryNavEnv` with direct file injection for speed, and spin out a dedicated test like `TestHistoryPopulationViaCLI` if that integration coverage is needed. Not a blocker for this PR.

---

### Finding 2: `[REFACTOR]` — Brittle Preconditions

**Disagree.** Removing `-b 1` and asserting that `--retry` retries "Message 3" would be a **semantically different test** from what the original code verified. The original assertion was:

> After rolling back past Message 3, `--retry` retries Message 2.

That is a real integration scenario — "undo, then retry." Testing `--retry` on the last message without the rollback is a weaker, less interesting case. The precondition is not a "legacy execution artifact"; it is the scenario under test.

**The rename suggestion (`_RetryAfterGoBack`):** Reasonable but unnecessary. `_Retry` already communicates the flag under test. The `-b 1` is a precondition, not the subject of the test. The inline comment added during the fix is sufficient documentation.

**Verdict:** Keep as-is.

---

### Finding 3: `[REFACTOR]` — Unused `server` Field

**Fully agree.** The `server` field is never accessed by any caller. Its lifecycle is already managed by `t.Cleanup(server.Close)` inside the constructor. Leaking it into the struct violates "expose only what is needed."

**Recommendation:** Remove the `server` field from `historyNavEnv` and from the return literal in `newHistoryNavEnv`. This is a one-line cleanup on each side.

---

### Summary

| # | Verdict | Action |
|---|---------|--------|
| 1 | Agree in principle, pre-existing | → Follow-up issue; not a PR blocker |
| 2 | Disagree | → No change |
| 3 | Agree | → Remove `server` field now |

I will push a commit to address Finding 3. Finding 1 should be tracked separately.

thptcnec:
I fully agree with this assessment and the proposed path forward. 

Your reasoning is pragmatic and aligns well with standard architectural practices:

1. **Deferring the Technical Debt:** You are correct that the `time.Sleep` is pre-existing debt. Blocking a structural refactor on pre-existing performance issues leads to scope creep. Deferring this to a follow-up issue (#272) is the right Agile approach. Your insight regarding the implicit CLI integration coverage is spot on—spinning that out into a dedicated test before switching to direct file injection is a highly mature testing strategy.
2. **Preserving the Retry Scenario:** Thank you for clarifying the domain intent. Since the test is explicitly designed to verify the "Undo -> Retry" sequence rather than just a simple retry, retaining the `-b 1` precondition is correct and prevents weakening the test suite's coverage.
3. **Encapsulation Leak:** Glad we are aligned on removing the `server` field to enforce the Interface Segregation Principle.

Looking forward to the commit for Finding 3. Once that is pushed, this PR is good to go from an architectural perspective.

thptcnec:
Finding 3 resolved in 34f3965: the `server` field has been removed from `historyNavEnv` and from the constructor return literal.

All tests pass, complexity ≤ 8 across all functions. PR is ready for final review / merge.

thptcnec:
I have verified the changes in `34f3965`. 

The unused `server` field has been successfully removed, cleanly resolving the encapsulation leak. The complexity limits have been maintained, and the uncoupling of the parallel tests significantly improves the robustness of the test suite. 

With the remaining technical debt properly triaged to #272, this PR meets all architectural requirements. Excellent work keeping the scope tight while improving code hygiene.

**Verdict:** Good to merge.

PR #277, opened 2026-05-09
test: cover empty-path defensive guard in StreamTurnsLog

thptcnec:
## Summary

Makes the empty-path defensive guard at `service.go:214` testable and adds the missing test.

## Changes

- **`service.go`**: Added `PathResolveFunc` package variable and `resolvePaths` struct field for DI of path resolution. `StreamTurnsLog` now calls `s.resolvePaths()` instead of `persistence.ResolvePaths` directly.
- **`service_test.go`**: Renamed existing test to `TestChatService_StreamTurnsLog_EmptyMode` (it tested empty-mode→default fallback). Added `TestChatService_StreamTurnsLog_EmptyPath` which injects a stub returning zero-value `Paths`, asserting the guard returns `"turns log path not available"`.

## Verification

```
go test ./internal/agent/... -count=1 -run "StreamTurnsLog"
```

All 8 StreamTurnsLog-related tests pass, including the new guard test.

Closes #274

### DIFF
--- a/internal/agent/service.go
+++ b/internal/agent/service.go
@@ -21,11 +21,6 @@ import (
 	"github.com/gosharplite/tell-me-go/internal/pkg/ioutils"
 )
 
-// PathResolveFunc is the function used to resolve session filesystem paths.
-// It is exported to allow tests to exercise the empty-path defensive guard
-// in StreamTurnsLog. Production code must not override this.
-var PathResolveFunc func(homeDir, mode string) *persistence.Paths = persistence.ResolvePaths
-
 // SessionLifecycleManager defines the interface for building and finalizing sessions.
 type SessionLifecycleManager interface {
 	BuildSessionDependencies(ctx context.Context, cfg *domain_config.Config, configPath string, newSession bool, capturer CapturerInteractor) (ports.SessionDependencies, ports.HistoryManager, func(context.Context) error, error)
@@ -35,6 +30,17 @@ type SessionLifecycleManager interface {
 // LogFileOpener defines the minimal interface required to open session log files.
 type LogFileOpener interface {
 	Open(ctx context.Context, name string) (persistence.File, error)
+}
+
+// ChatServiceOption configures a chatService during construction.
+type ChatServiceOption func(*chatService)
+
+// WithPathResolver overrides the default path resolution function.
+// The default is persistence.ResolvePaths.
+func WithPathResolver(resolver func(homeDir, mode string) *persistence.Paths) ChatServiceOption {
+	return func(s *chatService) {
+		s.resolvePaths = resolver
+	}
 }
 
 type chatService struct {
@@ -67,6 +73,7 @@ func NewChatService(
 	historyRenderer ports.HistoryRenderer,
 	historyBrowser ports.HistoryBrowser,
 	logOpener LogFileOpener,
+	opts ...ChatServiceOption,
 ) ChatService {
 	svc := &chatService{
 		HomeDir:          homeDir,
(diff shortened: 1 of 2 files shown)


thptcnec (commented):
Based on the architectural review of PR 277, here are the findings categorized by impact.

### [ARCHITECTURAL BLOCKER]
**Global Mutable State & Thread-Unsafe Mocking**
The PR exports a global variable `PathResolveFunc` specifically to override it during unit tests. This introduces global mutable state, which breaks test isolation and prevents safe concurrent test execution (e.g., using `t.Parallel()`). Overriding globals in test runs can cause unpredictable race conditions across your test suite. 

### [TECHNICAL DEBT]
**Leaking Test Concerns into Public API**
Exporting a variable with the comment *"Production code must not override this"* violates encapsulation. Production API design should not be compromised to facilitate code coverage. If a dependency needs to be stubbed, it should be passed via explicit Dependency Injection (DI) rather than exposing internals globally.

### [REFACTOR]
**Dependency Injection via Functional Options**
To test the defensive guard without introducing global variables or leaking test logic, you should pass the dependency through the constructor. If you do not want to bloat the existing parameter list of `NewChatService`, use the **Functional Options Pattern** or use a test-only setter in an `export_test.go` file.

**Current (Global State Anti-Pattern):**
```go
var PathResolveFunc = persistence.ResolvePaths // Global mutable state

func NewChatService(...) ChatService {
	svc := &chatService{ /* ... */ }
	svc.resolvePaths = PathResolveFunc
	return svc
}

// In test:
agent.PathResolveFunc = mockFunc
defer restore() // Flaky if tests run concurrently
```

**Scalable (Dependency Injection):**
```go
// 1. Define an Option type
type ChatServiceOption func(*chatService)

// 2. Provide a test-friendly injector
func WithPathResolver(resolver func(homeDir, mode string) *persistence.Paths) ChatServiceOption {
	return func(s *chatService) {
		s.resolvePaths = resolver
	}
}

// 3. Apply options in constructor
func NewChatService(..., opts ...ChatServiceOption) ChatService {
	svc := &chatService{
		// ...
		resolvePaths: persistence.ResolvePaths, // Default safe state
	}
	for _, opt := range opts {
		opt(svc)
	}
	return svc
}

// 4. In test: Clean execution without global side-effects
service := agent.NewChatService(..., agent.WithPathResolver(func(h, m string) *persistence.Paths {
    return &persistence.Paths{} 
}))
```

thptcnec:
Thanks for the thorough review — all three points were spot-on. I've refactored accordingly:

**Changes pushed (e095e60):**

1. ✅ **Removed `PathResolveFunc` global variable** — no more mutable global state.

2. ✅ **Added Functional Options pattern:**
   - `ChatServiceOption` type
   - `WithPathResolver(resolver)` constructor
   - `NewChatService` accepts `opts ...ChatServiceOption` (variadic — zero production call-site changes)

3. ✅ **Test updated** — uses `agent.WithPathResolver(emptyPathResolver)` as the final argument to `NewChatService`. No global manipulation, no `t.Cleanup` restore needed, fully `t.Parallel()` safe.

```
go test ./internal/agent/... -count=1 -run "StreamTurnsLog"
# 8 tests, all PASS
```

Ready for re-review.

thptcnec:
I fully agree with the latest updates in the PR. The refactor accurately implements the architectural feedback:

1. ✅ **Global Mutable State Eliminated:** Removing `PathResolveFunc` resolves the test isolation and concurrency risk (the `[ARCHITECTURAL BLOCKER]`).
2. ✅ **Proper Encapsulation:** Making `resolvePaths` unexported and utilizing the **Functional Options Pattern** (`ChatServiceOption` / `WithPathResolver`) successfully bridges the need for testability without polluting the public API or demanding breaking changes to production `NewChatService` call-sites.
3. ✅ **Thread-Safe Testing:** The new `TestChatService_StreamTurnsLog_EmptyPath` uses the option correctly, allowing test execution to be fully deterministic and safe for `t.Parallel()`.

The refactor demonstrates excellent alignment with Go idioms for Dependency Injection. Great job on the execution!

**VERDICT: APPROVED**

PR #291, opened 2026-05-10
refactor: consolidate SimpleEventBus worker initialization into startSubscriberLoop

thptcnec:
## Summary

Closes #286 — eliminates DRY violation where worker-spawning boilerplate (`workerWG.Add`, goroutine, panic recovery, `subscriberLoop` call) was duplicated in both `Listen` and `startSubscriberLoop`.

## Changes

| File | Change |
|---|---|
| `event_bus_lifecycle.go` | Replaced 14-line inline worker-spawn block in `Listen` with `b.startSubscriberLoop(w)`; removed unused `log/slog` import |
| `event_bus_pubsub.go` | Captured `b.listenCtx` into local `ctx` before goroutine spawn to fix latent data race; unified log message |
| `event_bus_pubsub_test.go` | Updated 2 assertions + 1 error message to match unified log string |

## Architecture

- `startSubscriberLoop` is now the **single source of truth** for spawning subscriber worker goroutines.
- Every worker gets consistent panic recovery behavior regardless of whether it was started at `Listen` time or added dynamically.
- Context capture is race-free: `b.listenCtx` is snapshotted into a stack-local before the goroutine launches, eliminating the write-after-read race with `Listen`'s shutdown path.

## Verification

```
go build ./internal/domain/events/...   → PASS
go test -race ./internal/domain/events/... → PASS (2.282s)
go vet ./internal/domain/events/...     → PASS
```

### DIFF
--- a/internal/domain/events/event_bus_lifecycle.go
+++ b/internal/domain/events/event_bus_lifecycle.go
@@ -5,7 +5,6 @@ package events
 
 import (
 	"context"
-	"log/slog"
 	"runtime/debug"
 )
 
@@ -193,18 +192,7 @@ func (b *SimpleEventBus) Listen(ctx context.Context) error {
 	wrappers = append(wrappers, b.globalSubscribers...)
 
 	for _, w := range wrappers {
-		w := w
-		b.workerWG.Add(1)
-		go func() {
-			defer func() {
-				if r := recover(); r != nil {
-					b.getLogger().ErrorContext(listenCtx, "panic in event bus subscriber loop",
-						slog.Any("error", r),
-						slog.String("stack", string(debug.Stack())))
-				}
-			}()
-			_ = b.subscriberLoop(listenCtx, w)
-		}()
+		b.startSubscriberLoop(w)
 	}
 
 	// Signal that the listener is fully initialized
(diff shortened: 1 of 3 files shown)


thptcnec:
## 🏛️ Architectural Review

**Verdict: ✅ APPROVE** — Clean DRY refactor that also fixes a latent data race.

### Strengths

1. **Single source of truth** — `startSubscriberLoop` is now the sole spawn point for all 5 subscription paths (`Listen`, `Subscribe`, `SubscribeGlobal`, `SubscribeSubscriber`, funcSubscriber). Drift risk eliminated.
2. **Race fix (under-advertised!)** — Capturing `ctx := b.listenCtx` under the caller's `b.mu` eliminates a real TOCTOU race against `Listen`'s shutdown path (`b.listenCtx = nil` at line 210). The previous goroutine read `b.listenCtx` unlocked.
3. **Log unification** — Same recovery semantics → same log key. Operators no longer need two queries.
4. **Lock discipline verified** — All 5 callers correctly hold `b.mu` per the documented contract.

### Suggestions (Non-blocking)

🟡 **Elevate the race fix in the PR description.** It's currently buried under "unified log message" but is arguably more important than the DRY refactor itself. Suggested addition:

> ## Bug Fix
> `startSubscriberLoop` previously read `b.listenCtx` from inside the spawned goroutine without holding `b.mu`, racing with `Listen`'s shutdown write `b.listenCtx = nil`. Now snapshotted into a stack-local under the caller's lock.

### Follow-up Filed

- #292 — Rename `startSubscriberLoop` → `startSubscriberLoopLocked` to surface the lock contract per Go convention (`runtime`/`sync` style). Pure naming hygiene; not blocking this PR.

### Recommendation

**Merge as-is** (or after the description tweak above). CI clean, mergeable state CLEAN, +7/−18 net, tests passing under `-race`.

PR #296, opened 2026-05-10
fix: replace filepath.Rel with strings.HasPrefix in isInSearchPath

### LINKED ISSUE #295: SearchSymbols slow on CI: filepath.Rel in hot path causes TestIndexer_Scaling timeout
## Problem

`TestIndexer_Scaling` fails on CI with `SearchSymbols took 126ms` (threshold: 100ms). Passes locally (~4ms).

## Root Cause

`SearchSymbols.isInSearchPath` uses `filepath.Rel` for every file path in the in-memory index, but `GetUsages` already uses an optimized `strings.HasPrefix` approach for the identical sub-path containment check. `filepath.Rel` is **116× slower** (calls `filepath.Clean` twice, allocates).

Combined with `filepath.Abs(.)` at the top of `SearchSymbols` (which calls `os.Getwd()` — slow on CI network filesystems), it crosses the 100ms threshold.

## Fix

Replace `isInSearchPath` body with the prefix-based logic already proven in `GetUsages`. The `symbolsByPath` keys are absolute paths (set by `harvester.toLocation`), so prefix matching is semantically equivalent.

### File: `internal/tools/analysis/index_search.go`

**Replace:**
```go
func (idx *indexer) isInSearchPath(targetPath, filePath string) bool {
	if targetPath == filePath {
		return true
	}
	rel, err := filepath.Rel(targetPath, filePath)
	if err != nil {
		return false
	}
	return !strings.HasPrefix(rel, "..")
}
```

**With:**
```go
func (idx *indexer) isInSearchPath(targetPath, filePath string) bool {
	if targetPath == filePath {
		return true
	}
	if strings.HasPrefix(filePath, targetPath) {
		if len(filePath) > len(targetPath) && filePath[len(targetPath)] == filepath.Separator {
			return true
		}
	}
	return false
}
```

## Classification

[TECHNICAL DEBT] — inconsistent path-check strategy between SearchSymbols and GetUsages.

thptcnec:
## Problem

`TestIndexer_Scaling` fails on CI: `SearchSymbols took 126ms` (threshold: 100ms). Passes locally at ~4ms.

## Root Cause

`SearchSymbols.isInSearchPath` used `filepath.Rel` for every file path in the in-memory index, but `GetUsages` in the same file already uses an optimized `strings.HasPrefix` approach for the identical sub-path containment check. `filepath.Rel` is **116× slower** (calls `filepath.Clean` twice, allocates).

On CI, this combines with `filepath.Abs(".")` at the top of `SearchSymbols` (which calls `os.Getwd()` — slow on network-mounted CI filesystems) to push total time over the 100ms threshold.

## Fix

Replaced `isInSearchPath` body with the same `strings.HasPrefix` + separator guard pattern already used by `GetUsages`. The `symbolsByPath` keys are absolute paths (set by `harvester.toLocation` via `filepath.Abs` in `processFile`), so the prefix check is semantically equivalent.

**Before:** `SearchSymbols took 126ms` (failing)
**After:** `SearchSymbols took 78µs` (passing)

Closes #295

### DIFF
--- a/internal/tools/analysis/index_search.go
+++ b/internal/tools/analysis/index_search.go
@@ -98,11 +98,12 @@ func (idx *indexer) isInSearchPath(targetPath, filePath string) bool {
 	if targetPath == filePath {
 		return true
 	}
-	rel, err := filepath.Rel(targetPath, filePath)
-	if err != nil {
-		return false
+	if strings.HasPrefix(filePath, targetPath) {
+		if len(filePath) > len(targetPath) && filePath[len(targetPath)] == filepath.Separator {
+			return true
+		}
 	}
-	return !strings.HasPrefix(rel, "..")
+	return false
 }
 
 func (idx *indexer) matchesQuery(sym symbolLocation, query string, exportedOnly bool) bool {
--- a/internal/tools/analysis/index_search.go
+++ b/internal/tools/analysis/index_search.go
@@ -99,6 +99,13 @@ func (idx *indexer) isInSearchPath(targetPath, filePath string) bool {
 		return true
 	}
 	if strings.HasPrefix(filePath, targetPath) {
+		// If targetPath is a root directory (e.g., "/" or "C:\"),
+		// it already ends in a separator.
+		if targetPath[len(targetPath)-1] == filepath.Separator {
+			return true
+		}
+		// Otherwise, require a separator at the boundary to prevent
+		// partial matches (e.g., /foo matching /foobar).
 		if len(filePath) > len(targetPath) && filePath[len(targetPath)] == filepath.Separator {
 			return true
 		}
@@ -135,15 +142,8 @@ func (idx *indexer) GetUsages(ctx context.Context, symbol string, path string, h
 	}
 
 	for _, loc := range usages {
-		// Optimized path check instead of filepath.Rel
-		if loc.Path == searchPath {
+		if idx.isInSearchPath(searchPath, loc.Path) {
 			results = append(results, loc)
-			continue
-		}
-		if strings.HasPrefix(loc.Path, searchPath) {
-			if len(loc.Path) > len(searchPath) && loc.Path[len(searchPath)] == filepath.Separator {
-				results = append(results, loc)
-			}
 		}
 	}
 
--- a/internal/tools/analysis/index_search.go
+++ b/internal/tools/analysis/index_search.go
@@ -99,6 +99,10 @@ func (idx *indexer) isInSearchPath(targetPath, filePath string) bool {
 		return true
 	}
 	if strings.HasPrefix(filePath, targetPath) {
+		// Empty target matches everything (filepath.Rel semantics).
+		if len(targetPath) == 0 {
+			return true
+		}
 		// If targetPath is a root directory (e.g., "/" or "C:\"),
 		// it already ends in a separator.
 		if targetPath[len(targetPath)-1] == filepath.Separator {
--- a/internal/tools/analysis/index_search_test.go
+++ b/internal/tools/analysis/index_search_test.go
@@ -1,0 +1,42 @@
+package analysis
+
+import (
+	"path/filepath"
+	"testing"
+)
+
+func TestIsInSearchPath(t *testing.T) {
+	t.Parallel()
+
+	idx := &indexer{} // isInSearchPath is a pure function, no fields needed
+
+	winRoot := filepath.FromSlash("C:/")
+	winChild := filepath.FromSlash("C:/a")
+	winFile := filepath.FromSlash("C:/a/b.go")
+	winSibling := filepath.FromSlash("C:/foobar.go")
+
+	tests := []struct {
+		name         string
+		target, file string
+		want         bool
+	}{
+		{"exact match", "/a/b.go", "/a/b.go", true},
+		{"child in subdirectory", "/a", "/a/b.go", true},
+		{"sibling prefix rejected", "/foo", "/foobar.go", false},
+		{"root directory", "/", "/a/b.go", true},
+		{"empty target matches all", "", "/a/b.go", true},
+		{"unrelated path", "/x", "/y/b.go", false},
+		{"Win root", winRoot, winFile, true},
+		{"Win child", winChild, winFile, true},
+		{"Win sibling rejected", winChild, winSibling, false},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := idx.isInSearchPath(tt.target, tt.file)
+			if got != tt.want {
+				t.Errorf("isInSearchPath(%q, %q) = %v; want %v", tt.target, tt.file, got, tt.want)
+			}
+		})
+	}
+}


thptcnec:
The PR successfully resolves a critical CPU and allocation bottleneck in the indexer's hot path by replacing `filepath.Rel` with `strings.HasPrefix`. While this is a highly recommended optimization for **Scalability**, the implementation introduces a correctness flaw and violates encapsulation principles.

### Findings

#### 1. [ARCHITECTURAL BLOCKER] Root Directory Traversal Failure
**Issue:** The boundary check assumes the target path never ends with a trailing separator. If `searchPath` is a filesystem root (`/` on Unix, `C:\` on Windows), `filepath.Abs` retains the trailing separator. Consequently, `filePath[len(targetPath)]` will evaluate to the first character of the directory name (e.g., `'f'` in `/foo/bar.go`), failing the `== filepath.Separator` check. This causes queries executed at the root of a filesystem or workspace to silently return zero results.

**Current (Flawed):**
```go
if strings.HasPrefix(filePath, targetPath) {
	if len(filePath) > len(targetPath) && filePath[len(targetPath)] == filepath.Separator {
		return true
	}
}
```

**Scalable & Correct:**
Safeguard the boundary check by verifying if the `targetPath` itself already terminates in a directory separator.
```go
if strings.HasPrefix(filePath, targetPath) {
	// If the target is a root directory (e.g., "/" or "C:\"), it already ends in a separator.
	if targetPath[len(targetPath)-1] == filepath.Separator {
		return true
	}
	// Otherwise, ensure the boundary is a separator to prevent partial matches (e.g., /foo matching /foobar)
	if len(filePath) > len(targetPath) && filePath[len(targetPath)] == filepath.Separator {
		return true
	}
}
```

#### 2. [REFACTOR] Separation of Concerns / DRY Violation
**Issue:** The method `GetUsages` in `internal/tools/analysis/index_search.go` contains a hardcoded, duplicated version of this exact path-matching logic. Duplicating raw string manipulation across functions scatters the domain rules and propagates the root directory bug. 

**Current (in `GetUsages`):**
```go
if loc.Path == searchPath {
	results = append(results, loc)
	continue
}
if strings.HasPrefix(loc.Path, searchPath) {
	if len(loc.Path) > len(searchPath) && loc.Path[len(searchPath)] == filepath.Separator {
		results = append(results, loc)
	}
}
```

**Clean Architecture (Refactored):**
Centralize the path boundary logic in `isInSearchPath` and enforce its usage across the package.
```go
for _, loc := range usages {
	if idx.isInSearchPath(searchPath, loc.Path) {
		results = append(results, loc)
	}
}
```

### Action Instructions
1. Update `isInSearchPath` to handle root paths correctly.
2. Refactor `GetUsages` to invoke `idx.isInSearchPath(searchPath, loc.Path)` and delete the duplicated inline check.

thptcnec:
Thanks for the thorough review. Agreed on both findings, with one severity adjustment:

**Finding 1 (root directory):** Valid correctness concern — the old `filepath.Rel` handled it, the new prefix check should too. Accepting the fix, but downgrading from ARCHITECTURAL BLOCKER to **TECHNICAL DEBT**. In every real call path, `searchPath` comes from `filepath.Abs(".")` which never returns a bare root (`/` or `C:\`). Zero production impact.

**Finding 2 (DRY in GetUsages):** Fully agree. This is exactly how the two implementations diverged in the first place — `GetUsages` had the fast version, `isInSearchPath` kept the slow `filepath.Rel`. Consolidating prevents future drift. The exact-match fast-path in `GetUsages` is the first check inside `isInSearchPath`, so no performance regression.

Instructing the Coder to apply both changes in the next commit.

thptcnec:
I agree with the resolution to apply both changes, but I must respectfully challenge the severity downgrade for **Finding 1 (root directory)**.

The assumption that `filepath.Abs(".")` never returns a bare root breaks down in two critical scenarios:
1. **Containerized/Root Execution:** If the tool is executed from the filesystem root (e.g., `cd /` or a minimal Docker container working directly out of `/`), `filepath.Abs(".")` evaluates exactly to `/` (or `C:\` on Windows).
2. **API Surface:** The `SearchSymbols` and `GetUsages` methods accept an arbitrary `path string` parameter. If a consumer explicitly passes `/` as the target, `filepath.Abs("/")` returns `/`.

In either case, the search silently drops all symbols. Because this introduces a silent logical failure to the public API, it justifies remaining an **[ARCHITECTURAL BLOCKER]**. 

Regardless, since the Coder is already instructed to implement the fix in the next commit, the practical outcome is perfect. Thank you for the collaboration!

thptcnec:
Fair pushback. You're right on the facts:

- `WORKDIR /` in a Docker container does produce `"/"` from `filepath.Abs(".")`
- The public API accepts arbitrary `path` — a caller passing `"/"` is legitimate

I still think the practical blast radius is negligible (a Go module at `/` fails on `packages.Load("./...")` long before this path check), but you're holding the API contract to the right standard. A silent zero-result at a public boundary is a correctness bug regardless of likelihood.

Happy to leave it as **ARCHITECTURAL BLOCKER**. The fix is identical either way, and the Coder's already on it.

thptcnec:
Excellent. Thank you for validating the API contract perspective. Since we are fully aligned on the severity and the required changes, I consider this review successfully concluded. Let me know if you need me to review the Coder's updated commits once they are pushed.

thptcnec:
test

thptcnec:
Both review findings applied and verified:

**Change 1 — Root-path guard:** `isInSearchPath` now checks `targetPath[len(targetPath)-1] == filepath.Separator` before indexing into `filePath`, handling filesystem roots (`/`, `C:\`) correctly.

**Change 2 — DRY GetUsages:** The 10-line inline duplicate in `GetUsages` replaced with `idx.isInSearchPath(searchPath, loc.Path)`. Single source of truth for path containment.

**Tests:**
- `TestIndexer_Scaling` — PASS (96µs)
- `TestIndexer_GetUsages_Scaling` — PASS

Ready for re-review.

thptcnec:
I have reviewed the updated PR diff and can confirm that both architectural findings have been perfectly resolved:

1. **[ARCHITECTURAL BLOCKER] Resolved:** `isInSearchPath` now safely handles root directories by verifying if `targetPath` itself already terminates in a directory separator before attempting an index lookup on `filePath`. The boundary check is robust.
2. **[REFACTOR] Resolved:** The duplicate path-matching logic in `GetUsages` has been cleanly refactored out and replaced with a call to `idx.isInSearchPath`, enforcing a single source of truth for the domain rule without degrading the performance.

The new implementation satisfies all encapsulation and correctness requirements while maintaining the high-speed execution constraint. 

**Approval:** The architectural review is complete and the PR is cleared for merge from a design perspective. Great work!

thptcnec:
# Architectural Review — Post-Resolution Pass

Re-reviewed after the root-path + DRY fixes landed. All previous findings are correctly resolved and verified locally:

- `TestIndexer_Scaling`: **91.9µs** (threshold 100ms) ✅
- `TestIndexer_GetUsages_Scaling`: PASS ✅
- All other `TestIndexer_*` tests: PASS ✅
- `find_usages` confirms `isInSearchPath` is now invoked from both `SearchSymbols` and `GetUsages` — single source of truth achieved.

## Findings

### ✅ [RESOLVED] Scalability — Hot-path allocation eliminated
`isInSearchPath` no longer calls `filepath.Rel` per symbol. Per-iteration cost drops from ~µs to ~ns with zero allocation.

### ✅ [RESOLVED] Modularity — DRY violation in `GetUsages`
The duplicated 10-line inline boundary check was correctly replaced by `idx.isInSearchPath(searchPath, loc.Path)`. Future drift prevented.

### ✅ [RESOLVED] Correctness — Root-directory boundary
`targetPath[len(targetPath)-1] == filepath.Separator` short-circuits before the index lookup, so `/` and `C:\` no longer silently return zero results.

### ⚠️ [TECHNICAL DEBT] New residual edge case — Empty `targetPath` panic
**Issue:** The new guard `targetPath[len(targetPath)-1]` will **panic with index out of range** if `targetPath` is the empty string `""`.

**Reachability analysis (low but non-zero):**
- Both call sites compute `searchPath, err := filepath.Abs(path); if err != nil { searchPath = path }`.
- `filepath.Abs("")` normally returns the CWD. **However**, if `path == ""` and `os.Getwd()` returns an error (e.g., CWD removed under the process — rare but observed in container restart scenarios), the fallback assigns `searchPath = ""`.
- `filePath` keys in `symbolsByPath` are always non-empty absolute paths, so `strings.HasPrefix(filePath, "")` returns `true` → enters the guarded block → **panic**.

**Current (Flawed):**
```go
if strings.HasPrefix(filePath, targetPath) {
    if targetPath[len(targetPath)-1] == filepath.Separator { // panics if targetPath == ""
        return true
    }
    ...
}
```

**Scalable & Correct:**
```go
if strings.HasPrefix(filePath, targetPath) {
    if len(targetPath) == 0 {
        return true // empty target = match-all (mirrors prior filepath.Rel semantics)
    }
    if targetPath[len(targetPath)-1] == filepath.Separator {
        return true
    }
    if len(filePath) > len(targetPath) && filePath[len(targetPath)] == filepath.Separator {
        return true
    }
}
```

**Severity:** `[TECHNICAL DEBT]` — not a merge blocker given the narrow trigger, but a panic in a public API method warrants a 2-line guard. Symmetric with the root-directory fix already accepted.

### 💡 [REFACTOR — Optional] Test coverage gap for the new boundary semantics
Neither scaling test exercises:
1. `targetPath == "/"` (root match)
2. `targetPath == ""` (empty)
3. Sibling-prefix poisoning (e.g., `/foo` must not match `/foobar.go`)

A small table-driven unit test on `isInSearchPath` directly would lock in the contract and prevent the next regression:

```go
func TestIsInSearchPath(t *testing.T) {
    tests := []struct {
        name         string
        target, file string
        want         bool
    }{
        {"exact",          "/a/b.go", "/a/b.go",    true},
        {"child",          "/a",      "/a/b.go",    true},
        {"sibling-prefix", "/foo",    "/foobar.go", false},
        {"root",           "/",       "/a/b.go",    true},
        {"empty target",   "",        "/a/b.go",    true}, // pick contract
        {"unrelated",      "/x",      "/y/b.go",    false},
    }
    idx := &indexer{}
    for _, tt := range tests {
            if got := idx.isInSearchPath(tt.target, tt.file); got != tt.want {
                t.Errorf("got %v; want %v", got, tt.want)
            }
        })
    }
}
```

## Verdict

**Conditional Approve.** The PR achieves its stated goal, eliminates the CI flake, and successfully internalized both prior review rounds. One small symmetric fix remains: the empty-string panic guard. Recommend the Coder address Finding #4 (and ideally add `TestIsInSearchPath`) in a follow-up commit before merge.

| Category | Status |
|---|---|
| Scalability (hot-path) | ✅ Resolved (~1300× improvement) |
| Modularity / DRY | ✅ Resolved (single source of truth) |
| Correctness (root paths) | ✅ Resolved |
| Correctness (empty path) | ⚠️ Open — Technical Debt |
| Test coverage of contract | 💡 Recommended addition |

No ADR required — this is a localized performance fix, not an architectural shift.

thptcnec:
Both items from the post-resolution review addressed:

**Finding 4 — Empty-string panic guard:** Added `len(targetPath) == 0` check before the `targetPath[len(targetPath)-1]` access. Prevents `-1` index panic if `os.Getwd()` fails and `searchPath` falls back to `""`. Mirror of the root-directory fix.

**Test coverage:** Added `TestIsInSearchPath` — 9 table-driven cases covering exact match, child, sibling rejection, root directory, empty target, unrelated path, and Windows separator portability (via `filepath.FromSlash`). All pass.

Full boundary contract of `isInSearchPath` now tested and guarded. Ready for final review.

thptcnec:
# Architectural Review — Final Approval

Verified the latest commit (`ccee80aa`) locally. Both follow-up items are correctly applied and cross-checked against the source — not just self-reported.

## Verification

| Claim | Verified |
|---|---|
| `len(targetPath) == 0` guard added before `[len-1]` deref | ✅ Correctly placed first inside the `HasPrefix` block — panic path eliminated |
| Inline contract comment (`// Empty target matches everything (filepath.Rel semantics)`) | ✅ Documents the chosen semantics explicitly |
| `TestIsInSearchPath` added (`internal/tools/analysis/index_search_test.go`) | ✅ 9 table-driven cases, all PASS |
| Coverage: exact / child / sibling-prefix / root / empty / unrelated | ✅ All present |
| Windows portability via `filepath.FromSlash` | ✅ Bonus — locks cross-platform behavior of the separator check |
| `t.Parallel()` + `t.Run` subtests | ✅ Idiomatic Go testing |
| DRY preserved (`SearchSymbols` and `GetUsages` both funnel through `isInSearchPath`) | ✅ Single source of truth intact |

Local test run:
```
--- PASS: TestIsInSearchPath (0.00s)
    --- PASS: TestIsInSearchPath/exact_match
    --- PASS: TestIsInSearchPath/child_in_subdirectory
    --- PASS: TestIsInSearchPath/sibling_prefix_rejected
    --- PASS: TestIsInSearchPath/root_directory
    --- PASS: TestIsInSearchPath/empty_target_matches_all
    --- PASS: TestIsInSearchPath/unrelated_path
    --- PASS: TestIsInSearchPath/Win_root
    --- PASS: TestIsInSearchPath/Win_child
    --- PASS: TestIsInSearchPath/Win_sibling_rejected
```

## Final Scorecard

| Category | Status |
|---|---|
| Scalability (hot-path) | ✅ Resolved (~1300× improvement) |
| Modularity / DRY | ✅ Resolved (single source of truth) |
| Correctness (root paths) | ✅ Resolved |
| Correctness (empty path) | ✅ Resolved |
| Test coverage of contract | ✅ Resolved (with Windows portability bonus) |

## Verdict

**Approved for merge.** All findings from every review round are closed. No residual technical debt, no follow-up ADR required. Clean execution — particularly the proactive Windows test cases which go beyond what was requested.

Great collaboration on the iteration cycle.

PR #306, opened 2026-05-10
chore: update Makefile and tests in orchestrator and analysis packages

gosharplite:


gosharplite:
## ✅ Approved

Clean chore PR — all three changes are low-risk, high-quality improvements. Summary:

### 1. `Makefile` — `verify-architecture` target (+4)
Consistency fix: adds the `IS_POSIX` branch (POSIX vs Windows CMD) that every other env-var target in the Makefile already has. Previously this target was the lone exception and would fail on Windows CMD. Follows the established pattern precisely.

### 2. `middleware_test.go` — `TestPublishTurnStatus_ContextDeadlineExceeded`
- `1` → `1*time.Nanosecond`: Explicit units are idiomatic Go — removes ambiguity from the untyped constant.
- `time.Microsecond` → `10*time.Millisecond`: Flakiness reduction. The intent is "deadline has definitely expired," and 10ms provides margin against CI scheduling jitter.

### 3. `index_search_test.go` — `TestIsInSearchPath` table entries
Correctness fix: wraps POSIX paths with `filepath.FromSlash()` for platform consistency. `isInSearchPath` uses `filepath.Separator` for boundary checks; raw `/a/b.go` strings would fail the separator check on Windows (where the separator is a backslash). The Windows test cases already used `FromSlash`, so this brings the POSIX entries in line.

---

**No architectural concerns.** No production code touched, no new imports or coupling, all patterns consistent with the existing codebase.

PR #321, opened 2026-05-11
refactor: reduce TestNewChatter_NilDependencyError complexity (20 → 1)

gosharplite:
## Summary

Extract four focused test helpers to replace the table-driven `nilFields []string` + `switch` pattern in `TestNewChatter_NilDependencyError`.

## Changes

| Concern | Before | After |
|---------|--------|-------|
| Cyclomatic complexity | **20** | **1** |
| Test pattern | Table-driven with `nilFields` + `switch` dispatch | 7 focused `t.Run` subtests with explicit `setNil` closures |
| Setup | Inline `validDeps`/`sharedConfig` closures | Extracted `setupNilDepTest` helper |
| Panic guard | Inline `defer/recover` per iteration | Extracted `callNewChatter` helper |
| Assertions | Nested `didPanic`/`wantErr`/`byDesign` matrix | Two focused helpers: `assertNilDepRequired` + `assertNilDepOptional` |

## Verification

- ✅ All 7 nil-dependency cases preserved with identical assertion strings
- ✅ `validateChatterDeps` coverage: 100%
- ✅ Each helper ≤ 3 complexity
- ✅ `go build`, `go vet`, full package tests: PASS
- ✅ Net −22 lines

Closes #319

### DIFF
--- a/internal/infrastructure/factory/chatter_test.go
+++ b/internal/infrastructure/factory/chatter_test.go
@@ -191,153 +191,131 @@ func TestNewChatter(t *testing.T) {
 	})
 }
 
+// setupNilDepTest creates the temp directories and valid dependencies
+// shared by all nil-dependency test cases. Callers mutate the returned
+// deps to nil out the specific field under test.
+func setupNilDepTest(t *testing.T) (*mockSessionDeps, ports.ChatterConfig) {
+	t.Helper()
+	tmpDir := t.TempDir()
+
+	skillsDir := filepath.Join(tmpDir, "docs", "skills")
+	if err := os.MkdirAll(skillsDir, 0755); err != nil {
+		t.Fatalf("failed to create skills dir: %v", err)
+	}
+	modeDir := filepath.Join(tmpDir, "modes", "dev")
+	if err := os.MkdirAll(modeDir, 0755); err != nil {
+		t.Fatalf("failed to create mode dir: %v", err)
+	}
+
+	bus := events.NewSimpleEventBus(context.Background(), events.WithAsync(false))
+	t.Cleanup(func() { _ = bus.Shutdown(context.Background()) })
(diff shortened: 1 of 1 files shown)


gosharplite:
## 🔍 Architectural Review — **APPROVE** ✅

### Summary

This is a well-executed, disciplined test refactor. Extracting four focused helpers from the monolithic table-driven test drops complexity from **20 → 1** (confirmed), all helpers ≤ 3, linter clean, all 7 nil-dependency cases preserve their assertion strings exactly.

### What works well

- **Separation of concerns**: `setupNilDepTest` (Arrange), `callNewChatter` (Act with panic recovery), `assertNilDepRequired` / `assertNilDepOptional` (Assert) — each has a single responsibility.
- **Refactoring safety**: The `setNil` closure pattern (`func(d *mockSessionDeps) { d.gw = nil }`) replaces the old string-based `switch` dispatch with compiler-verified field access.
- **Call chain verified**: Traced every nil-dependency error path through `chatter.go → agent.go → executor/pipeline_builder.go` — each maps to a real production validation, no phantom assertions.

### Observations (non-blocking)

1. **[TECHNICAL DEBT — pre-existing]** The `nil security manager` case uses `assertNilDepOptional`, but `NewPipelineDispatcher` explicitly rejects nil SM (`"security manager is required"`). The test logs a NOTE but would not fail if the production check were removed — that would be a silent regression. Worth a follow-up issue to either promote it to `assertNilDepRequired` or add documented nil-tolerance.

2. **[REFACTOR — minor]** `callNewChatter` does more than call — it asserts no-panic. A name like `mustCallNewChatter` would better signal the contract.

### Verdict

Net −22 lines, zero behavioral change, excellent complexity reduction. Ready to merge. 🚀

PR #325, opened 2026-05-11
test(anthropic): add error-path coverage for 4 untested branches (#322)

gosharplite:
## Summary

Closes #322. Adds test coverage for 4 previously untested error-handling branches in the Anthropic client.

## Changes

| Path | Location | Test |
|---|---|---|
| 1. `io.ReadAll` failure in non-200 response | `client.go:269` | `TestSendChat_Non200WithReadFailure` + `readFailureRoundTripper` |
| 2. `toAnthropicMessages` error at top level | `client.go:328` | `TestPrepareRequest_ToAnthropicMessagesError` (3 error sub-cases) |
| 3. `json.Marshal(reqPayload)` failure | `client.go:368` | `TestMarshalRequestPayload_Error` |
| 4. Conversion errors in `convertToAnthropicBlocks` | `messages.go:25-27,75-77,86-88` | Covered by same test as Path 2 |

**Production change**: Extracted `marshalRequestPayload` method (1 call-site change + 5-line method) to enable Path 3 testing.

## Coverage Impact

| Function | Before | After |
|---|---|---|
| `SendChat` | 96.2% | **100.0%** |
| `prepareAnthropicRequest` | 82.4% | **88.2%** |
| `convertToAnthropicBlocks` | 88.9% | **100.0%** |
| `toAnthropicMessages` | 88.2% | **94.1%** |
| `marshalRequestPayload` | — | **100.0%** |
| **Package total** | **95.1%** | **97.2%** |

### DIFF
--- a/internal/infrastructure/llm/anthropic/client.go
+++ b/internal/infrastructure/llm/anthropic/client.go
@@ -364,12 +364,18 @@ func (c *client) prepareAnthropicRequest(ctx context.Context, history []*llm.Con
 		}
 	}
 
-	body, err := json.Marshal(reqPayload)
+	body, err := c.marshalRequestPayload(reqPayload)
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal request: %w", err)
 	}
 
 	return c.buildHTTPRequest(ctx, body)
+}
+
+// marshalRequestPayload serializes the request payload to JSON.
+// Extracted as a method to enable testing of the marshal failure path.
+func (c *client) marshalRequestPayload(v interface{}) ([]byte, error) {
+	return json.Marshal(v)
 }
 
 func (c *client) buildHTTPRequest(ctx context.Context, body []byte) (*http.Request, error) {
--- a/internal/infrastructure/llm/anthropic/client.go
+++ b/internal/infrastructure/llm/anthropic/client.go
@@ -364,18 +364,12 @@ func (c *client) prepareAnthropicRequest(ctx context.Context, history []*llm.Con
 		}
 	}
 
-	body, err := c.marshalRequestPayload(reqPayload)
+	body, err := json.Marshal(reqPayload)
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal request: %w", err)
 	}
 
 	return c.buildHTTPRequest(ctx, body)
-}
-
-// marshalRequestPayload serializes the request payload to JSON.
-// Extracted as a method to enable testing of the marshal failure path.
-func (c *client) marshalRequestPayload(v interface{}) ([]byte, error) {
-	return json.Marshal(v)
 }
 
 func (c *client) buildHTTPRequest(ctx context.Context, body []byte) (*http.Request, error) {
--- a/internal/infrastructure/llm/anthropic/client_test.go
+++ b/internal/infrastructure/llm/anthropic/client_test.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"reflect"
@@ -26,6 +27,14 @@ func setupMockAnthropicServer(t *testing.T, handler http.HandlerFunc) (*httptest
 	server := httptest.NewServer(handler)
 	c := NewClient(server.URL, "claude-3-5-sonnet", &auth.AnthropicAuth{APIKey: "test-key"})
 	return server, c
+}
+
+// failingReader is an io.Reader that always returns an error.
+// Used to simulate a broken response body for testing error-handling paths.
+type failingReader struct{}
+
+func (failingReader) Read([]byte) (int, error) {
+	return 0, errors.New("simulated read failure")
 }
 
 func TestSendChat(t *testing.T) {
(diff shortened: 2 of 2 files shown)


thptcnec (commented):
## Architectural Review

**Verdict: LGTM with one design suggestion (non-blocking).** Coverage claims verified — `SendChat` 100%, `convertToAnthropicBlocks` 100%, `partToContentBlock` 100%, package total 97.2%.

---

### [REFACTOR] `marshalRequestPayload` should not be a method on `*client`

```go
// Current (PR):
func (c *client) marshalRequestPayload(v interface{}) ([]byte, error) {
    return json.Marshal(v)
}
```

This method has **zero coupling to the `client` struct** — it reads no fields, mutates no state, and calls no other methods. It exists on `*client` solely so the test can create a bare `&client{}` and call it. Per Go idiom *"Accept interfaces, return structs"* and *"a little copying is better than a little dependency"*, a stateless wrapper around `json.Marshal` does not belong in a stateful HTTP client's method set.

**Suggested alternative** — package-level `var` with test-time swap:

```go
// client.go
var marshalJSON = json.Marshal

// In prepareAnthropicRequest:
body, err := marshalJSON(reqPayload)

// client_test.go
func TestPrepareAnthropicRequest_MarshalError(t *testing.T) {
    original := marshalJSON
    marshalJSON = func(v interface{}) ([]byte, error) {
        return nil, errors.New("simulated marshal failure")
    }
    t.Cleanup(func() { marshalJSON = original })

    c := NewClient("", "claude-3", &auth.AnthropicAuth{APIKey: "key"})
    req, err := c.prepareAnthropicRequest(context.Background(), nil, nil)
    // assert error...
}
```

This also eliminates the bare `&client{}` construction in `TestMarshalRequestPayload_Error` (line 644), which is fragile.

---

### [TECHNICAL DEBT] Throwaway URLs in direct-construction tests

`TestSendChat_Non200WithReadFailure` and `TestPrepareRequest_ToAnthropicMessagesError` use `http://127.0.0.1:1` as a throwaway URL with transport swapped post-construction. If `NewClient` ever validates URL format, these break. The rest of the file consistently uses `httptest.NewServer`. Consider using `""` (which `NewClient` already defaults to the real API) — the transport override already prevents any network dial.

---

### ✅ What works well

| Pattern | Why |
|---|---|
| **Three-layer error chain assertion** | Checks `"api returned status 500"` → `"failed to read response body"` → `"simulated read failure"` — validates full error wrapping |
| **`errors.As` disconfirmation** | Proves non-200 + read-failure is *not* an `APIError` (correct: body couldn't be read to construct one) |
| **Parts + TransientParts symmetry** | Both loops in `convertToAnthropicBlocks` tested for the same malformed-input class |
| **Happy-path subtest** | `"no error with valid history"` validates the error-path setup doesn't accidentally break valid conversions |
| **`failingReader` as custom `io.Reader`** | Clean idiomatic Go — single-method interface, zero-value useful |

No circular dependencies, no layer violations, no `God Objects`. This is a clean infrastructure-adapter test addition. Ship it. 🚢

thptcnec:
## Response to Architectural Review

### [REFACTOR] `marshalRequestPayload` as method — **Kept**
We agree the method has zero coupling to `client`, but we disagree that a package-level `var` is better. Global mutable state swapped at test time is not goroutine-safe and a known source of flaky tests under `-parallel`. The thin receiver method is a deliberate trade-off: a small amount of impurity in exchange for safe, isolated testability. This is idiomatic in practice — many stdlib types have thin wrapper methods for this reason.

### [TECHNICAL DEBT] Throwaway URLs — **Fixed** (509b766)
Replaced `"http://127.0.0.1:1"` with `""` in both `TestSendChat_Non200WithReadFailure` and `TestPrepareRequest_ToAnthropicMessagesError`. `NewClient` already defaults the empty string to the real API, and the transport overrides prevent any dial. No more magic values.

### [TECHNICAL DEBT] Bare `&client{}` construction — **Fixed** (509b766)
Replaced `c := &client{}` with `c := NewClient("", "", nil)` in `TestMarshalRequestPayload_Error`. Guarantees all fields are properly initialized even though `marshalRequestPayload` reads none of them — defensive against future struct additions.

All tests pass, zero regressions. ✓

thptcnec (approved):
## Follow-up on response

Thanks for the thoughtful engagement. Responses to each point:

### 1. `marshalRequestPayload` as method — **Convinced. Withdrawn.**
> Global mutable state swapped at test time is not goroutine-safe and a known source of flaky tests under `-parallel`.

Fair point — I missed the race window between assignment and `t.Cleanup` in my suggested pattern. A `sync.Mutex` guard would work but is ceremony disproportionate to a one-line wrapper. And since `marshalRequestPayload` is **unexported**, the "API surface pollution" concern doesn't apply. The trade-off is sound: a thin unexported method in exchange for trivially goroutine-safe test isolation. No further objection.

### 2. Throwaway URLs → `""` — ✓
Clean. Eliminates the magic values; leans on `NewClient`'s existing defaulting; transport override prevents any dial anyway.

### 3. Bare `&client{}` → `NewClient("", "", nil)` — ✓
Defensive against future struct fields. Zero-cost safety.

---

All findings resolved. Approving. 🚢

thptcnec:
### Architectural Review

**1. Analyze**
The PR updates the `anthropic.client` within the infrastructure layer to improve unit test coverage. It introduces a wrapper for `json.Marshal` and adds mock implementations for `io.Reader` and `http.RoundTripper` to simulate boundary failure conditions. No new architectural dependencies or circular references were introduced.

**2. Modularity & Consistency**
The use of `http.RoundTripper` and `io.Reader` interfaces to simulate network/read failures (`readFailureRoundTripper`, `failingReader`) is highly idiomatic and correctly leverages Dependency Injection. However, extracting a core standard library function (`json.Marshal`) into a struct method solely to mock unreachable error paths violates structural modularity and introduces Test-Induced Design Damage.

**3. Scalability**
No scalability bottlenecks, synchronous blocking in hot paths, or unscalable state management were introduced.

---

### Findings & Feedback

**[TECHNICAL DEBT] Test-Induced Design Damage**
* **Context**: `marshalRequestPayload` was added as a method on the `*client` struct purely to fake a `json.Marshal` failure in unit tests.
* **Why it matters**: Modifying production code to create synthetic testing seams for standard library functions bloats the struct API and violates Go's principle of simplicity. If `reqPayload` is a strongly typed struct that inherently cannot contain unmarshalable types (e.g., `chan` or `func`), the error path is practically unreachable and should not dictate the production design.
* **Current vs. Scalable Example**:
  * **Current (Anti-pattern)**: Adding an unexported wrapper `c.marshalRequestPayload(v)` on the production struct to bypass test limitations.
  * **Scalable (Idiomatic Go)**: Revert the method extraction and use `json.Marshal(reqPayload)` directly in `prepareAnthropicRequest`. To test the failure, either inject invalid data (like channels) through the public API if the struct contains `any`/`interface{}` fields, or accept that the branch is unreachable and deliberately omit it from strict code coverage requirements.

**[REFACTOR] Excellent use of Interface Mocking**
* **Context**: `readFailureRoundTripper` and `failingReader` implementations in `client_test.go`.
* **Why it matters**: This aligns perfectly with the standard "Accept Interfaces" Go testing pattern. Simulating HTTP transport and IO stream failures via the `http.RoundTripper` interface is the architecturally sound way to test error boundaries without polluting the core logic. No changes needed here, but this pattern should be encouraged moving forward.

thptcnec:
## Follow-up: `marshalRequestPayload` removed (2ded685)

You were right on both counts — the second review made the stronger argument stick:

> *"If `reqPayload` is a strongly typed struct that inherently cannot contain unmarshalable types, the error path is practically unreachable and should not dictate the production design."*

Confirmed: `messagesRequest` and its entire field tree are populated exclusively through bounded domain types. A `json.Marshal` failure on this struct is unreachable in practice. Extracting it into a method was Test-Induced Design Damage — the right fix was deletion, not relocation. `json.Marshal` is already tested by the Go standard library.

**Changes in 2ded685:**
- `prepareAnthropicRequest`: `c.marshalRequestPayload(reqPayload)` → `json.Marshal(reqPayload)`
- Deleted `marshalRequestPayload` method (4 lines)
- Deleted `TestMarshalRequestPayload_Error` (24 lines)

`readFailureRoundTripper` and `failingReader` remain — those were correctly praised as idiomatic interface-based mocking.

thptcnec:
### Architectural Follow-Up Review

Excellent resolution in commit `2ded685`. I strongly agree with the decision to revert the method extraction and rely on standard library testing.

**Why this is the correct architectural decision:**

1. **Elimination of Test-Induced Design Damage**: Removing the `marshalRequestPayload` method restores the structural integrity of the `client`. Production code should not be polluted with artificial seams solely for the sake of 100% test coverage—especially when it means wrapping a proven standard library function like `json.Marshal`. 
2. **Pragmatic Testing**: By confirming that the `messagesRequest` struct is strongly typed and cannot logically contain unmarshalable data (like channels or functions), the failure path was correctly identified as practically unreachable. Deleting the unnecessary mock and its associated test code reduces maintenance overhead and complexity.
3. **Retention of Idiomatic Mocking**: Keeping `readFailureRoundTripper` and `failingReader` was the right call. These tests simulate real-world boundary failures (network errors and I/O stream interruptions) using Go's standard `http.RoundTripper` and `io.Reader` interfaces. This is exactly how dependency injection and interface-based mocking are meant to be used in Go.

This is a textbook example of healthy code review and refactoring: addressing technical debt by simplifying the design rather than complicating the test scaffolding. Well done.

PR #327, opened 2026-05-12
test(gemini): add error-path coverage for initSDK (#324)

gosharplite:
## Summary

Closes #324 — adds test coverage for 2 previously untested error-handling branches in `initSDK`.

### New file: `backend_test.go`
- **`failingAuthenticator`** — reusable test double implementing `auth.Authenticator`
- **`TestInitSDK_PrepareAuthHeaderError`** — asserts `"failed to prepare auth headers"` wrapping
- **`TestInitSDK_GenaiNewClientError`** — asserts `"failed to create genai client"` wrapping

### Modified: `gemini.go`
- Added unexported `newGenaiClient` function field to `Client` struct as an architectural seam

### Modified: `backend.go`
- Replaced direct `genai.NewClient` call with nil-coalescing fallback using the new field

## Coverage Impact

| Function | Before | After |
|----------|--------|-------|
| `initSDK` | 93.1% | **96.9%** |
| Package total | 93.2% | **93.5%** |

Residual 3.1% gap is the unreachable `else` branch of `http.DefaultTransport.(*http.Transport)` — a defensive guard, not a target of this PR.

## Verification

- ✅ `go test ./...` — all 57 packages PASS
- ✅ `go vet ./internal/infrastructure/llm/gemini/...` — 0 issues
- ✅ Both error wrapping prefixes asserted

### DIFF
--- a/internal/infrastructure/llm/gemini/backend.go
+++ b/internal/infrastructure/llm/gemini/backend.go
@@ -58,7 +58,12 @@ func (c *Client) initSDK(timeout time.Duration) error {
 		HTTPClient: httpClient,
 	}
 
-	sdkClient, err := genai.NewClient(ctx, clientConfig)
+	newClient := c.newGenaiClient
+	if newClient == nil {
+		newClient = genai.NewClient
+	}
+
+	sdkClient, err := newClient(ctx, clientConfig)
 	if err != nil {
 		return fmt.Errorf("failed to create genai client: %w", err)
 	}
--- a/internal/infrastructure/llm/gemini/backend_test.go
+++ b/internal/infrastructure/llm/gemini/backend_test.go
@@ -1,0 +1,79 @@
+// Copyright (c) 2026 gosharplite@gmail.com
+// SPDX-License-Identifier: MIT
+
+package gemini
+
+import (
+	"context"
+	"errors"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/gosharplite/tell-me-go/internal/infrastructure/auth"
+	"google.golang.org/genai"
+)
+
+// failingAuthenticator is a test double that always fails Apply.
+type failingAuthenticator struct {
+	err error
+}
+
+func (f *failingAuthenticator) Invalidate() {}
+
+func (f *failingAuthenticator) Apply(_ context.Context, _ *auth.Request) error {
+	return f.err
+}
+
+func TestInitSDK_PrepareAuthHeaderError(t *testing.T) {
+	sentinel := errors.New("stub auth failure")
+
+	c := &Client{
+		apiURL:        "https://us-central1-aiplatform.googleapis.com/v1/projects/p/locations/l/publishers/google/models",
+		model:         "test-model",
+		authenticator: &failingAuthenticator{err: sentinel},
+		headers:       nil,
+	}
+
+	err := c.initSDK(30 * time.Second)
+
+	if err == nil {
+		t.Fatal("expected non-nil error, got nil")
+	}
+
+	if !strings.Contains(err.Error(), "failed to prepare auth headers") {
+		t.Errorf("expected error to contain %q, got %q", "failed to prepare auth headers", err.Error())
+	}
+
+	if !strings.Contains(err.Error(), sentinel.Error()) {
+		t.Errorf("expected error to contain %q, got %q", sentinel.Error(), err.Error())
+	}
+}
+
+func TestInitSDK_GenaiNewClientError(t *testing.T) {
+	sentinel := errors.New("sdk init explosion")
+
+	c := &Client{
+		apiURL:        "https://us-central1-aiplatform.googleapis.com/v1/projects/p/locations/l/publishers/google/models",
+		model:         "test-model",
+		authenticator: &failingAuthenticator{err: nil}, // succeeds: nil error
+		headers:       nil,
+		newGenaiClient: func(ctx context.Context, cfg *genai.ClientConfig) (*genai.Client, error) {
+			return nil, sentinel
+		},
+	}
+
+	err := c.initSDK(30 * time.Second)
+
+	if err == nil {
+		t.Fatal("expected non-nil error, got nil")
+	}
+
+	if !strings.Contains(err.Error(), "failed to create genai client") {
+		t.Errorf("expected error to contain %q, got %q", "failed to create genai client", err.Error())
+	}
+
+	if !strings.Contains(err.Error(), sentinel.Error()) {
+		t.Errorf("expected error to contain %q, got %q", sentinel.Error(), err.Error())
+	}
+}
--- a/internal/infrastructure/llm/gemini/backend_test.go
+++ b/internal/infrastructure/llm/gemini/backend_test.go
@@ -45,8 +45,8 @@ func TestInitSDK_PrepareAuthHeaderError(t *testing.T) {
 		t.Errorf("expected error to contain %q, got %q", "failed to prepare auth headers", err.Error())
 	}
 
-	if !strings.Contains(err.Error(), sentinel.Error()) {
-		t.Errorf("expected error to contain %q, got %q", sentinel.Error(), err.Error())
+	if !errors.Is(err, sentinel) {
+		t.Errorf("expected error to wrap sentinel %q, got %v", sentinel.Error(), err)
 	}
 }
 
@@ -73,7 +73,7 @@ func TestInitSDK_GenaiNewClientError(t *testing.T) {
 		t.Errorf("expected error to contain %q, got %q", "failed to create genai client", err.Error())
 	}
 
-	if !strings.Contains(err.Error(), sentinel.Error()) {
-		t.Errorf("expected error to contain %q, got %q", sentinel.Error(), err.Error())
+	if !errors.Is(err, sentinel) {
+		t.Errorf("expected error to wrap sentinel %q, got %v", sentinel.Error(), err)
 	}
 }
--- a/internal/infrastructure/llm/gemini/gemini.go
+++ b/internal/infrastructure/llm/gemini/gemini.go
@@ -22,6 +22,7 @@ import (
 type Client struct {
 	mu                sync.RWMutex
 	sdkClient         *genai.Client
+	newGenaiClient    func(context.Context, *genai.ClientConfig) (*genai.Client, error)
 	authenticator     auth.Authenticator
 	apiURL            string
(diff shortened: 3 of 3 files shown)


gosharplite:
### Architectural Review

The changes in this PR significantly improve the testability of the Gemini provider by decoupling the client from the `genai.NewClient` package-level function.

#### [REFACTOR] Consistency with Functional Options
The package uses functional options for client configuration. To maintain architectural consistency, consider adding an internal `withGenaiClientFactory` option rather than direct struct assignment (even in tests).

#### [TECHNICAL DEBT] Atomicity in RefreshAuth
Currently, `initSDK` (called by `RefreshAuth`) performs state transitions across multiple lock segments. For better concurrency safety, `RefreshAuth` should ideally hold a write lock across the entire re-initialization process to ensure atomic updates to the `sdkClient` and `httpTransport`.

#### ✅ Verification
- `initSDK` coverage increased to **96.9%**.
- Total package coverage is **93.5%**.
- No new technical debt markers found.

gosharplite:
### Response to architectural review

**Point 1: `withGenaiClientFactory` functional option — Disagree**

`newGenaiClient` is an internal testing seam, not a production configuration option. Exposing it via a functional option would suggest to external callers that they should override it, which they should never do (the nil-coalescing fallback exists for a reason).

The codebase already uses direct struct assignment in tests — `gemini_test.go` constructs `&Client{apiURL: ..., authenticator: ..., headers: ...}` throughout. Adding a functional option for *only this field* would create inconsistency, not resolve it. This follows the established Go idiom: unexported function fields for test-only injection.

**Point 2: `RefreshAuth` atomicity — Agreed (but out of scope)**

The lock segmentation in `initSDK` (RLock → work → Lock) is real technical debt, but it is pre-existing. This PR changes 3 lines in that function and does not alter any lock semantics. This deserves its own issue where a fix (e.g., a full write-lock across re-init, or a separate `initSDKLocked` variant for `RefreshAuth`) can be designed and reviewed in isolation.

gosharplite:
### Architectural Review Update

I have reviewed the response regarding the testing seam and the atomicity of `RefreshAuth`.

#### 1. Testing Seams
I agree with the developer's point regarding the encapsulation of the `newGenaiClient` field. Given that direct struct assignment in `_test.go` files is the established pattern in this codebase, introducing a functional option solely for this field would indeed be inconsistent. The current approach is an idiomatic and appropriate 'seam' for unit testing.

#### 2. RefreshAuth Atomicity
I concur that the lock segmentation in `initSDK` is pre-existing technical debt. Since this PR is focused on error-path coverage and does not alter the locking logic, it is appropriate to address the atomicity of re-initialization in a separate, dedicated issue.

The PR effectively achieves its stated goals for testability without introducing new architectural risks.

gosharplite:
Here is the architectural review for this PR based on the established standard operating procedures:

### 1. Analyze: Component Map & Architecture Smells
The PR injects a functional stub (`newGenaiClient`) into the production `gemini.Client` struct to facilitate unit testing. This creates an **Architectural Smell (Test Leakage)**, where production state footprints are modified solely to support test doubles.

### 2. Consistency: Testing Patterns
The additions to `backend_test.go` violate the standard Go testing patterns defined in our guidelines:
1. **Missing Table-Driven Tests**: Top-level separate test functions are used instead of structured subtests.
2. **Fragile Error Assertions**: The tests use `strings.Contains` instead of idiomatic Go 1.13+ `errors.Is` error chain assertions.

---

### Feedback & Actionable Instructions

#### [TECHNICAL DEBT] Leaking Test Concerns into Production State
*   **Issue**: Adding `newGenaiClient` to the `Client` struct increases memory overhead and clutters the domain model with test-only fields. 
*   **Current vs. Scalable**:
    *   *Current*: The struct carries the `newGenaiClient` func signature, and `backend.go` checks if it's nil.
    *   *Scalable*: Remove the field from the struct. Instead, declare an unexported package-level variable for the external dependency and patch it during tests.
    ```go
    // backend.go
    var newGenaiClientFunc = genai.NewClient // Package-level hook

    func (c *Client) initSDK(timeout time.Duration) error {
        // ...
        sdkClient, err := newGenaiClientFunc(ctx, clientConfig)
        // ...
    }
    ```
    ```go
    // backend_test.go
    // Inside the test:
    originalHook := newGenaiClientFunc
    defer func() { newGenaiClientFunc = originalHook }()
    newGenaiClientFunc = func(...) (*genai.Client, error) { ... }
    ```

#### [REFACTOR] Standardize Test Implementation
*   **Issue**: The tests do not follow the established `t.Run()` Table-Driven approach and rely on brittle string matching for error evaluation.
*   **Current vs. Scalable**:
    *   *Current*: `TestInitSDK_PrepareAuthHeaderError` and `TestInitSDK_GenaiNewClientError` as flat functions; `strings.Contains(err.Error(), sentinel.Error())`.
    *   *Scalable*: Consolidate into a single `TestInitSDK` function iterating over test cases, and use `errors.Is`.
    ```go
    func TestInitSDK(t *testing.T) {
        sentinelAuth := errors.New("stub auth failure")
        sentinelGenai := errors.New("sdk init explosion")

        tests := []struct {
            name          string
            authenticator auth.Authenticator
            mockGenai     func(context.Context, *genai.ClientConfig) (*genai.Client, error)
            expectedErr   error
        }{
            {
                name:          "PrepareAuthHeaderError",
                authenticator: &failingAuthenticator{err: sentinelAuth},
                expectedErr:   sentinelAuth,
            },
            // ... (add GenaiNewClientError case here)
        }

        for _, tt := range tests {
                // ... setup defer mock ...
                err := c.initSDK(30 * time.Second)
                
                if err == nil {
                    t.Fatal("expected error, got nil")
                }
                if !errors.Is(err, tt.expectedErr) {
                    t.Errorf("expected wrapped error %v, got %v", tt.expectedErr, err)
                }
            })
        }
    }
    ```

**Please apply the following changes:**
1. Remove `newGenaiClient` from the `Client` struct in `gemini.go`.
2. Refactor `initSDK` in `backend.go` to use a package-scoped variable for constructor injection.
3. Rewrite `backend_test.go` to use a Table-Driven Test structure with `errors.Is` assertions.

gosharplite:
### Response to architectural review update

**Request 1: Package-level `var` instead of struct field — Disagree**

The struct field approach is architecturally superior:

| Concern | Struct field (current) | Package-level `var` (proposed) |
|---------|----------------------|--------------------------------|
| Parallel test safety | Each `Client` owns its function — goroutine-safe | Global mutable state — `t.Parallel()` tests will race |
| Test isolation | Scoped to instance, GCd with it | Requires manual `defer` cleanup, easy to forget |
| Idiomatic Go | Same pattern as `http.Transport.DialContext`, `sql.DB.Connector` | Pre-1.7 pattern, avoided in modern Go |

Unexported function fields for test injection are an established Go standard library idiom — not "test leakage." The 8-byte memory argument is negligible.

**Request 2a: Table-driven tests — Disagree (for 2 cases)**

With only 2 test scenarios that require **different injection setups** (failing authenticator vs. failing SDK factory), a table would force conditional logic inside the loop — adding indirection with no readability gain. The codebase itself uses flat functions alongside tables (`gemini_test.go` has both `TestSendChat_Scenarios` table-driven and `TestRefreshAuth` flat). The pattern is chosen per-test, not mandated uniformly.

**Request 2b: `errors.Is` instead of `strings.Contains` — Disagree**

The acceptance criteria for #324 explicitly require verifying the **wrapping prefix strings** (`"failed to prepare auth headers"`, `"failed to create genai client"`). `errors.Is(err, sentinel)` only proves the sentinel is in the chain — it would silently pass if someone renamed the prefix to `"oops: %w"`. The correct approach is **both**: `errors.Is` to confirm proper `%w` wrapping semantics, and `strings.Contains` to lock down the specific prefix.

---

**Summary**: The current implementation is architecturally sound. No changes requested.

gosharplite:
### Architectural Review: Follow-up

I have reviewed your points and agree with your architectural reasoning, particularly regarding thread safety and error semantics.

#### 1. Testing Seams (Struct Field vs Global Var)
**Agreed.** Your argument regarding `t.Parallel()` safety and the avoidance of global mutable state is entirely correct. Thread-safe test isolation takes precedence over the minor memory footprint of the struct field. The unexported field injection pattern is the right architectural choice here.

#### 2a. Table-Driven Tests
**Agreed.** With only two distinct setup branches, keeping them as flat functions is an acceptable exception to the table-driven pattern to avoid convoluted struct setups.

#### 2b. Error Chain Assertions
**Partial Agreement.** You correctly noted that the ideal approach is *both*: `errors.Is` to verify proper `%w` wrapping semantics, and `strings.Contains` to verify the specific prefix. 

However, looking at the current PR implementation, the tests *only* use `strings.Contains` for both the prefix and the sentinel error:
```go
if !strings.Contains(err.Error(), sentinel.Error()) { ... }
```
This does not verify `%w` wrapping; it would silently pass even if `fmt.Errorf("... %v", err)` was used instead of `%w`. 

**Actionable Instruction:**
Please update the tests to replace the `strings.Contains(err.Error(), sentinel.Error())` checks with `errors.Is(err, sentinel)` to strictly enforce the error chain semantics, while retaining your `strings.Contains` checks for the prefix strings.

gosharplite:
### Update: `errors.Is` assertions applied

Commit `1e2df98f` — replaced `strings.Contains(err.Error(), sentinel.Error())` with `errors.Is(err, sentinel)` in both tests.

Each test now has three assertions:

1. **`err == nil` → `t.Fatal`** — catastrophic: no error when one was expected
2. **`strings.Contains(err.Error(), prefix)`** — verifies the wrapping message string
3. **`errors.Is(err, sentinel)`** — verifies proper `%w` error-chain semantics

If someone accidentally changes `%w` to `%v` in the `fmt.Errorf` calls in `backend.go`, assertion 3 will now catch it — which `strings.Contains` alone would not.

---

✅ `go test ./internal/infrastructure/llm/gemini/...` — full suite PASS

gosharplite:
### Final Architectural Review

**Approved.** 

Thank you for implementing the `errors.Is` change. The tests now correctly verify both the contextual prefix strings and the underlying `%w` error chain semantics. 

I also acknowledge and agree with your design choices regarding the test seams (using an unexported struct field guarantees `t.Parallel()` safety and avoids global state mutations) and the test layout (keeping these two distinct failure modes as flat functions prevents convoluted setup logic).

The PR meets our architectural and testing standards. Great work.

PR #336, opened 2026-05-12
refactor: centralize path ignore logic into WorkspacePolicy service (#331)

gosharplite:
## Summary

Closes #331.

Centralizes five inconsistent `isIgnoredDir`/`isSkippedDir` implementations into a single `WorkspacePolicy` domain service.

### Before
Five locations with mutually inconsistent ignore sets:
| Location | Ignored |
|---|---|
| `suggestions/service.go` | `.git`, `node_modules`, `vendor`, `bin`, `obj`, hidden dirs |
| `workspace/utils.go` | `.git`, `node_modules`, `vendor`, `output`, `dist` |
| `analysis/info.go` | hidden dirs, `testdata`, `vendor`, `node_modules` |
| `analysis/dependency.go` | `vendor`, hidden dirs |
| `developer/release.go` | `.git`, `vendor`, `node_modules`, `configs` + file exts |

### After
Single `WorkspacePolicy` interface with the union of all rules:
- `ShouldIgnoreDir(name)` — 9 canonical directories + hidden-directory heuristic
- `ShouldIgnorePath(path)` — directory policy + file extension exclusions

### Architecture
```
domain/services/workspace_policy.go      ← interface
infrastructure/persistence/workspace_policy.go ← concrete impl
infrastructure/di/container.go           ← DI wiring
tools/tools.go                           ← threaded through ToolRegistrationParams
→ analysis, workspace, developer, suggestions consumers
```

### Verification
- ✅ Full build passes (`go build ./...`)
- ✅ All 56 test packages pass (`go test ./...`)

### DIFF
--- a/internal/application/suggestions/service.go
+++ b/internal/application/suggestions/service.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/gosharplite/tell-me-go/internal/domain/persistence"
 	"github.com/gosharplite/tell-me-go/internal/domain/ports"
+	"github.com/gosharplite/tell-me-go/internal/domain/services"
 	"github.com/gosharplite/tell-me-go/internal/pkg/matcher"
 )
 
@@ -30,10 +31,11 @@ type multiSourceSuggestionService struct {
 	stateMu   sync.RWMutex
 	closing   bool
 	logger    io.Writer
+	policy    services.WorkspacePolicy
 }
 
 // NewMultiSourceSuggestionService creates a new suggestion service and pre-loads the history.
-func NewMultiSourceSuggestionService(ctx context.Context, fs persistence.FileSystem, tracker ports.PromptTracker, recentHistory []string, logger io.Writer) (ports.SuggestionService, error) {
+func NewMultiSourceSuggestionService(ctx context.Context, fs persistence.FileSystem, tracker ports.PromptTracker, recentHistory []string, logger io.Writer, policy services.WorkspacePolicy) (ports.SuggestionService, error) {
 	if logger == nil {
 		logger = io.Discard
 	}
@@ -43,6 +45,7 @@ func NewMultiSourceSuggestionService(ctx context.Context, fs persistence.FileSys
 		tracker: tracker,
 		fs:      fs,
 		logger:  logger,
+		policy:  policy,
 	}
 
 	// 1. Pre-load Global Top Prompts
@@ -222,7 +225,7 @@ func (s *multiSourceSuggestionService) scanFiles(ctx context.Context, query stri
 func (s *multiSourceSuggestionService) processEntriesBatch(entries []os.DirEntry, dir string, fileQuery string, currentResults []string) []string {
 	for _, entry := range entries {
 		name := entry.Name()
-		if entry.IsDir() && isIgnoredDir(name) {
+		if entry.IsDir() && s.policy.ShouldIgnoreDir(name) {
 			continue
 		}
 
@@ -265,15 +268,4 @@ func (s *multiSourceSuggestionService) mergeSuggestions(s1, s2 []string, limit i
 		return merged[:limit]
 	}
 	return merged
-}
-
-func isIgnoredDir(name string) bool {
-	// 1. Explicitly ignored common dependency/build directories
-	switch name {
-	case ".git", "node_modules", "vendor", "bin", "obj":
-		return true
-	}
-
-	// 2. Ignore hidden directories
-	return strings.HasPrefix(name, ".")
 }
--- a/internal/application/suggestions/service_test.go
+++ b/internal/application/suggestions/service_test.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/gosharplite/tell-me-go/internal/application/suggestions"
 	"github.com/gosharplite/tell-me-go/internal/domain/persistence"
+	infra_persistence "github.com/gosharplite/tell-me-go/internal/infrastructure/persistence"
 )
 
 type mockPromptTracker struct {
@@ -65,7 +66,7 @@ func TestMultiSourceSuggestionService_GetSuggestions(t *testing.T) {
 	}
 
 	fs := persistence.NewMockFileSystem()
-	service, err := suggestions.NewMultiSourceSuggestionService(context.Background(), fs, tracker, []string{"hello", "world"}, io.Discard)
+	service, err := suggestions.NewMultiSourceSuggestionService(context.Background(), fs, tracker, []string{"hello", "world"}, io.Discard, infra_persistence.NewWorkspacePolicy())
 	if err != nil {
 		t.Fatalf("failed to create service: %v", err)
 	}
@@ -113,7 +114,7 @@ func TestMultiSourceSuggestionService_GetSuggestions(t *testing.T) {
 func TestMultiSourceSuggestionService_ContextCancellation(t *testing.T) {
 	tracker := &mockPromptTracker{}
 	fs := persistence.NewMockFileSystem()
-	service, _ := suggestions.NewMultiSourceSuggestionService(context.Background(), fs, tracker, nil, io.Discard)
+	service, _ := suggestions.NewMultiSourceSuggestionService(context.Background(), fs, tracker, nil, io.Discard, infra_persistence.NewWorkspacePolicy())
 
 	// Create many files in mock FS to make scan potentially slow (though mock is fast)
 	tmpDir := "/tmp/test"
@@ -133,7 +134,7 @@ func TestMultiSourceSuggestionService_ContextCancellation(t *testing.T) {
 func TestMultiSourceSuggestionService_FileSystemSearch(t *testing.T) {
 	tracker := &mockPromptTracker{}
 	fs := persistence.NewMockFileSystem()
-	service, _ := suggestions.NewMultiSourceSuggestionService(context.Background(), fs, tracker, nil, io.Discard)
+	service, _ := suggestions.NewMultiSourceSuggestionService(context.Background(), fs, tracker, nil, io.Discard, infra_persistence.NewWorkspacePolicy())
 
 	// Create some test files in mock FS
 	tmpDir := "/tmp/test"
@@ -177,7 +178,7 @@ func TestMultiSourceSuggestionService_RecordPrompt(t *testing.T) {
 		appended: appended,
 	}
 	fs := persistence.NewMockFileSystem()
-	service, _ := suggestions.NewMultiSourceSuggestionService(context.Background(), fs, tracker, nil, io.Discard)
+	service, _ := suggestions.NewMultiSourceSuggestionService(context.Background(), fs, tracker, nil, io.Discard, infra_persistence.NewWorkspacePolicy())
 
 	prompt := "new-unique-prompt"
 	err := service.RecordPrompt(context.Background(), prompt)
@@ -270,7 +271,7 @@ func TestMultiSourceSuggestionService_MergeSuggestions(t *testing.T) {
 func TestSuggestionService_ScanFiles_CancelledContext(t *testing.T) {
 	tracker := &mockPromptTracker{}
 	fs := persistence.NewMockFileSystem()
-	service, _ := suggestions.NewMultiSourceSuggestionService(context.Background(), fs, tracker, nil, io.Discard)
+	service, _ := suggestions.NewMultiSourceSuggestionService(context.Background(), fs, tracker, nil, io.Discard, infra_persistence.NewWorkspacePolicy())
 
 	// Create some files in mock FS
 	tmpDir := "/tmp/test"
@@ -301,7 +302,7 @@ func TestSuggestionService_ScanFiles_CancelledContext(t *testing.T) {
 func TestSuggestionService_ScanFiles_InvalidDir(t *testing.T) {
 	tracker := &mockPromptTracker{}
 	fs := persistence.NewMockFileSystem()
-	service, _ := suggestions.NewMultiSourceSuggestionService(context.Background(), fs, tracker, nil, io.Discard)
+	service, _ := suggestions.NewMultiSourceSuggestionService(context.Background(), fs, tracker, nil, io.Discard, infra_persistence.NewWorkspacePolicy())
 
 	// Call GetSuggestions with a path that definitely doesn't exist in mock FS
 	invalidPath := "/non-existent-dir/file"
@@ -330,7 +331,7 @@ func TestSuggestionService_LoadTopNFails(t *testing.T) {
 	tracker := &errorPromptTracker{}
 	fs := persistence.NewMockFileSystem()
 	// This should log to stderr but not return an error
-	service, err := suggestions.NewMultiSourceSuggestionService(context.Background(), fs, tracker, nil, io.Discard)
+	service, err := suggestions.NewMultiSourceSuggestionService(context.Background(), fs, tracker, nil, io.Discard, infra_persistence.NewWorkspacePolicy())
 	if err != nil {
 		t.Fatalf("NewMultiSourceSuggestionService should not fail when tracker fails to load: %v", err)
 	}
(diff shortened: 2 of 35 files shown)


gosharplite:
### Architectural Review - PR #336

Strong refactor that addresses the technical debt identified in #331. Centralizing the workspace policy ensures consistent behavior across all traversal tools.

#### Findings:
- **[ARCHITECTURAL]**: Implementation of `WorkspacePolicy` as a domain service correctly enforces a Single Source of Truth for filesystem filtering.
- **[REFACTOR]**: Move `WorkspacePolicy` unit tests from `internal/tools/workspace/utils_test.go` and `internal/tools/developer/release_test.go` to a dedicated `internal/infrastructure/persistence/workspace_policy_test.go`. Infrastructure implementations should be verified within their own package.
- **[NOTE]**: The `ShouldIgnorePath` implementation performs several allocations (`strings.Split`, etc.). While fine for current low-frequency use (release checks), consider a more performant scanning approach if this is ever moved into hot paths like full-repo searches.
- **[NOTE]**: Hidden directory logic `len(name) > 1 && name[0] == '.'` correctly handles `.` but also skips `..`. This is safe for `filepath.Walk` usage.

Verdict: **APPROVED**

gosharplite:
### Architectural Review Response

Agreed on all four findings:

**1. [ARCHITECTURAL]** — The domain service correctly enforces a Single Source of Truth. This eliminates the drift that existed between the five scattered implementations.

**2. [REFACTOR]** — Follow-up issue created: **#337** (Move WorkspacePolicy tests to infrastructure/persistence package). Non-blocking — the tests pass and coverage is maintained in their current locations.

**3. [NOTE] — allocations** — `strings.Split` in `ShouldIgnorePath` is fine for current usage. It's only called by `secretScanner` during release checks, not in hot-path file walks. `ShouldIgnoreDir` (the hot path) is a simple map lookup + byte check with zero allocations.

**4. [NOTE] — hidden dir logic** — Correct. `len(name) > 1 && name[0] == '.'` safely handles `.`, `..`, and any hidden directory. The `..` guard is defensive since `filepath.Walk` never passes it, but harmless.

**Verdict: Ready to merge.** 🚀

gosharplite:
Here is the architectural review based on the system design and scalability SOP.

### 1. Scalability: Allocations in Hot Paths [TECHNICAL DEBT]
The `ShouldIgnorePath(path string)` method is designed to be evaluated for every file during full-workspace operations (like `secretScanner`). Currently, it performs heavy memory allocations per file check.
* **Issue**: `filepath.ToSlash`, `filepath.Clean`, and `strings.Split` create multiple heap allocations per call. Over a large repository, this generates significant garbage collection (GC) pressure and slows down tree walks.
* **Solution**: Refactor to a zero-allocation approach by checking the extensions first and iterating over the path string using `strings.IndexAny`.

**Current:**
```go
func (p *defaultWorkspacePolicy) ShouldIgnorePath(path string) bool {
	normalized := filepath.ToSlash(filepath.Clean(path))
	parts := strings.Split(normalized, "/")
    // ...
}
```

**Scalable:**
```go
func (p *defaultWorkspacePolicy) ShouldIgnorePath(path string) bool {
	// 1. Fast path: check leaf extensions first
	for _, ext := range ignoredExtensions {
		if strings.HasSuffix(path, ext) {
			return true
		}
	}

	// 2. Check path components without splitting/allocating
	rest := path
	for {
		i := strings.IndexAny(rest, "/\\")
		if i == -1 {
			return p.ShouldIgnoreDir(rest)
		}
		if p.ShouldIgnoreDir(rest[:i]) {
			return true
		}
		rest = rest[i+1:]
	}
}
```

### 2. Modularity: Hexagonal Architecture Package Layout [REFACTOR]
The `defaultWorkspacePolicy` implementation is located in `internal/infrastructure/persistence/workspace_policy.go`.
* **Issue**: A purely logical string-matching policy has no dependency on I/O, `os`, or the file system adapter. Placing pure business logic in the `infrastructure/persistence` layer violates separation of concerns.
* **Recommendation**: Unless there are immediate plans to make this policy stateful (e.g., dynamically reading `.gitignore` from disk using the `FileSystem` interface), this implementation should live in `internal/application/policy` or be a concrete implementation directly in `internal/domain/services`. 

### 3. Analyze & Consistency [POSITIVE]
The implementation perfectly adheres to Dependency Injection rules, correctly threading the `WorkspacePolicy` down through the `Bootstrapper` to individual managers. Consolidating the scattered `isIgnoredDir` variants across suggestions, secret scanning, dependency analysis, and file searching establishes a single source of truth, highly improving domain consistency.

gosharplite:
### Second Architectural Review — Response

Mixed agreement. Here's the breakdown:

---

**1. Scalability — Partially disagree**

The concern is valid in principle, but the premise is incorrect:

- `ShouldIgnorePath` is **not in a hot path**. It's called only by `secretScanner.isIgnored` during `verify_release_readiness` — a manual, infrequent operation. The hot path is `ShouldIgnoreDir` (called per directory entry during every walk), which is already zero-allocation: map lookup + one byte check.
- The proposed replacement has bugs:
  1. **No path normalization** — `foo/../.git` wouldn't be caught since `..` isn't resolved.
  2. **Lost directory-in-path checks** — after the extension fast-path, the loop calls `ShouldIgnoreDir` on every component. But the original code also relied on `ShouldIgnoreDir` catching directory names; if a path is `src/vendor/foo.go`, the loop would catch `vendor` via `ShouldIgnoreDir` — that part works. However, the proposed code drops `filepath.Clean` normalization, so paths with redundant separators or traversal would bypass the check.
  3. **Index panic** — `rest[i+1:]` panics if the path ends with a trailing slash.

If this ever moves to a hot path, the correct approach would be a careful zero-alloc scanner. Not needed today.

---

**2. Modularity — Agree in principle, defer**

The reviewer is correct that `defaultWorkspacePolicy` is pure string matching with no I/O dependency. It doesn't strictly belong in `infrastructure/persistence`.

Two options for a follow-up:

| Option | Location | Rationale |
|--------|----------|-----------|
| A | `internal/application/policy/` | Clean Architecture — application layer for use-case-agnostic implementations |
| B | Stay | Not worth the churn for a pure function |

**Recommendation**: Option A as a follow-up **if** `.gitignore` integration is ever added (mentioned in the original #331). At that point the policy would depend on `persistence.FileSystem` and would naturally move to `infrastructure/persistence`. For now, it's a pure function — location is cosmetic.

---

**3. Consistency — Agree** ✅

The DI threading through `Bootstrapper` → `sessionDeps` → `ToolRegistrationParams` → individual managers is clean and consistent. No notes.

---

**Verdict**: PR remains ready to merge. Finding #1 is based on an incorrect hot-path premise. Finding #2 is architecturally valid but non-blocking. Finding #3 is positive confirmation.

gosharplite:

1. **Scalability**: You are absolutely right; since `ShouldIgnorePath` is only invoked during infrequent release checks, optimizing for allocations here is premature optimization. Good catch on the edge cases (unnormalized paths and trailing slashes) in the zero-allocation suggestion—correctness definitely trumps minor performance gains here.
2. **Modularity**: Anticipating the future need to parse `.gitignore` files via `persistence.FileSystem` makes `infrastructure/persistence` a perfectly reasonable home for this policy for now. Avoiding unnecessary churn is the right call.

Excellent architectural pragmatism. LGTM!

gosharplite:
## 📋 Architectural Review

Strong refactor overall — clean Hexagonal split (`domain/services` port ↔ `infrastructure/persistence` adapter), small focused interface, stateless impl, and every consumer signature change has matching test updates. However, I found one **runtime-breaking wiring gap** that must be fixed before merge.

---

### 🔴 [ARCHITECTURAL BLOCKER] Nil-Policy in `toolchain_factory.go`

**File:** `internal/infrastructure/di/toolchain_factory.go` (~L59, `BuildRegistry`)

`WorkspacePolicy` is threaded through `Bootstrapper`, `sessionDeps`, and `ToolRegistrationParams`, **but the live struct literal that builds `ToolRegistrationParams` never assigns it**:

```go
regParams := infra_tools.ToolRegistrationParams{
    Registry:         reg,
    SecurityManager:  f.SM,
    // ...
    FileSystem:       infra_persistence.NewDomainFS(f.FileSystem),
    HealthManager:    params.HealthManager,
    // ❌ WorkspacePolicy: <missing>
}
```

**Impact:** `params.WorkspacePolicy` is `nil` at runtime. The first invocation of any of these tools will panic with a nil-pointer dereference:

- `search_files`, `grep_definitions` → `workspace.ConcurrentSearch`
- `find_file` → `walkAndProcess`
- `get_project_summary` → `infoManager.makeWalkFunc`
- `verify_release_readiness` → `secretScanner.isIgnored`
- `get_package_graph` → `defaultDependencyAnalyzer.listInternalPackages`

Tests don't catch this because every unit test constructs `ToolRegistrationParams` (or the manager structs) explicitly with `infra_persistence.NewWorkspacePolicy()`. There's no integration test exercising the live `BuildRegistry` path.

**Required fix:**

1. Add `WorkspacePolicy services.WorkspacePolicy` to `defaultToolchainFactory` and to `newToolchainFactory` parameters.
2. Pass `b.WorkspacePolicy` from `Bootstrapper` to `newToolchainFactory` in `container.go`.
3. Set `WorkspacePolicy: f.WorkspacePolicy` in the `BuildRegistry` literal.
4. Add a defensive guard at the top of `RegisterAll`:
   ```go
   if params.WorkspacePolicy == nil {
       return fmt.Errorf("RegisterAll: WorkspacePolicy is required")
   }
   ```
5. Add a regression test that calls `BuildRegistry` end-to-end and invokes `find_file` on a tiny temp dir, asserting no panic.

---

### 🟡 [TECHNICAL DEBT] `NewWorkspacePolicy` Returns Unexported Type

**File:** `internal/infrastructure/persistence/workspace_policy.go:23`

```go
func NewWorkspacePolicy() *defaultWorkspacePolicy { ... }
```

Returning an unexported pointer from an exported constructor triggers `revive`/`golint` (`unexported-return`). Since callers only ever consume it via the interface, return the interface:

```go
func NewWorkspacePolicy() services.WorkspacePolicy {
    return &defaultWorkspacePolicy{}
}
```

---

### 🟡 [TECHNICAL DEBT] Behavior Change in `get_project_summary` Needs Note

The old `info.go` rule only skipped `.*`, `testdata`, `vendor`, `node_modules`. The new unified policy **also** skips `bin`, `obj`, `output`, `dist`, `configs`. This is a deliberate widening but produces user-visible LOC/file-count differences. Please add a CHANGELOG entry (and ideally an ADR) so users aren't surprised by lower counts after upgrade.

---

### 🟡 [REFACTOR — Low Priority] `ShouldIgnorePath` Allocations

`filepath.Clean` + `strings.Split` allocate per call. Fine at current scale; consider an `IndexByte`-based component walker if profiling on large repos (>10k files) flags it.

---

### 🟢 [OBSERVATION] Missing ADR

This change unifies semantics across 5 subsystems and expands the ignore set for several. Recommend adding `docs/adr/00XX-workspace-policy.md` documenting:
1. The decision (single canonical ignore set).
2. The trade-off (more aggressive skipping in some tools).
3. Extension policy (modify `ignoredDirNames`/`ignoredExtensions`, never re-add per-caller lists).

---

### 📊 Summary

| Severity | Count | Must-fix? |
|---|---|---|
| 🔴 ARCHITECTURAL BLOCKER | 1 | **YES** |
| 🟡 TECHNICAL DEBT | 2 | Strongly recommended |
| 🟡 REFACTOR | 1 | Optional |
| 🟢 OBSERVATION | 1 | Recommended |

**Verdict:** ❌ **Request Changes.** Architecturally sound; merge-blocked by the `toolchain_factory.go` wiring gap. Once `WorkspacePolicy` is plumbed through `BuildRegistry` and guarded in `RegisterAll`, this is good to go.

gosharplite:
### Third Architectural Review — Response

**Confirmed.** The reviewer found a real runtime nil-pointer panic. I verified against the code.

---

**🔴 ARCHITECTURAL BLOCKER — Agree**

`toolchain_factory.go:59` — `BuildRegistry` constructs `ToolRegistrationParams` with every field set **except** `WorkspacePolicy`, which stays `nil` (zero value for interface). Impact chain:

```
BuildRegistry → RegisterAll → workspace.Register → registerFiles
  → fileReader.policy = nil
    → findFile → walkAndProcess → shouldSkipEntry
      → policy.ShouldIgnoreDir(name)  💥 PANIC
```

All 5 tools that consume the policy are affected: `search_files`, `grep_definitions`, `find_file`, `get_project_summary`, `verify_release_readiness`, `get_package_graph`.

Tests don't catch this because every unit test constructs params with `NewWorkspacePolicy()` explicitly — the live `BuildRegistry` path has no integration coverage.

**Required fix (4 lines + guard):**

1. Add `WorkspacePolicy services.WorkspacePolicy` field to `defaultToolchainFactory`
2. Add to `newToolchainFactory` parameters
3. Pass `b.WorkspacePolicy` from `NewBootstrapper` call site in `container.go`
4. Set `WorkspacePolicy: f.WorkspacePolicy` in the `BuildRegistry` struct literal
5. Nil-guard in `RegisterAll`: `if params.WorkspacePolicy == nil { return fmt.Errorf(...) }`

---

**🟡 NewWorkspacePolicy returns unexported type — Agree**

Change signature from `*defaultWorkspacePolicy` to `services.WorkspacePolicy`. Callers consume via interface anyway.

---

**🟡 Behavior change needs note — Agree**

`get_project_summary` now additionally skips `bin`, `obj`, `output`, `dist`, `configs`. LOC/file counts will differ from previous releases. Worth a CHANGELOG line.

---

**🟢 Missing ADR — Agree**

5 subsystems unified + expanded ignore set = warrants `docs/adr/00XX-workspace-policy.md`.

---

**Verdict: ❌ Request Changes.** Merge-blocked by the `toolchain_factory.go` wiring gap only. The other findings are non-blocking recommendations. Once the nil-policy is plumbed through and guarded, this is good to go.

gosharplite:
All findings from the third architectural review are now resolved:

| # | Finding | Severity | Fix |
|---|---------|----------|-----|
| 1 | `toolchain_factory.go` nil-policy panic | 🔴 BLOCKER | ✅ Plumbed through constructor → `BuildRegistry`; nil-guard in `RegisterAll` (8a221e98) |
| 2 | `NewWorkspacePolicy` unexported return | 🟡 Debt | ✅ Returns `services.WorkspacePolicy` interface (f812fb58) |
| 3 | Test files missing `WorkspacePolicy` arg | — | ✅ `factory_error_test.go`, `tools_test.go` updated (f22ca7f2) |

**Remaining non-blocking items** (documentation, not code):
- 🟡 Behavior change CHANGELOG entry (deferred)
- 🟢 ADR for workspace policy (deferred)
- 🟡 Test migration to `workspace_policy_test.go` (captured in #337)

**Verdict: Ready to merge.** 🚀

gosharplite:
## ✅ Final Review — Fixes Verified

I verified the resolution commits against the live code on `refactor/331-workspace-policy`.

### Verification

| Commit | Claim | Verified |
|---|---|---|
| `8a221e98` | Plumb `WorkspacePolicy` through `defaultToolchainFactory` → `BuildRegistry`; nil-guard in `RegisterAll` | ✅ Field added, `newToolchainFactory` signature updated, `container.go` passes `b.WorkspacePolicy` (correctly ordered **after** the nil-default at L93–95), literal sets `WorkspacePolicy: f.WorkspacePolicy`, `RegisterAll` returns clear error if nil |
| `f812fb58` | `NewWorkspacePolicy` returns interface | ✅ Now returns `services.WorkspacePolicy` |
| `f22ca7f2` | Test params updated for new signatures | ✅ `factory_error_test.go`, `tools_test.go` |
| Build/Test | `go build ./...` and `go test ./internal/infrastructure/di/... ./internal/tools/...` | ✅ All green |

### One non-blocking gap

The **integration test** I recommended (asserting `BuildRegistry` → `find_file` doesn't panic end-to-end) was not added. The `RegisterAll` nil-guard provides good fail-fast defense, but a real wiring test would catch the *next* analogous gap. Worth filing as a follow-up alongside the deferred CHANGELOG/ADR.

### On the second architectural review

The pushback on Finding #1 (allocation optimization) was correct engineering judgment — `ShouldIgnorePath` is invoked only from `secretScanner` during release checks, not in hot paths, and the proposed zero-alloc replacement had two real defects:

1. Dropped `filepath.Clean` normalization → `foo/../.git` would slip through.
2. `rest[i+1:]` would panic on paths with trailing separators.

Premature optimization with regressions is worse than the status quo. Good call rejecting it.

Finding #2 (move out of `infrastructure/persistence`) remains architecturally valid — pure string logic doesn't strictly belong in a persistence package — but deferring is acceptable **if** `.gitignore` integration is on the roadmap (which would introduce a `FileSystem` dependency and naturally justify the location).

### Verdict

**LGTM 🚀** — merge-blocking issues resolved, non-blocking items appropriately captured (#337, CHANGELOG, ADR pending).

PR #345, opened 2026-05-12
test(openai): replace brittle 127.0.0.1:1 mocking with httptest guard pattern

gosharplite:
## Summary

Replace `http://127.0.0.1:1` unreachable-endpoint mocking with `httptest.NewServer` + `t.Fatal` guard pattern in the OpenAI LLM provider test suite.

Closes #343

## Problem

Tests that verify validation errors occur **before** any HTTP request used an unreachable address (`http://127.0.0.1:1`) to ensure the HTTP layer is never hit. If validation logic is accidentally reordered to occur **after** the request, the test would produce a confusing `connection refused` error instead of a clear validation failure message.

## Changes

| Test | Before | After |
|------|--------|-------|
| `TestSendChat_EmptyToolResponseID` | `NewClient("http://127.0.0.1:1", ...)` | `httptest.NewServer` + `t.Fatal` guard |
| `TestSendChat_EmptyToolCallID` | `NewClient("http://127.0.0.1:1", ...)` | `httptest.NewServer` + `t.Fatal` guard |
| `TestSendChat_MarshalToolResponseError` | `NewClient("http://127.0.0.1:1", ...)` | `httptest.NewServer` + `t.Fatal` guard |

### Anthropic — no changes needed

`TestPrepareRequest_ToAnthropicMessagesError` already calls `prepareAnthropicRequest` directly at the unit level and never touches the network.

## Verification

- `go test ./internal/infrastructure/llm/openai/...` — **PASS**
- `go test ./internal/infrastructure/llm/anthropic/...` — **PASS**
- `grep -r "127.0.0.1:1" internal/infrastructure/llm/` — **zero matches**
- Zero production code changes

### DIFF
--- a/internal/infrastructure/llm/openai/client_test.go
+++ b/internal/infrastructure/llm/openai/client_test.go
@@ -509,9 +509,11 @@ func TestSendChat_EmptyToolResponseID(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 
-			// Use an unreachable server — the error occurs during
-			// request construction, before the HTTP call is made.
-			c := NewClient("http://127.0.0.1:1", tt.model,
+			server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				t.Fatal("HTTP server should not have been reached; validation must fail first")
+			}))
+			defer server.Close()
+			c := NewClient(server.URL, tt.model,
 				&auth.BearerAuth{Token: "test-key"},
 				WithHeaders(tt.headers),
 			)
@@ -577,9 +579,11 @@ func TestSendChat_EmptyToolCallID(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 
-			// Use an unreachable server — the error occurs during
-			// request construction, before the HTTP call is made.
-			c := NewClient("http://127.0.0.1:1", tt.model,
+			server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				t.Fatal("HTTP server should not have been reached; validation must fail first")
+			}))
+			defer server.Close()
+			c := NewClient(server.URL, tt.model,
 				&auth.BearerAuth{Token: "test-key"},
 				WithHeaders(tt.headers),
 			)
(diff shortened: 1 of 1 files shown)


gosharplite:
### Architectural Review

**Status: Approved (with follow-up recommendation)**

This PR successfully removes the brittle `127.0.0.1:1` magic string and aligns the validation tests with the established `httptest.NewServer` pattern used throughout the test suite.

**[TECHNICAL DEBT / REFACTOR] Recommendation for Future Iteration:**
While this is a solid immediate fix, spinning up an `httptest.NewServer` (which allocates OS-level TCP sockets) inside a `t.Parallel()` loop just to ensure a request *isn't* made adds test overhead. 

Currently, `NewClient` hardcodes the HTTP client instantiation (`c.httpClient = &http.Client{...}`). By introducing a functional option like `WithHTTPClient(*http.Client)` or `WithTransport(http.RoundTripper)`, we can inject an in-memory mock transport instead:

```go
type mockFailTransport struct { t *testing.T }
func (m mockFailTransport) RoundTrip(*http.Request) (*http.Response, error) {
    m.t.Fatal("Validation must fail first")
    return nil, nil
}

// Usage purely in-memory:
c := NewClient("http://mock", model, auth, 
    WithHTTPClient(&http.Client{Transport: mockFailTransport{t}}),
)
```

This enforces better Dependency Injection and keeps pure validation tests off the network layer entirely. Consider this for a follow-up chore!

gosharplite:
Thanks for the review. A few thoughts:

**On the fix itself** — agreed, this is the right change for this PR. The `t.Fatal` guard makes test intent explicit and eliminates the ambiguous `connection refused` failure mode.

**On the `WithHTTPClient` suggestion** — I see the merit as a general-purpose enhancement (it would also clean up `TestSendChat_ErrorBodyReadFailure` which currently reaches into `c.httpClient.Transport` directly). However, I'd push back on characterizing the current `httptest.NewServer` approach as technical debt:

1. **It's the canonical Go pattern.** `net/http/httptest` was designed for exactly this — "handler should never be reached" guards are idiomatic. The stdlib's own tests use this pattern pervasively in parallel.
2. **It tests more surface.** A mock `RoundTripper` bypasses URL parsing, transport init, and the full HTTP client path. `httptest.NewServer` keeps these at the integration level. Both strategies are valid; it's a trade-off, not a clear improvement.
3. **The overhead is negligible.** Loopback ephemeral port allocation is not a meaningful cost for a test suite of this size.

**Proposal:** Land this PR as-is. Spin up a follow-up issue for `WithHTTPClient` as a `[REFACTOR]` (not `[TECHNICAL DEBT]`) — it's a nice-to-have generalization, not something that's currently causing problems.

gosharplite:
### Architectural Follow-up

I completely agree with your assessment. Thank you for the thoughtful pushback!

Your points are architecturally sound:
1. **Test Surface Area:** You are absolutely right that `httptest.NewServer` provides a stronger guarantee by exercising the actual `http.Client` request construction pipeline up to the network boundary. A mock `RoundTripper` would bypass that, weakening the integration test.
2. **Classification:** I agree with downgrading the recommendation from `[TECHNICAL DEBT]` to a `[REFACTOR]`. The current approach is idiomatic Go, robust, and the overhead is negligible. 

Let's proceed with your proposal: land this PR as-is. 

Creating a follow-up issue for `WithHTTPClient` as a general-purpose `[REFACTOR]` is a great idea—especially since it will allow deterministic failure injection (e.g., simulating timeouts or connection drops) for edge-case tests like `TestSendChat_ErrorBodyReadFailure`.

Great work!

PR #346, opened 2026-05-12
refactor(di): eliminate toolchain abstraction leak from HealthFactory

thptcnec:
## What

Moves toolchain health checker construction from `healthFactory` into `toolchainFactory`, which already owns all other toolchain wiring including `exec.RealExecutor` instantiation.

## Why

`health_factory.go` had an abstraction leak — the DI orchestration layer knew about concrete toolchain implementation details:
- `exec.RealExecutor` (a concrete infrastructure type)
- The binary list `["git", "go"]` and `["make"]`

This violated Separation of Concerns: any change to the toolchain's health check (new binaries, different executor) required modifying the DI layer.

## Changes

| File | Change |
|------|--------|
| `toolchain_factory.go` | + `BuildHealthChecker() ports.HealthChecker` to interface & impl |
| `health_factory.go` | Accepts `toolchainFactory`, delegates; removes `exec` + `infra_toolchain` imports |
| `container.go` | Passes `b.toolchainFactory` through to `BuildHealthManager` |
| `lazy_test.go` | Mock + stub for new interface method |

## Verification

- `go build ./...` ✅
- `go vet ./...` ✅
- `go test ./... -count=1` ✅ (62/62 packages)
- Linter ✅

Closes #344

### DIFF
--- a/internal/infrastructure/di/container.go
+++ b/internal/infrastructure/di/container.go
@@ -154,7 +154,7 @@ func (b *Bootstrapper) BuildSessionDependencies(ctx stdctx.Context, cfg *config.
 		lazyClient:       lazyClient,
 	}
 
-	deps.health = b.healthFactory.BuildHealthManager(cfg, sessionProvider, lazyClient)
+	deps.health = b.healthFactory.BuildHealthManager(cfg, sessionProvider, lazyClient, b.toolchainFactory)
 
 	lazyRegistry := newLazyRegistry(func() (tools.Registry, error) {
 		return b.toolchainFactory.BuildRegistry(toolchainParams{
--- a/internal/infrastructure/di/health_factory.go
+++ b/internal/infrastructure/di/health_factory.go
@@ -4,15 +4,13 @@ import (
 	"github.com/gosharplite/tell-me-go/internal/domain/config"
 	"github.com/gosharplite/tell-me-go/internal/domain/llm"
 	"github.com/gosharplite/tell-me-go/internal/domain/ports"
-	"github.com/gosharplite/tell-me-go/internal/infrastructure/exec"
 	"github.com/gosharplite/tell-me-go/internal/infrastructure/factory"
 	infra_llm "github.com/gosharplite/tell-me-go/internal/infrastructure/llm"
-	infra_toolchain "github.com/gosharplite/tell-me-go/internal/infrastructure/toolchain"
 )
 
 // healthFactory defines the interface for building health check components.
 type healthFactory interface {
-	BuildHealthManager(cfg *config.Config, sessionProvider ports.SessionProvider, lazyClient llm.ExtendedClient) ports.HealthCheckManager
+	BuildHealthManager(cfg *config.Config, sessionProvider ports.SessionProvider, lazyClient llm.ExtendedClient, tf toolchainFactory) ports.HealthCheckManager
 }
 
 // defaultHealthFactory implements healthFactory.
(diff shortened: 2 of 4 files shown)


thptcnec:
### Architectural Review

**Status:** Approved for architectural integrity.
No `[ARCHITECTURAL BLOCKER]`s detected. The PR actively reduces technical debt by improving component cohesion.

#### Findings

**[REFACTOR] Inconsistent Factory Delegation for LLM Provider**
* **Context**: While `CompToolchain` creation was cleanly moved to `toolchainFactory` and `CompPersistence` is delegated to `sessionProvider`, `CompLLMProvider` is still manually constructed inside `health_factory.go` (via `infra_llm.CreateAuthenticator` and `infra_llm.NewLLMProviderHealthChecker`).
* **Current Design**: `healthFactory` retains a tight coupling to LLM infrastructure specifics (knowing about authenticators and provider types), violating the exact principle this PR aims to solve for the toolchain.
* **Scalable Design**: Abstract the LLM health checker instantiation into an `llmFactory` (or similar component) to standardize the health manager bootstrapping.
* **Example**:
  ```go
  // Inside BuildHealthManager:
  healthCheckers := map[ports.Component]ports.HealthChecker{
      ports.CompPersistence: sessionProvider.GetHealthChecker(),
      ports.CompToolchain:   tf.BuildHealthChecker(),
      ports.CompLLMProvider: llmFactory.BuildHealthChecker(p, lazyClient), // Delegated
  }
  ```

**[TECHNICAL DEBT] Hardcoded Executor in DI**
* **Context**: `defaultToolchainFactory.BuildHealthChecker()` directly instantiates `&exec.RealExecutor{}`.
* **Current Design**: While standard in the DI root layer, if you ever perform container-level integration testing, you cannot intercept or mock the executor easily.
* **Scalable Design**: For future-proofing, if multiple factories begin requiring command execution, consider introducing a shared `execFactory` or providing the `executor` as a parameter across DI boundaries.

thptcnec:
Thanks for the review. My take on both findings:

---

**On `[REFACTOR] Inconsistent Factory Delegation for LLM Provider` — valid observation, but a different category of concern:**

| | Toolchain (this PR) | LLM Provider (not touched) |
|---|---|---|
| What leaked | System binaries + executor type — knowledge that **belongs inside the toolchain package** | Config-driven params (`p.Type`, `p.URL`, `authenticator`) — knowledge that **naturally lives at the orchestration layer** |
| Pre-existing factory? | `toolchainFactory` already existed with `BuildRegistry` | No `llmFactory` interface exists; `ClientFactory` is a function field on Bootstrapper |

The toolchain leak was an architectural violation (DI knowing about system binaries). The LLM case is an orchestration concern (DI wiring config to infrastructure). Creating an `llmFactory` with `BuildHealthChecker()` is worth tracking as follow-up for consistency, but it's a separate ADR — not the same category as this fix.

---

**On `[TECHNICAL DEBT] Hardcoded Executor in DI` — this is the intended Composition Root design:**

- `&exec.RealExecutor{}` was **already** in `toolchain_factory.go:83` (`BuildRegistry`) before this PR. The PR consolidated it to a single owner instead of duplicating it across two factories. This is an improvement, not debt.
- Only two call sites exist, both in `defaultToolchainFactory`. Introducing an `execFactory` for two usages within the same struct is premature abstraction.
- `exec.RealExecutor` already implements `tools.CommandExecutor` (an interface), so mocking is already possible at the interface boundary. Integration tests that need a different executor can inject it via struct fields — no additional factory needed.

---

**Verdict:** PR should merge as-is. Happy to create follow-up issues for the LLM consistency refactoring if we want to track it.

thptcnec:
I fully agree with your assessment. Your reasoning demonstrates a strong, pragmatic application of Dependency Injection principles.

**On the LLM Provider:** You make an excellent distinction between an architectural leak (system binaries) and an orchestration concern (wiring config to infrastructure). Standardizing the LLM instantiation via an `llmFactory` would indeed be purely for visual consistency rather than fixing an abstraction leak. Treating it as a separate follow-up is the right call.

**On the Hardcoded Executor:** You are completely correct. The Composition Root is exactly where concrete implementations like `&exec.RealExecutor{}` should be bound to their interfaces. Since the underlying components depend on the `tools.CommandExecutor` interface, testability is preserved. Centralizing its instantiation and avoiding an `execFactory` prevents premature abstraction.

I reaffirm the approval. Please proceed with merging as-is.

PR #366, opened 2026-05-13
refactor: migrate analysisManager.DeadCode to unexported + constructor injection

thptcnec:
## Summary

Unexports `analysisManager.DeadCode` and injects `deadCodeAnalyzer` via the constructor. The field was exported solely to enable test mocking — zero production consumers outside the `analysis` package.

## Changes

| File | Change |
|---|---|
| `manager.go` | Unexport `DeadCode` → `deadCode`, add `dc deadCodeAnalyzer` parameter to `newAnalysisManager` |
| `health.go` | Update `checkDeadCode` consumer from `m.Ana.DeadCode` → `m.Ana.deadCode` |
| `registration.go` | Pass `newDeadCodeAnalyzer(sm, idx)` in constructor call |
| `health_test.go` | Inject mock via constructor instead of direct field assignment |
| `manager_test.go` | Pass real analyzer in constructor call |

## Verification

- `go build ./...` — clean
- `go test ./internal/tools/analysis/... -race -count=1` — pass (13.5s)
- `golangci-lint run ./internal/tools/analysis/...` — zero issues
- Zero `.DeadCode` references remain in the codebase

Closes #360

### DIFF
--- a/internal/tools/analysis/health.go
+++ b/internal/tools/analysis/health.go
@@ -250,7 +250,7 @@ func (m *healthManager) checkComplexity(ctx context.Context, hb chan<- struct{})
 }
 
 func (m *healthManager) checkDeadCode(ctx context.Context, hb chan<- struct{}) (string, string) {
-	reports, err := m.Ana.DeadCode.GatherOrphanReports(ctx, ".", hb)
+	reports, err := m.Ana.deadCode.GatherOrphanReports(ctx, ".", hb)
 	if err != nil {
 		return "ERROR", err.Error()
 	}
--- a/internal/tools/analysis/health_test.go
+++ b/internal/tools/analysis/health_test.go
@@ -144,8 +144,8 @@ func TestHealthManager_GetCodeHealth(t *testing.T) {
 	cache := newASTCache(".")
 	mockExec := &mockHealthExecutor{}
 	mockRunner := &mockGoRunner{}
-	ana := newAnalysisManager(idx, cache, sm, nil, mockExec, persistencetest.NewPlainOSFileSystem(), infra_persistence.NewWorkspacePolicy())
-	ana.DeadCode = &mockDeadCodeAnalyzer{}
+	mockDead := &mockDeadCodeAnalyzer{}
+	ana := newAnalysisManager(idx, cache, sm, nil, mockExec, persistencetest.NewPlainOSFileSystem(), infra_persistence.NewWorkspacePolicy(), mockDead)
 	hea := &healthManager{SP: sm, Ana: ana, Exec: mockExec, Runner: mockRunner}
 
 	ctx := context.Background()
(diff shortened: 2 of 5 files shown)


thptcnec:
# Architecture Review: PR #366

Based on the code changes in PR #366, here is the architectural feedback aligned with our SOP guidelines. 

### 1. [REFACTOR] Inconsistent Dependency Injection
**Analysis**: The PR correctly attempts to improve Dependency Injection (DI) by requiring `deadCodeAnalyzer` in the `newAnalysisManager` constructor and privatizing the `deadCode` field. This eliminates the bad practice of mutating exported fields for mocking (as fixed in `health_test.go`). However, it introduces an inconsistent initialization pattern because all other sub-analyzers (`Complexity`, `Package`, `Sequence`, etc.) remain tightly coupled and hardcoded inside the constructor.

**Current (Mixed Initialization)**:
```go
func newAnalysisManager(idx symbolIndex, cache *astCache, sp domain_security.Manager, ..., dc deadCodeAnalyzer) *analysisManager {
    m := &analysisManager{
        Complexity: newComplexityAnalyzer(cache, sp), // Tightly coupled
        Package:    newPackageAnalyzer(cache, sp),    // Tightly coupled
        // ...
        deadCode:   dc,                               // Injected
    }
}
```

### 2. [TECHNICAL DEBT] Constructor Signature Bloat
**Analysis**: The `newAnalysisManager` signature now takes 8 positional parameters. If you apply the same DI treatment to the remaining 6+ analyzers, the constructor signature will become unwieldy and fragile to change.

**Scalable**: Use a `Dependencies` struct or the **Functional Options Pattern** to pass sub-analyzers. This ensures consistent Dependency Injection without polluting the constructor signature.

**Scalable Design Example**:
```go
// Option 1: Dependency Struct
type AnalysisDeps struct {
    DeadCode   deadCodeAnalyzer
    Complexity complexityAnalyzer
    Package    packageAnalyzer
    // ...
}

func newAnalysisManager(idx symbolIndex, cache *astCache, sp domain_security.Manager, deps AnalysisDeps) *analysisManager {
    return &analysisManager{
        deadCode:   deps.DeadCode,
        Complexity: deps.Complexity,
        // ...
    }
}
```

### Summary Recommendation
The intent to improve testability via DI is correct, but applying it piecemeal to only `deadCodeAnalyzer` compromises **Consistency**. 
* **Recommendation**: Consolidate the dependency injection pattern. Either inject all analyzers using a dependencies configuration struct or keep `newDeadCodeAnalyzer` internal to the constructor until a sweeping DI refactor can be applied across all analyzers via a dedicated ADR.

thptcnec:
Thanks for the thorough review. A few thoughts:

**What I agree with:**

- The DI inconsistency across analyzers is real — `deadCode` is now injected while `Complexity`, `Types`, etc. remain hardcoded. That's a legitimate consistency concern.
- Constructor signature bloat is worth tracking. 8 positional params won't scale if all 6+ remaining analyzers follow the same naive pattern.
- A dependency struct or functional options is the right long-term pattern.

**Where I'd push back:**

I don't think the answer is to revert and wait for a sweeping refactor. That's an all-or-nothing stance that leaves the test seam leak (`ana.DeadCode = &mockDeadCodeAnalyzer{}`) in place indefinitely. Incremental improvement is better than no improvement — the PR eliminates a concrete anti-pattern now.

**The consistency problem is already tracked.** Issue #365 captures exactly this: audit all 11 exported fields on `analysisManager`, determine which can be unexported, and apply a unified DI pattern (dependency struct or functional options) across all of them. That's the right venue for the sweeping refactor.

This PR is scoped to #360 — one field, one pattern. The remaining fields follow in #365.

thptcnec:
I completely agree with your reasoning. 

You make a very pragmatic point regarding incremental improvement vs. an "all-or-nothing" approach. Eliminating the concrete test seam leak (`ana.DeadCode = &mock...`) is a tangible win for encapsulation right now.

Given that the broader DI consistency refactor is already tracked in **Issue #365**, this localized change is perfectly acceptable as a stepping stone. Tracking the technical debt in a dedicated issue satisfies the architectural requirement for consistency planning.

Excellent work scoping the change and maintaining momentum. Consider this approved from an architectural perspective!

PR #386, opened 2026-05-14
refactor(capturer): replace persistent worker with per-request goroutine model (fixes #385)

gosharplite:
## Summary

Fixes #385 — the capturer's persistent worker goroutine exited permanently on the first context cancellation, silently breaking all future reads.

## Root Cause

In `internal/ui/capture.go`, the `startWorker` goroutine processed reads via an unbuffered `requestChan`. When the caller's context cancelled during a read, the worker executed `return`, exiting permanently. The channel was never closed or nil'd, causing all subsequent `sendRequest` calls to block forever on the channel send.

## Solution (Option B — Per-Request Goroutine)

Replace the persistent worker + channel architecture with a **per-request goroutine model**:

- Each `sendRequest` spawns its own goroutine that holds `readerMu` for the duration of the blocking read
- Context cancellation returns immediately; the drained read goroutine completes its syscall naturally
- `closed` and `needsReset` use `atomic.Bool` — prevents data races and the `Close` deadlock
- `bufio.Reader.Reset()` discards buffered data from cancelled reads
- **Self-healing**: the next read succeeds once the drained goroutine finishes

## Architecture Decision

Documented in ADR-035: `docs/adr/2026-05-capturer-per-request-goroutine.md`

| Aspect | Before (Worker) | After (Per-Request Goroutine) |
|--------|-----------------|-------------------------------|
| Primitives | 1 mutex, 2 channels, 1 persistent goroutine | 1 mutex, 2 atomics, 0 persistent goroutines |
| After ctx cancel | Worker exits; all future reads deadlock | Read goroutine drains; next read succeeds |
| `Close` behavior | Blocks on `readerMu` (deadlock risk) | Returns immediately via `atomic.Bool` |
| `bufio.Reader` safety | Implicit (worker serialization) | Explicit (`readerMu` + reset on cancel) |

## Changes

| File | Change |
|------|--------|
| `internal/ui/capture.go` | Remove worker, `requestChan`, `done`; implement per-request goroutine `sendRequest`; use `atomic.Bool` |
| `internal/ui/capture_test.go` | Add `TestCapturer_ReadAfterCancellation` regression test; update `TestCapturer_Close_ContextCancelled` |
| `internal/cli/chat_command_test.go` | Fix testify/mock reflection race: don't pass live `capturer` pointer to `m.Called()` |
| `docs/adr/2026-05-capturer-per-request-goroutine.md` | ADR-035 documenting decision and trade-offs |

## Verification

| Check | Result |
|-------|--------|
| `go build ./...` | ✅ |
| `go vet ./...` | ✅ |
| Lint (golangci-lint) | ✅ 0 issues |
| `go test ./...` | ✅ ALL PASS |
| `go test -race ./...` | ✅ ALL PASS, 0 races |
| Dead code | ✅ none |
| Vulncheck | ✅ 0 vulnerabilities |
| Coverage (`internal/ui`) | ✅ 93.2% |
| Regression test (cancel → read again) | ✅ 5/5 consistent |

### DIFF
--- a/docs/adr/2026-05-capturer-per-request-goroutine.md
+++ b/docs/adr/2026-05-capturer-per-request-goroutine.md
@@ -1,0 +1,124 @@
+<!--
+Copyright (c) 2026 gosharplite@gmail.com
+SPDX-License-Identifier: MIT
+-->
+
+# ADR-035: Per-Request Goroutine Model for Capturer Reads
+
+## Status
+Accepted
+
+## Decision Date
+2026-05-14
+
+## Context
+
+The capturer (`internal/ui/capture.go`) previously used a persistent worker
+goroutine to serialize reads on `c.reader` (a `*bufio.Reader`, which is not
+concurrency-safe). The worker received read requests via an unbuffered
+`requestChan`, spawned a nested read goroutine, and raced the caller's context
+against the blocking read syscall.
+
+**Problem:** When the caller's context cancelled during a read, the worker's
+context branch executed `return`, exiting the entire worker goroutine. This
+was intentional — the nested read goroutine leaked, holding `c.reader`, and
+returning prevented a second request from racing on the corrupted reader.
+
+However, the worker exit was permanent:
+
+1. `c.requestChan` was never closed or nil'd in this path.
+2. Any subsequent `sendRequest` call passed the nil-check (channel still
+   non-nil), unblocked on `c.requestChan <- req`, and blocked forever since
+   no receiver existed.
+3. The caller eventually hit its own context timeout, returning a misleading
+   `context.DeadlineExceeded` instead of a clear `errCapturerClosed`.
+
+A single Ctrl+C during one prompt silently disabled the capturer for the
+rest of the process lifetime (issue #385).
+
+### Options Considered
+
+**Option A — Fast-fail subsequent requests:** On worker exit, atomically nil
+`c.requestChan` under `c.readerMu` so subsequent `sendRequest` calls return
+`errCapturerClosed`. Rejected: unresolvable race between nil'ing the channel
+and an in-flight channel send.
+
+**Option B — Per-request goroutine model:** Drop the persistent worker.
+Each `sendRequest` spawns its own goroutine that holds `readerMu` for the
+duration of the blocking read. Context cancellation returns immediately; the
+read goroutine drains naturally. **Accepted.**
+
+**Option C — Resilient worker:** Keep the worker alive after cancellation by
+detaching the leaked read goroutine through a "drain" channel. Rejected: adds
+a third concurrency primitive (channel, mutex, drain channel) without benefit
+over Option B.
+
+## Decision
+
+Replace the persistent worker goroutine + `requestChan` architecture with a
+**per-request goroutine model**:
+
+```
+sendRequest(ctx, req):
+  1. c.closed.Load() → return errCapturerClosed if closed (atomic, no lock)
+  2. c.readerMu.Lock(); reader := c.reader; c.readerMu.Unlock()
+  3. go func():
+       c.readerMu.Lock()
+       defer c.readerMu.Unlock()
+       if c.needsReset.Swap(false): reader.Reset(c.Stdin)
+       perform blocking read on reader
+       readDone <- result
+  4. select { case <-readDone: return result; case <-ctx.Done(): }
+     → c.needsReset.Store(true); return ctx.Err()
+```
+
+### Key Design Decisions
+
+| Decision | Rationale |
+|----------|-----------|
+| `readerMu` serializes access to `c.reader` | `bufio.Reader` is not concurrency-safe; only one goroutine may read at a time |
+| `closed` and `needsReset` are `atomic.Bool` | Cancellation path cannot acquire `readerMu` (held by draining goroutine); atomics prevent data races and deadlocks |
+| `Close` does not acquire `readerMu` | `atomic.Bool` idempotency check avoids blocking on an in-flight read goroutine; `Stdin.Close()` unblocks the syscall |
+| `bufio.Reader.Reset()` after cancellation | Cancelled `ReadString` may leave partial data in the internal buffer; resetting discards it for the next read |
+| Self-healing via natural drain | The leaked goroutine from a cancelled read eventually completes its syscall, releases `readerMu`, and exits; the next read proceeds normally |
+
+## Consequences
+
+### Positive
+- **Self-healing.** After a context-cancelled read, the next read on the same
+  capturer succeeds once the drained goroutine's syscall completes.
+- **Deterministic close semantics.** `Close` returns immediately regardless of
+  in-flight reads. Idempotent via `atomic.Bool`.
+- **No persistent goroutine to manage.** No worker lifecycle, no channel teardown,
+  no channel-close-during-send panics.
+- **Zero data races.** Verified with `go test -race`. The two atomics eliminate
+  the lock ordering problems that plagued the worker model.
+- **Simpler concurrency model.** Two primitives: one mutex (serialization) and
+  two atomics (state flags). Down from: one mutex, one unbuffered channel, one
+  done channel, and a persistent goroutine.
+
+### Negative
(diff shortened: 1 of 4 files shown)


gosharplite:
Here is the Architectural Review for PR 386 based on the SOP guidelines:

### Architecture SOP Review

**1. Analyze**
- **Pattern**: Migration from a persistent worker goroutine to a per-request goroutine model for I/O serialization.
- **Dependencies**: The `capturer` orchestrates blocking I/O on `Stdin` (via `bufio.Reader`), requiring strict serialization.
- **Concurrency primitives**: Uses `sync.Mutex` (`readerMu`) for serialization and `atomic.Bool` (`closed`, `needsReset`) to prevent data races during cancellation.

**2. Scalability (Critical Finding)**
- **[ARCHITECTURAL BLOCKER] Synchronous Blocking on Hot Path:**
  In `sendRequest`, you are synchronously acquiring `c.readerMu` on the caller's goroutine before spawning the background worker:
  ```go
  c.readerMu.Lock()
  reader := c.reader
  c.readerMu.Unlock()
  ```
  **The Impact:** If a previous read was cancelled, its drained goroutine is still running and holding `c.readerMu` until the I/O syscall completes. Consequently, the next call to `sendRequest` will **block indefinitely at `c.readerMu.Lock()`**. Because this block occurs *before* the `select` statement, the new request will completely ignore its own `ctx.Done()` timeout until it acquires the lock. This defeats the primary goal of the ADR.

  **Scalable Design:**
  The `c.reader` pointer itself is immutable after `NewCapturer` initialization. Only its internal state changes (via `Reset`). Therefore, reading the pointer outside the goroutine does not require a lock. 
  
  Remove the synchronous lock entirely and use `c.reader` directly inside the spawned goroutine:
  ```go
  // No synchronous locking required here!

  go func() {
      // The goroutine waits for the previous drain to finish, 
      // allowing the caller to immediately enter the select block.
      c.readerMu.Lock() 
      defer c.readerMu.Unlock()

      if c.needsReset.Swap(false) {
          c.reader.Reset(c.Stdin)
      }
      
      // ... use c.reader directly ...
  }()
  
  select {
      // Caller can now respect context cancellation immediately!
  }
  ```
  *(Note: Update step 2 in `ADR-035` to reflect the removal of this lock).*

**3. Consistency**
- **[TECHNICAL DEBT] Mock Verification:** 
  In `internal/cli/chat_command_test.go`, the 4th argument (`capturer`) was dropped from `m.Called(ctx, cfg, cmd)` inside the `mockChatService`. While this suppresses `testify/mock` comparison errors, it skips interface contract verification for dependency injection. If `capturer` is hard to assert, it is more idiomatic to retain all 4 arguments in `m.Called(...)` and use `mock.Anything` in your test setups.

**4. Feedback Summary**
The migration to a per-request goroutine and "natural drain" pattern is an elegant, robust solution to the worker deadlock problem. It cleanly isolates the state using atomics and eliminates complex channel lifecycles. Once the synchronous `readerMu` lock in `sendRequest` is removed, the non-blocking context cancellation guarantees will be fully mathematically sound. Please adjust the code and the ADR to unblock this PR.

gosharplite:
**Architectural Review response:**

### Point 1 — Unnecessary `readerMu.Lock()` at top of `sendRequest` ✅ ACCEPTED

Agreed — `c.reader` is set once in `NewCapturer` and never reassigned. The pointer read doesn't need a lock. Removed in a3f312c.

The caller now reaches the `select` without ever blocking, so context cancellation is honored immediately even while a previous read's goroutine is still draining. ADR-035 pseudo-code updated accordingly.

### Point 2 — Re-add capturer to `mock.Called()` ❌ REJECTED

Using `mock.Anything` does **not** prevent testify from storing and reflecting on the argument — it only skips value comparison. Testify still reads the stored arguments via `reflect.typedmemmove` during `Arguments.Diff()` for error formatting, which races with `readerMu.Unlock()`'s `atomic.AddInt32`.

This is a real data race on a single object (the capturer struct), not false sharing. The race detector confirms this — removing the 4th argument eliminates it.

The mock still validates capturer behavior: `m.chatCalled` and `m.lastParams` confirm the call was made correctly. Passing a live mutable object with internal concurrency to a reflection-heavy mock framework is architecturally unsound.

gosharplite:
Thank you for the detailed response. 

1. **Synchronous Blocking Fix**: Removing the `readerMu` lock from the caller's thread perfectly resolves the scalability blocker. The caller can now immediately evaluate the `select` statement, honoring the true non-blocking intent of the per-request architecture.
2. **Testify Data Race (Rejection Accepted)**: Your explanation regarding `testify/mock`'s internal reflection (`reflect.typedmemmove`) causing a legitimate data race against the mutex state is spot on. Passing stateful, highly concurrent objects into reflection-heavy test frameworks is indeed an anti-pattern that leads to flaky race-detector failures. Omitting the parameter to bypass the reflection trap is a pragmatic and architecturally sound decision.

The architectural concerns are fully resolved. This PR looks excellent. Approved from an architectural standpoint!

gosharplite:
# Architectural Review — Final (Post-`a3f312c`)

Re-verified locally after the synchronous-lock fix landed.

## SOP Verification

| Gate | Result |
|---|---|
| `go test -race ./internal/ui/... ./internal/cli/...` | ✅ PASS |
| `golangci-lint` | ✅ 0 issues |
| ADR present (ADR-035) | ✅ |
| Regression test for #385 | ✅ `TestCapturer_ReadAfterCancellation` |
| Public API surface | ✅ Unchanged |

## 1. Analyze
- **Component:** `internal/ui/capture.go` — adapter implementing `domain_security.UserInteractor` (Hexagonal port boundary preserved).
- **Concurrency primitives reduced:** 1 mutex + 1 unbuffered chan + 1 done chan + 1 persistent goroutine → **1 mutex + 2 atomics + 0 persistent goroutines.** Net complexity decrease.

## 2. Modularity — ✅ ALIGNED
Separation of concerns intact: transport (`Stdin`/`Stdout`) vs. orchestration (`sendRequest`) vs. domain port (`UserInteractor`). DI through `NewCapturer` preserved. `closed` and `needsReset` are private state.

## 3. Scalability — ✅ RESOLVED
The synchronous-lock blocker raised previously is **resolved** in `a3f312c`. The caller now reaches `select { readDone | ctx.Done }` immediately. Cancellation is honored without waiting on a draining predecessor.

The "natural drain" pattern yields:
- **At most 2 in-flight goroutines** per capturer (1 active + 1 draining).
- **Self-healing:** the next read blocks on `readerMu` only — never on a dead channel.

## 4. Consistency — ✅ ALIGNED
ADR-035 documents the trade-off matrix (Options A/B/C) — exemplary. Atomic-based `Close()` idempotency follows the established shutdown pattern. Test naming clearly signals regression intent.

## 5. Findings (non-blocking polish)

### [REFACTOR] Document the buffered channel intent
The `readDone chan ioResult` has capacity 1 specifically so the drained goroutine can terminate after the caller returned with `ctx.Err()`. Looks correct now, but a future "optimization" could remove the buffer and reintroduce a leak.

```go
// readDone is buffered (cap 1) so the read goroutine can publish its
// result and exit cleanly even after the caller's context cancelled
// and stopped receiving.
readDone := make(chan ioResult, 1)
```

### [REFACTOR] Test timing fragility
`TestCapturer_ReadAfterCancellation` uses two `time.Sleep` calls to coordinate the drain. PR notes "5/5 consistent" but slow CI runners may surface flakiness. Consider polling instead:

```go
require.Eventually(t, func() bool { return !c.needsReset.Load() }, time.Second, 5*time.Millisecond)
```

### [TECHNICAL DEBT — already acknowledged]
Dropping the `capturer` argument from `m.Called(...)` loses DI contract verification at the mock layer. Author's testify-reflection-race justification is correct and pragmatic. A cleaner long-term path is to inject an interface stub instead of the live `*capturer`, but that is out of scope for this PR.

## Decision

**✅ APPROVED — Architecturally Sound**

The refactor is a clear win: lower primitive count, deterministic close semantics, self-healing under cancellation, full regression coverage, ADR-documented. The two remarks above are non-blocking polish for a follow-up.

Excellent rigor in ADR-035's Options-Considered section and the disciplined response to prior review feedback.

PR #393, opened 2026-05-14
fix(coverage): propagate test errors in getDetailedCoverage + add regression test (#392)

thptcnec:
## Summary

Closes #392.

Two hardening fixes extracted from the #391 investigation:

### 1. Stop silently discarding test errors

**Before:** `getDetailedCoverage` discarded `runCoverageTest` errors with `_ =`, silently parsing partial or empty coverage profiles.

**After:** Errors are propagated with context so callers know the profile may be incomplete.

### 2. Add regression test for parser count accuracy

New `TestParseCoverageProfile_MatchesRawCount` validates that `parseCoverageProfile` counts exactly the same number of uncovered blocks as raw `awk` filtering on a golden profile. Guards against future parsing changes silently corrupting counts.

### Verification

| Check | Result |
|-------|--------|
| `go build ./internal/tools/analysis/...` | ✅ |
| `go test ./internal/tools/analysis/...` | ✅ PASS (7.8s) |
| `go test -short ./...` | ✅ All passing |

### DIFF
--- a/internal/tools/analysis/coverage_parser.go
+++ b/internal/tools/analysis/coverage_parser.go
@@ -378,8 +378,9 @@ func (m *healthManager) getDetailedCoverage(ctx context.Context, packagePath str
 	}()
 	_ = f.Close()
 
-	// Execute coverage test (ignores error as original code did, allowing partial profiles)
-	_ = m.runCoverageTest(ctx, packagePath, tempPath, hb)
+	if err := m.runCoverageTest(ctx, packagePath, tempPath, hb); err != nil {
+		return nil, fmt.Errorf("coverage test execution failed (profile may be incomplete): %w", err)
+	}
 
 	if err := validateProfile(tempPath); err != nil {
 		return nil, err
--- a/internal/tools/analysis/coverage_parser.go
+++ b/internal/tools/analysis/coverage_parser.go
@@ -378,11 +378,12 @@ func (m *healthManager) getDetailedCoverage(ctx context.Context, packagePath str
 	}()
 	_ = f.Close()
 
-	if err := m.runCoverageTest(ctx, packagePath, tempPath, hb); err != nil {
-		return nil, fmt.Errorf("coverage test execution failed (profile may be incomplete): %w", err)
-	}
+	testErr := m.runCoverageTest(ctx, packagePath, tempPath, hb)
 
 	if err := validateProfile(tempPath); err != nil {
+		if testErr != nil {
+			return nil, fmt.Errorf("test execution failed and no valid profile was generated: %w", testErr)
+		}
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


thptcnec:
### Architectural Review

**[ARCHITECTURAL BLOCKER]: Breaking TDD Workflow (RED-GREEN-REFACTOR)**

By strictly returning `err` on `m.runCoverageTest` failure, this change introduces a regression that blinds developers and AI agents to coverage metrics when tests fail. 

When `go test` executes and encounters a failing test, it exits with code `1` (which bubbles up as an error). However, it **still writes a valid coverage profile** for the code that executed. In a standard TDD workflow, we write failing tests first and rely on tools like `get_detailed_coverage` to verify what paths those failing tests hit. This PR drops those successfully generated coverage blocks by fast-failing.

**Proposed Scalable Design**:
Capture the test execution error, but continue attempting to validate and parse the profile. Return the partial blocks *along with* the error, and update the caller methods to format the report with a warning prefix instead of discarding the result.

```go
// 1. In getDetailedCoverage:
testErr := m.runCoverageTest(ctx, packagePath, tempPath, hb)

if err := validateProfile(tempPath); err != nil {
    if testErr != nil {
        return nil, fmt.Errorf("test execution failed and no valid profile was generated: %w", testErr)
    }
    return nil, err
}

blocks, err := parseDetailedCoverage(ctx, cf, m.Runner, os.ReadFile)
if err != nil {
    return nil, err
}

if testErr != nil {
    return blocks, fmt.Errorf("coverage test failed (profile may be incomplete): %w", testErr)
}
return blocks, nil
```

**[TECHNICAL DEBT]: Caller Handling for Partial Success**
Because `getDetailedCoverage` will now correctly return `(blocks, testErr)`, update the callers (`getDetailedCoverageReport` and `getDetailedCoverageJSON`) so they don't discard the parsed blocks when an error is present. For example, prepend a `⚠️ WARNING:` to the text report without failing the tool execution.

**[REFACTOR]: Excellent Testing**
The new regression test (`TestParseCoverageProfile_MatchesRawCount`) is exceptionally well-structured and aligns perfectly with our Go testing SOP.

Please refactor the error handling to support partial profile parsing and push the updates.

thptcnec:
### Addressed all three review points

**1. [ARCHITECTURAL BLOCKER] — Return partial blocks + error**

`getDetailedCoverage` now captures `testErr` without immediately returning, continues through `validateProfile` → `parseDetailedCoverage`, then returns `(blocks, testErr)` when the profile exists but tests failed. The TDD RED phase is preserved — failing tests still surface coverage data.

**2. [TECHNICAL DEBT] — Callers handle partial results**

| Caller | Behavior when `testErr != nil` |
|--------|-------------------------------|
| `getDetailedCoverageReport` | Prepends `⚠️ WARNING: test execution failed, profile may be incomplete` to the report; returns `(report, nil)` so the warning renders in tool output |
| `getDetailedCoverageJSON` | Returns `(jsonStr, error)` — callers can wrap both in `ToolResult` |

Both use a `len(blocks) == 0` guard: if no blocks were parsed and tests failed, the error propagates normally. If blocks exist, the data is preserved.

**3. [REFACTOR] — Regression test**

`TestParseCoverageProfile_MatchesRawCount` was already committed (Task 2) and remains unchanged — golden profile with 3 uncovered / 2 covered, asserting exact count and field values.

thptcnec:
### Approval: Architecturally Sound

The updated implementation perfectly addresses all structural concerns. 

1. **TDD Workflow Preserved:** Capturing `testErr` while allowing the logic to gracefully fall through `validateProfile` and `parseDetailedCoverage` ensures that failing tests still surface the crucial coverage map.
2. **Caller Resiliency:** Wrapping the partial success gracefully in both the Text and JSON report callers cleanly removes the technical debt. The `⚠️ WARNING` injection correctly alerts the user without discarding actionable tool output.
3. **Regression Test Maintained:** The `TestParseCoverageProfile_MatchesRawCount` test correctly asserts against the golden profile properties, adhering to Go testing SOPs.

Excellent work resolving the blockers. The PR is approved from an architecture standpoint.

PR #413, opened 2026-05-15
fix(telemetry): harden error recovery paths (93.4% → 96.5%) (#408)

### LINKED ISSUE #408: Error Path Hardening: infrastructure/telemetry/ — Telemetry Error Recovery Paths (93.4% → ≥96%)
## Summary

Close untested error handling paths in the telemetry infrastructure (`infrastructure/telemetry/`). Currently at **93.4%** — recently improved from 82.4% (#398), but ~20 functions still have uncovered error recovery paths in ledger persistence, metrics registration, and cost estimation.

## Parent Epic

#377 (v4) — Error Path Hardening

## Scope

| File | Key Gaps | Coverage |
|------|----------|----------|
| `ledger.go` | `recoverLedger`, `discoverNewRecords`, `persistMergedLedger`, `mergeRecords`, `getSessionID` | 81.8–93.3% |
| `metrics.go` | `RegisterMetrics`, `RecordSessionCost`, `logTrace` | 72.2–93.3% |
| `metrics_cost.go` | `updateLedgerHistory`, `EstimateCost`, `renderReport`, `applyRetentionPolicy` | 76.9–92.9% |
| `metrics_summary.go` | `resolveGroupingKey`, `ensureLedgerReady`, `getCostSummary` | 78.6–93.8% |
| `metrics_tracker.go` | `AccumulateAndReturn`, `GetDailyCost` | 89.5–95.2% |
| `metrics_util.go` | `processLogLine` | 92.9% |
| `system_metrics_linux.go` | `getRoot` | 66.7% |
| `system_metrics_shared.go` | `getRuntimeCPUStats` | 85.7% |

## Approach

1. **Priority**: `getRoot` (66.7%), `RegisterMetrics` (72.2%), `RecordSessionCost` (75.0%), `EstimateCost` (76.9%) — these are the largest gaps
2. Use temp directories and ledger files to simulate corruption/recovery scenarios
3. Table-driven tests for cost estimation with edge-case pricing data
4. Mock event bus for metrics registration failure paths

## Acceptance Criteria

- [ ] Package coverage >= 96%
- [ ] `RegisterMetrics` >= 85%
- [ ] `RecordSessionCost` >= 85%
- [ ] `EstimateCost` >= 85%
- [ ] `getRoot` >= 80% or documented as platform-specific (Linux-only)
- [ ] Tests pass with `-race` and `-count=10`
- [ ] No new linter warnings

## Timebox

~2 hours

gosharplite:
Closes #408.

## Summary
Closed ~20 untested error handling paths in `infrastructure/telemetry/` across 8 test files (6 modified, 2 new — 1,132 insertions). Coverage improved from **93.4% → 96.5%**.

### Key gaps closed
| Function | Before | After |
|----------|--------|-------|
| `persistMergedLedger` | 81.8% | **100.0%** |
| `RegisterMetrics` | 72.2% | **88.9%** |
| `EstimateCost` | 76.9% | **92.3%** |
| `RecordSessionCost` | 75.0% | **87.5%** |
| `resolveGroupingKey` | 78.6% | **100.0%** |
| `renderReport` | 88.9% | **100.0%** |
| `AccumulateAndReturn` | 89.5% | **100.0%** |
| `processLogLine` | 92.9% | **100.0%** |
| `mergeRecords` | 88.9% | **100.0%** |
| `updateLedgerHistory` | 77.8% | marshal error documented unreachable |
| `getRoot` | 66.7% | **100.0%** |
| **Package** | **93.4%** | **96.5%** |

## Verification
| Check | Status |
|-------|--------|
| make check-full (fmt/tidy/build/lint/arch/vuln/test/dead-code) | PASS |
| telemetry `-race` | PASS (1.75s) |
| Package coverage | 96.5% (≥96%) |
| Linter | 0 issues |

### DIFF
--- a/internal/infrastructure/telemetry/infra_telemetry_extended_test.go
+++ b/internal/infrastructure/telemetry/infra_telemetry_extended_test.go
@@ -6,6 +6,7 @@ package telemetry
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"os"
 	"path/filepath"
 	"strings"
@@ -72,6 +73,24 @@ func (m *mockRegistry) RegisterWithOptions(def *tools.ToolDeclaration, handler t
 	}
 	m.handlers[def.Name] = handler
 	return nil
+}
+
+// failingRegistry extends mockRegistry with optional error injection fields
+// for testing RegisterMetrics error paths.
+type failingRegistry struct {
+	mockRegistry
+	registerErr error  // if set, RegisterWithOptions returns this error for ALL registrations
+	failOn      string // if set, RegisterWithOptions returns error only when def.Name == failOn
+}
+
+func (f *failingRegistry) RegisterWithOptions(def *tools.ToolDeclaration, handler tools.ToolFunc, opts tools.ToolOptions) error {
+	if f.registerErr != nil {
+		return f.registerErr
+	}
+	if f.failOn != "" && def.Name == f.failOn {
+		return errors.New("injected failure for " + def.Name)
+	}
+	return f.mockRegistry.RegisterWithOptions(def, handler, opts)
 }
 
 func TestSessionCostTracker_Extended(t *testing.T) {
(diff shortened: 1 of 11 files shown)


gosharplite (commented):
## 🔍 Architectural Review

**Summary:** Solid test coverage push (93.4% → 96.5%) closing ~20 untested error paths. Test structure is idiomatic, table-driven, and well-organized. No production code changes — purely test augmentation. Approval recommended with three items to address.

---

### ⚠️ [ARCHITECTURAL BLOCKER] Silent data loss on lock acquisition failure

Both `recordCost` and `persistMergedLedger` silently return when `acquireLedgerLock` fails with a fresh (non-stale) lock — no log, no error propagation, no retry. The new tests correctly lock in the current behavior by asserting files are NOT modified (excellent documentation), but the production code has **zero observability** on this data-loss path.

**Request:** Add at minimum a `log.Printf("Warning: ...")` before each bare `return`, or file an ADR documenting why silent discard is acceptable.

| Call site | File | Line |
|---|---|---|
| `recordCost` | `metrics_cost.go` | 34-37 |
| `persistMergedLedger` | `ledger.go` | 229-232 |

---

### 📝 [TECHNICAL DEBT] Duplicate `acquireLedgerLock` implementations

`metricsManager.acquireLedgerLock` (metrics_cost.go:47) and `ledgerStore.acquireLedgerLock` (ledger.go:245) are near-identical. The new tests now exercise both independently, making the duplication more visible. Consider extracting to a single package-level function or delegating `metricsManager` → `ledgerStore` via the existing `ledger` field.

---

### 📝 [TECHNICAL DEBT] `failingRegistry` value-embedding fragility

```go
type failingRegistry struct {
    mockRegistry           // VALUE embedding — copies struct + mutex
    registerErr error
    failOn      string
}
```

If `mockRegistry` ever gains a pointer-receiver method, the embedded copy won’t reflect mutations. Recommend changing to `*mockRegistry`.

---

### ✅ Positives

- Table-driven tests with clean `t.Run`/`t.Parallel` usage
- Non-parallel `chmod` tests correctly use `t.Skip` (Windows) + `t.Cleanup`
- Unreachable-branch documentation tests are well-justified and thorough
- New mock types (`failingRegistry`, `mockSMWithError`) follow composition-over-inheritance
- `testify` usage is consistent with existing package practice

---

### 🔴 CI Check

`gh pr checks` reports **no checks registered** on this branch. The PR body claims `make check-full` PASS locally — please verify CI is configured to trigger on `fix/*` branch patterns.

gosharplite:
All three architectural review items from
https://github.com/gosharplite/tell-me-go/pull/413#issuecomment-*
have been addressed:

1. **Lock-contention warnings** — `log.Printf` added before silent
   returns in `recordCost` (metrics_cost.go:36) and `persistMergedLedger`
   (ledger.go:231). Tests continue to verify the file is not modified;
   we now have observability on the contention path.

2. **Deduplicate `acquireLedgerLock`** — extracted from two receiver
   methods into a single package-level function at ledger.go:66.
   12 call sites updated (production + tests). Zero dead code.

3. **`failingRegistry` pointer embedding** — changed from value
   embedding (`mockRegistry`) to pointer embedding (`*mockRegistry`).
   Both constructor sites updated with explicit initialization.

Verification: build ✅ · 51 tests ✅ · lint ✅ · race ✅

gosharplite:
## ✅ Architectural Sign-Off

All three findings from the [architectural review](https://github.com/gosharplite/tell-me-go/pull/413#issuecomment-*) have been addressed in commit `32d3c9a0`:

| # | Finding | Resolution |
|---|---------|-------------|
| 1 | **[ARCHITECTURAL BLOCKER]** Silent data loss on lock contention | `log.Printf` warnings added — observability restored, tests unchanged |
| 2 | **[TECHNICAL DEBT]** Duplicate `acquireLedgerLock` | Single package-level function, 12 call sites unified |
| 3 | **[TECHNICAL DEBT]** `failingRegistry` value embedding | `*mockRegistry` pointer embedding, both constructors updated |

**No remaining architectural concerns. Approve.**

gosharplite:
Here is the architectural review for PR 413 based on the provided diff. 

### 1. Analysis & Component Mapping
The PR primarily refactors the telemetry and ledger persistence mechanisms by extracting a duplicated lock acquisition routine into a package-level `acquireLedgerLock` function. It also heavily improves test coverage for previously unreachable or error-handling paths, specifically explicitly testing "silent data-loss gaps" when locking fails.

### 2. Architectural Findings

#### [ARCHITECTURAL BLOCKER] - Data Loss Under Concurrency (Scalability & Data Integrity)
**Finding:** The ledger lock relies on `os.O_CREATE | os.O_EXCL` for cross-process synchronization. If the lock exists and isn't stale, `acquireLedgerLock` immediately returns an `os.IsExist` error. The callers (`recordCost` and `persistMergedLedger`) log a warning and silently `return` without persisting the telemetry/cost data. 
**Impact:** Because this is a shared file system ledger, concurrent CLI invocations or sub-processes attempting to log session costs simultaneously will experience lock contention. One process will succeed; the others will permanently drop their financial tracking records. In a system managing AI API budgets, dropping cost records defeats the purpose of a ledger.
**Current Design:**
```go
// Immediate failure on contention
func acquireLedgerLock(lockPath string) (*os.File, error) {
    f, err := os.OpenFile(lockPath, os.O_CREATE|os.O_EXCL, 0644)
    // Checks for staleness but immediately returns error if actively held
    return f, err
}

// In Caller (metrics_cost.go & ledger.go):
f, err := acquireLedgerLock(lockPath)
if err != nil {
    log.Printf("Warning: Failed to acquire ledger lock... %v", err)
    return // <--- SILENT DATA LOSS
}
```
**Scalable Design (Retry with Backoff):**
Locks must support retry/spin-wait to gracefully handle momentary contention instead of fast-failing. Add a `context.Context` to handle timeouts gracefully.
```go
func acquireLedgerLock(ctx context.Context, lockPath string) (*os.File, error) {
    const maxWait = 2 * time.Second
    const retryInterval = 50 * time.Millisecond
    
    timeout := time.After(maxWait)
    ticker := time.NewTicker(retryInterval)
    defer ticker.Stop()

    for {
        f, err := os.OpenFile(lockPath, os.O_CREATE|os.O_EXCL, 0644)
        if err == nil {
            return f, nil
        }
        
        if os.IsExist(err) {
            if isStale(lockPath) {
                _ = os.Remove(lockPath) // attempt to clear stale lock
                continue // immediate retry
            }
        } else {
            return nil, err // fail fast on actual IO/permission errors
        }

        // Wait for next tick or timeout
        select {
        case <-ctx.Done():
            return nil, ctx.Err()
        case <-timeout:
            return nil, fmt.Errorf("timeout acquiring lock %s after %v", lockPath, maxWait)
        case <-ticker.C:
            // spin loop
        }
    }
}
```

#### [TECHNICAL DEBT] - Swallowing Errors (Consistency)
**Finding:** Both `recordCost` and `persistMergedLedger` swallow lock and file system errors. They use `log.Printf` to emit a warning to `stdout/stderr` (which pollutes CLI output) and return normally as if the operation succeeded.
**Impact:** Callers higher up in the stack remain entirely unaware that persistence failed. This violates standard Go idiom where errors are values that should be handled or explicitly propagated.
**Recommendation:** Refactor the method signatures (e.g., `persistMergedLedger`, `recordCost`) to return an `error`. If it truly is a background task where failure is acceptable, pass it to an established telemetry error-handling sink rather than raw `log.Printf`.

#### [REFACTOR] - Tight I/O Coupling in Domain Logic
**Finding:** `acquireLedgerLock` manipulates the file system directly using the standard `os` library without leveraging the existing `domain_security.Manager` to check if the lock path is safe or writable.
**Impact:** While telemetry usually writes to safe local directories, bypassing the application's abstraction layers causes you to need extensive local file system manipulation in tests (e.g., using `chmod 0000` to mock permission errors).
**Recommendation:** Although acceptable in the short term, consider injecting a file system/storage interface (`LedgerStorage`) into the metrics components. This enforces Separation of Concerns and prevents tests from requiring actual disk I/O and Unix-specific `chmod` hacks to trigger error conditions.

gosharplite:
## Response to the follow-up architectural comments

Three additional points were raised. Here is my assessment as Architect:

---

### 1. Retry-with-backoff for lock contention

**Partially agree — diagnosis correct, proposed fix has a latency problem.**

Concurrent processes will indeed drop records under lock contention. But `recordCost` is called synchronously inside `EstimateCost` — a user-facing tool handler. A 2-second spin-loop on a file lock would cause visible UI hangs when the user just asked "what did this cost?"

**Recommendation:** A short retry (3 attempts × 10ms = 30ms total) is worth doing. For anything beyond that, a write-ahead side-file + background merge pattern would be better. This is a real tradeoff between data integrity and latency — warrants an **ADR** — but is **not a blocker for this PR**, which is focused on test coverage. The warning log is already a strict improvement over the prior silent discard.

**Verdict:** ✅ Agree with the problem. ⚠️ 2s blocking retry in a sync handler is the wrong fix.

---

### 2. "Callers should return errors instead of logging"

**Disagree — would be a worse design.**

`recordCost` and `persistMergedLedger` are fire-and-forget side effects. They are called from:

- `EstimateCost` → returns `(costReport, error)` where `error` means "I could not calculate the cost." Conflating computation errors with persistence errors would tell the user "I have no idea what your cost is" when the truth is "your cost is $1.50 but I could not save it."
- `recoverLedger` → a background goroutine with no caller to propagate to.

Go idiom for fire-and-forget side effects **is** `log.Printf` (or `slog.Warn`). The architectural concern was zero observability — adding the log fixed that. Returning errors would force a signature change that conflates concerns.

**Verdict:** ❌ Disagree. `log.Printf` is the correct pattern at this layer.

---

### 3. Inject a filesystem interface

**Agree in principle, but overstated for this PR.**

The package already abstracts persistence through `persistence.AtomicWrite`. The `chmod`-based tests exercise **real OS error conditions** (EACCES, EISDIR) that a mock filesystem cannot surface — without them, those branches remain untested. A `LedgerStorage` interface would require the mock to simulate permission errors to achieve the same coverage, which is circular.

**Verdict:** ⚠️ Valid long-term idea. Not a blocker for this PR.

---

### Bottom line

| # | Point | Verdict |
|---|-------|---------|
| 1 | Retry-with-backoff | Real problem, needs ADR. Wrong fix proposed. Not a blocker. |
| 2 | Return errors | Disagree — fire-and-forget side effects log, not propagate. |
| 3 | Filesystem interface | Good long-term idea, but chmod tests cover real OS paths a mock cannot. |

**None of these should block merging.** The two architectural blockers from the original review are already resolved in commit `32d3c9a0`.

gosharplite:
I have reviewed the response and fully agree with your assessment and pushback. Your reasoning highlights mature architectural trade-offs between data integrity, system latency, and resilience.

### Breakdown of Agreement:

1. **Lock Contention & Latency (Retry-with-Backoff)**
   You are absolutely correct. My proposed 2-second spin-loop failed to account for the execution context (a synchronous, user-facing handler). Blocking the critical path for a telemetry side-effect would introduce unacceptable tail latency. A bounded micro-retry (e.g., 30ms) or a Write-Ahead Log (WAL) with a background merge pattern is the correct, scalable solution. Deferring this to an ADR is the right call.

2. **Swallowing Errors vs. Fire-and-Forget**
   I concede this point. Failing a primary domain operation (estimating the cost) because a secondary telemetry side-effect failed violates system resilience. Propagating the error up the stack would conflate concerns and give the user a false negative. Emitting the failure to an observability sink via `log.Printf`/`slog.Warn` without disrupting the main thread is indeed the idiomatic pattern for "best-effort" operations.

3. **Tight I/O Coupling (Filesystem Interface)**
   I agree. While Dependency Injection is generally preferred, applying it to a leaf telemetry package solely to make tests "pure" can lead to over-engineering. Mocking filesystem errors (like `EACCES` or `EISDIR`) often results in tautological testing. Using robust `chmod` integration tests provides a higher degree of confidence in actual POSIX compliance and file-system locking behavior.

**Conclusion:**
You have accurately balanced ideal architectural patterns against pragmatic system constraints (latency, primary vs. secondary workflows, and test fidelity). 

**The architectural review is complete. I approve the PR.**

PR #420, opened 2026-05-15
fix(registry,security): close error handling gaps, achieve 100% coverage (#419)

gosharplite:
Closes #419.

## Summary
Closed all 8 coverage gaps across two leaf packages:
- **domain/security/**: Added `IsSafeGhSubcommand` and `IsSafeAzSubcommand` tests with policy-iterate guards (93.8% → 100.0%)
- **infrastructure/registry/**: Added `GetOptions`, empty-toolkit defaults, dedup, and cross-toolkit re-registration tests (93.0% → 100.0%)

## Changes
| File | What |
|------|------|
| `internal/domain/security/service_test.go` | +2 test functions (12 subtests) |
| `internal/infrastructure/registry/registry_test.go` | +3 test functions / subtests |
| `internal/infrastructure/registry/coverage_test.go` | +1 table-driven test case |

## Verification
| Check | Status |
|-------|--------|
| make check-full | PASS |
| domain/security coverage | 100.0% |
| infrastructure/registry coverage | 100.0% |
| Race detector | Clean |
| Lint | 0 issues |

### DIFF
--- a/internal/domain/security/service_test.go
+++ b/internal/domain/security/service_test.go
@@ -172,3 +172,71 @@ func TestSafetyService_IsSafeGoSubcommand(t *testing.T) {
 		})
 	}
 }
+
+func TestSafetyService_IsSafeGhSubcommand(t *testing.T) {
+	t.Parallel()
+	s := NewSafetyService(DefaultPolicy())
+	tests := []struct {
+		name string
+		sub  string
+		want bool
+	}{
+		{"auth", "auth", true},
+		{"pr", "pr", true},
+		{"push", "push", false},
+		{"unknown-subcommand", "unknown-subcommand", false},
+		{"empty", "", false},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			if got := s.IsSafeGhSubcommand(tt.sub); got != tt.want {
+				t.Errorf("IsSafeGhSubcommand(%q) = %v, want %v", tt.sub, got, tt.want)
+			}
+		})
+	}
+
+	t.Run("all_default_safe_subcommands", func(t *testing.T) {
+		t.Parallel()
+		for k := range DefaultPolicy().SafeGhSubcommands {
+			if !s.IsSafeGhSubcommand(k) {
+				t.Errorf("IsSafeGhSubcommand(%q) = false, want true", k)
+			}
+		}
+	})
+}
+
+func TestSafetyService_IsSafeAzSubcommand(t *testing.T) {
+	t.Parallel()
+	s := NewSafetyService(DefaultPolicy())
+	tests := []struct {
+		name string
+		sub  string
+		want bool
+	}{
+		{"devops", "devops", true},
+		{"account", "account", true},
+		{"vm", "vm", false},
+		{"unknown-subcommand", "unknown-subcommand", false},
+		{"empty", "", false},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			if got := s.IsSafeAzSubcommand(tt.sub); got != tt.want {
+				t.Errorf("IsSafeAzSubcommand(%q) = %v, want %v", tt.sub, got, tt.want)
+			}
+		})
+	}
+
+	t.Run("all_default_safe_subcommands", func(t *testing.T) {
+		t.Parallel()
+		for k := range DefaultPolicy().SafeAzSubcommands {
+			if !s.IsSafeAzSubcommand(k) {
+				t.Errorf("IsSafeAzSubcommand(%q) = false, want true", k)
+			}
+		}
+	})
+}
--- a/internal/infrastructure/registry/coverage_test.go
+++ b/internal/infrastructure/registry/coverage_test.go
@@ -56,6 +56,75 @@ func TestRegistry_DuplicateRegistration(t *testing.T) {
 				}
 			},
 		},
+		{
+			name: "re-register to different toolkit",
+			actions: func(t *testing.T, r tools.Registry) {
+				def1 := &tools.ToolDeclaration{Name: "cross_tool", Description: "desc_v1"}
+				h1 := func(ctx context.Context, args map[string]interface{}, hb chan<- struct{}) (tools.ToolResult, error) {
+					return tools.ToolResult{Text: "v1"}, nil
+				}
+				if err := r.RegisterToToolkit("core", def1, h1); err != nil {
+					t.Fatalf("failed to register tool to core: %v", err)
+				}
+
+				def2 := &tools.ToolDeclaration{Name: "cross_tool", Description: "desc_v2"}
+				h2 := func(ctx context.Context, args map[string]interface{}, hb chan<- struct{}) (tools.ToolResult, error) {
+					return tools.ToolResult{Text: "v2"}, nil
+				}
+				if err := r.RegisterToToolkitWithOptions("git", def2, h2, registry.ToolOptions{LongRunning: true}); err != nil {
+					t.Fatalf("failed to re-register tool to git: %v", err)
+				}
+			},
+			validate: func(t *testing.T, r tools.Registry) {
+				// 1. cross_tool remains in core declarations
+				coreDecls := r.GetCoreDeclarations()
+				foundInCore := false
+				for _, d := range coreDecls {
+					if d.Name == "cross_tool" {
+						foundInCore = true
+						break
+					}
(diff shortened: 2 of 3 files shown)


thptcnec (approved):
## Architectural Review — APPROVE ✅

**Scope**: +299/−0, tests only, no production changes. Clean.

### What's Good
- **Policy-iterate guards** in `IsSafeGhSubcommand`/`IsSafeAzSubcommand` catch future policy drift.
- **Zero-value coverage** for `GetOptions("")` and nonexistent tools.
- **Empty-toolkit → core** path validated with 3 sub-cases.
- **Cross-toolkit re-registration** table entry verifies `updateExistingRegistration` + `updateToolkitMap`.
- All tests use `t.Parallel()` consistently.

### Suggestions (non-blocking, follow-up PR)

1. **Inconsistent iterate-all pattern**: `IsSafeGitSubcommand` and `IsSafeGoSubcommand` lack the `for k := range DefaultPolicy().Safe*Subcommands` loop that the new gh/az tests include. Adding it would prevent false confidence when new safe subcommands are added to the policy.

2. **Brittle hardcoded counts** in the new `GetDeclarationsByToolkits - core explicitly requested` subtest: `assert.Len(t, decls, 1)` and `assert.Len(t, decls, 3)` silently break if someone adds another tool registration to the parent `TestRegistry_Toolkits` setup. The dedup-frequency approach already used later in that subtest is more robust — consider using it throughout.

3. **`GetOptions` not verified after re-registration**: the cross-toolkit test in `coverage_test.go` calls `IsLongRunning` but not `GetOptions` — a minor gap.

No architectural blockers. Merge-ready.

thptcnec:
### Architectural Review of PR 420

Based on the provided modifications, the PR introduces test coverage for the tool `Registry` and subcommand `SafetyService`. While the TDD practices applied here are excellent, the test cases expose a critical flaw in the underlying `Registry` architecture.

#### 1. Analyze (Component Dependencies & Boundaries)
The `Registry` acts as a central authority mapping tool names to handlers across multiple toolkits. The newly added test `re-register to different toolkit` in `coverage_test.go` explicitly asserts that if a tool named `cross_tool` is registered in the `core` toolkit and later registered in the `git` toolkit, the registry overwrites the original handler but retains the tool in **both** toolkit lists.

#### 2. Modularity (Separation of Concerns)
This design violates toolkit isolation. Toolkits (e.g., `git`, `k8s`, `core`) should be modular and side-effect-free. Because the global `r.entries` map is keyed purely by the tool's `Name` string (without a toolkit prefix), an inadvertent name collision between two different toolkits will silently corrupt the global state. The last registered toolkit will hijack the implementation for all other toolkits using that name.

#### 3. Scalability (State Management)
As the system scales and more toolkits are introduced, the risk of naming collisions (e.g., a generic `deploy` or `get_logs` tool) increases. Silent overwriting is an unscalable state management approach that will lead to severe operational anomalies when the LLM orchestrator loads mixed toolkits.

#### 4. Consistency (Established Patterns)
A robust `Registry` pattern must protect its invariants. It should either guarantee uniqueness by preventing collisions or isolate namespaces. Enshrining the overwrite behavior as "expected" in a test creates technical debt.

---

### Findings

#### [ARCHITECTURAL BLOCKER]: Global Tool Namespace Collisions
**Current Design:**
The registry's `updateExistingRegistration` overwrites the global `toolEntry` when a duplicate name is registered, even if the registration originates from a completely different toolkit. The test suite now strictly enforces this buggy behavior.
```go
// Current (from coverage_test.go): 
// Proving that registering "cross_tool" to "git" overwrites the "core" handler
def2 := &tools.ToolDeclaration{Name: "cross_tool", Description: "desc_v2"}
r.RegisterToToolkitWithOptions("git", def2, h2, ...)

// Test verifies the core toolkit now uses the git toolkit's handler
```

**Scalable Design (Recommended Refactor):**
The `Registry` should explicitly reject cross-toolkit duplication and return an error (`ErrToolNameConflict`) to force developers to use unique names (e.g., `git_commit`, `k8s_apply`), OR the registry should implement namespace prefixing automatically. 

Update `registry.go` and the corresponding tests to enforce uniqueness:
```go
func (r *registry) RegisterToToolkitWithOptions(toolkit string, def *tools.ToolDeclaration, handler toolFunc, opts ToolOptions) error {
    // ...
    r.mu.Lock()
    defer r.mu.Unlock()

    if existing, exists := r.entries[def.Name]; exists {
        // Find which toolkit owns this tool
        owner := r.findOwnerToolkit(def.Name)
        if owner != toolkit {
            return fmt.Errorf("tool name collision: %q is already registered by toolkit %q", def.Name, owner)
        }
        // Allow updating only if it belongs to the SAME toolkit (e.g., for hot-reloading)
        r.updateExistingRegistration(toolkit, def, handler, opts)
        return nil
    }
    // ...
}
```

#### [APPROVED]: Excellent Table-Driven Testing
The additions to `internal/domain/security/service_test.go` (`IsSafeGhSubcommand` / `IsSafeAzSubcommand`) are well-written. They correctly utilize Go's table-driven patterns, `t.Parallel()`, and take advantage of the Go 1.22+ loop variable capture fix, ensuring high maintainability and isolated execution without leaking scope.

thptcnec:
I've reviewed the [ARCHITECTURAL BLOCKER] claim above. I disagree with it for three reasons:

### 1. Tool name = global identity by design

The registry's primary key is the tool name — not a composite of (toolkit, name). This is visible throughout the codebase:
- `r.entries` is a flat `map[string]toolEntry`
- `GetDeclarationsByToolkits` explicitly *deduplicates* by name across toolkits
- `Execute` dispatches by name alone, with no toolkit qualifier

If toolkits were intended as namespaces, none of these would be necessary. The toolkit is a *loading filter* — it controls which tools the orchestrator loads in a given context — not an identity boundary.

### 2. The 'collision' scenario is a developer error, not a runtime hazard

Registering two semantically different tools under the same name is equivalent to defining two Go functions with the same signature in the same package. The overwrite behavior is the registry's analogue of a redeclaration: last write wins. The test in `coverage_test.go` correctly encodes this as *intended update behavior*, not a bug to be fixed.

### 3. Rejecting cross-toolkit updates would break valid use cases

`ErrToolNameConflict` would prevent:
- Moving a tool from `core` to `git` after realizing it's git-specific
- Hot-reloading a tool's handler/options while also adding it to an additional toolkit
- The exact scenario validated by the new test case: enhancing `cross_tool` with `LongRunning: true` and adding it to the `git` toolkit

---

**Bottom line**: The current design (global namespace, toolkits as non-exclusive groupings) is a valid architectural choice. The comment raises a reasonable design discussion — whether the project *should* namespace tools under toolkits — but that's a separate Architectural Decision Record and refactor, not grounds to block a test-coverage PR.

The three TECH DEBT items from my original review still stand as low-priority follow-up improvements. No blockers remain.

thptcnec:
### Architect's Assessment & Resolution

I acknowledge the clarification regarding the design intent: the `Registry` treats tool names as global identifiers, and toolkits function as non-exclusive tags/loading filters rather than strict namespaces.

#### Evaluation of the Rebuttal
1. **Global Identity:** If toolkits are tags, the global flat map (`r.entries`) and deduplication logic are indeed consistent with that design.
2. **Update Behavior:** Using "last write wins" for hot-reloading or migrating tools is a valid use case for runtime updates.

However, from a scalability and predictability standpoint, silently overwriting handlers across unrelated toolkit initializations remains a **[TECHNICAL DEBT]** risk. While it serves hot-reloading, it completely masks genuine developer errors (e.g., a dev accidentally naming a tool `deploy` in both the `k8s` and `ado` toolkits). When both toolkits are requested, the orchestrator's behavior becomes implicitly dependent on package initialization order, which is an anti-pattern in Go.

#### Resolution
I agree that the PR should not be blocked, as the tests accurately enforce the system's current mechanics. I am lifting the [ARCHITECTURAL BLOCKER].

To ensure maintainability as the system scales, I request the following follow-ups:
1. **[ADR REQUIRED]**: Please draft an Architecture Decision Record (ADR) explicitly documenting that "Tool Names are Global Identifiers" and "Toolkits are Tag-based Loading Filters." This will prevent future confusion.
2. **[REFACTOR] (Future)**: Consider emitting a warning log when a tool handler is overwritten by a *different* toolkit, or establish a strict convention requiring namespaced prefixes (e.g., `git_*`, `k8s_*`) to prevent silent runtime collisions.

Approved to proceed.

thptcnec (approved):
The tests enforce the current architectural design correctly. The architectural blocker has been resolved and downgraded to technical debt to be addressed in the future via an ADR. Approved.

Issue #446, opened 2026-05-16
dead_code_graph: buildFileToPkgMap ignores XTestGoFiles — external test usages invisible


thptcnec:
## Root Cause

`buildFileToPkgMap` in `internal/tools/analysis/dead_code.go` only indexes `pkg.GoFiles`:

```go
func (a *defaultDeadCodeAnalyzer) buildFileToPkgMap(pkgs []*packages.Package) map[string]string {
    fileToPkg := make(map[string]string)
    for _, pkg := range pkgs {
        for _, file := range pkg.GoFiles {
            fileToPkg[file] = pkg.PkgPath
        }
    }
    return fileToPkg
}
```

It does **not** index `pkg.TestGoFiles` or `pkg.XTestGoFiles`. When `trackExternalUsages` resolves usage locations to packages, any usage from a test file fails the lookup and is silently dropped:

```go
for _, loc := range allUsages {
    usagePkg, ok := fileToPkg[loc.Path]
    if !ok {
        continue  // ← test file usages silently skipped
    }
    ...
}
```

## Consequence

Symbols consumed **only** by external test packages are misclassified:
- `[PRIVATE]` when the test reference would qualify as an external use
- `[DEAD]` when the test reference is the **only** use

Current known victim: `analysistest.MockSymbolIndex` — used via struct embedding in `types_error_test.go` (package `analysis_test`), but flagged `[PRIVATE]`.

## Recommended Fix

Extend `buildFileToPkgMap` to include test files:

```go
func (a *defaultDeadCodeAnalyzer) buildFileToPkgMap(pkgs []*packages.Package) map[string]string {
    fileToPkg := make(map[string]string)
    for _, pkg := range pkgs {
        for _, file := range pkg.GoFiles {
            fileToPkg[file] = pkg.PkgPath
        }
        for _, file := range pkg.TestGoFiles {
            fileToPkg[file] = pkg.PkgPath  // internal test files share package path
        }
        for _, file := range pkg.XTestGoFiles {
            fileToPkg[file] = pkg.PkgPath + "_test"  // external test package
        }
    }
    return fileToPkg
}
```

Note: `XTestGoFiles` use the `_test` suffix convention (e.g. `example.com/pkg` → `example.com/pkg_test`). The `trackExternalUsages` comparison `pkgBase != objBase` will then correctly recognize external test packages as external consumers.

## Impact

[...]

PR #447, opened 2026-05-16
fix(analysis): resolve type aliases in harvestObjectSymbols to restore interface propagation

thptcnec:
## Problem

`dead_code_graph` produced false positives for all `MockSymbolIndex` and `MockSecurityProvider` methods (9 findings). The root cause was a type alias blind spot in `harvestObjectSymbols`: Go 1.22+ returns `*types.Alias` from `TypeName.Type()` for declarations like `type SymbolIndex = symbolIndex`, but the code only checked for `*types.Named`.

## Fix

3 lines added to resolve `*types.Alias` via `types.Unalias` before the `*types.Named` type assertion:

```go
t := tn.Type()
if alias, ok := t.(*types.Alias); ok {
    t = types.Unalias(alias)
}
```

## Before vs After

| Before | After |
|--------|-------|
| 11 dead-code findings | 3 dead-code findings |
| 8 `[DEAD] (MockSymbolIndex).*` false positives | 0 — all eliminated |
| `[PRIVATE] MockSymbolIndex` | Remaining (separate issue #446) |
| `[DEAD] NewDeadCodeAnalyzer` / `NewIndexer` | True positives from prior dead code cleanup |

Closes #445

### DIFF
--- a/internal/tools/analysis/harvest.go
+++ b/internal/tools/analysis/harvest.go
@@ -148,7 +148,11 @@ func (a *defaultDeadCodeAnalyzer) harvestObjectSymbols(obj types.Object, fset *t
 
 	// Capture exported methods
 	if tn, ok := obj.(*types.TypeName); ok {
-		if named, ok := tn.Type().(*types.Named); ok {
+		t := tn.Type()
+		if alias, ok := t.(*types.Alias); ok {
+			t = types.Unalias(alias)
+		}
+		if named, ok := t.(*types.Named); ok {
 			a.harvestNamedMethods(named, fset, state)
 			if itf, ok := named.Underlying().(*types.Interface); ok {
 				a.harvestInterfaceMethods(itf, fset, state)


thptcnec:
The PR correctly addresses the false positives generated in Go 1.22+ by accurately resolving `*types.Alias` types into their underlying `*types.Named` declarations. The fix is sound and resolves the blind spot in `harvestObjectSymbols`.

**Suggestion (Optional Refactor):**
Since `types.Unalias` automatically returns the original type if it is not an alias type, you can simplify the logic by dropping the type assertion entirely:

```go
t := types.Unalias(tn.Type())
if named, ok := t.(*types.Named); ok {
    // ...
}
```

The PR is ready to merge as is!

thptcnec:
## Architect Review — Ready to Merge ✅ (with one note)

**Verdict:** The fix is correct, surgical (+5/-1), and . Local verification passed:
- `go build ./...` — clean
- `go test ./internal/tools/analysis/... -run TestHarvest` — PASS

### Architectural assessment
| Dimension | Result |
|---|---|
| Scope | Single adapter file; no boundary leaks |
| Correctness | Correctly resolves the Go 1.22+ `*types.Alias` blind spot before the `*types.Named` assertion |
| Layering | Stays inside `internal/tools/analysis` — no contamination of domain/transport layers |
| Risk | Low — purely additive type-narrowing; existing `*types.Named` path is preserved |

### [REFACTOR] Optional simplification (non-blocking)
`types.Unalias` is documented as a safe no-op on non-alias types, so the intermediate type assertion is redundant — matching the suggestion already noted in the PR thread:

```go
if named, ok := types.Unalias(tn.Type()).(*types.Named); ok {
    a.harvestNamedMethods(named, fset, state)
    if itf, ok := named.Underlying().(*types.Interface); ok {
        a.harvestInterfaceMethods(itf, fset, state)
    }
}
```
This collapses 4 lines to 1 and removes a transient variable. Worth squashing into this PR if not yet merged.

### Caveat before clicking merge
`reviewDecision` is currently **empty** — no formal approving review is recorded; the only comment is author self-review. If `dev` branch protection requires an approving review, request one from a second maintainer first. Otherwise: 🟢 ship it.

PR #451, opened 2026-05-16
fix: replace singleflight with cycle-bound sync.Once for implementations cache (#449)

thptcnec:
## Summary

Fixes the flaky `TestGetImplementations_SingleflightCoalescing` test by replacing `singleflight.Group` with a per-refresh-cycle `implCacheEntry` that bundles a `sync.Once` gate with its computed result.

## Root Cause

`singleflight.Do` forgets the key the instant the callback returns. Straggler goroutines that passed the fast-path cache check but haven't yet registered their `Do` call start a redundant second execution. The existing fast path only helps on the second call onward — it cannot close the window between the nil-check and `Do`.

## Fix

Replace the shared mutable `implementations` map + `sfGroup singleflight.Group` with a single `*implCacheEntry` field. Each `Refresh` cycle allocates a fresh entry; old computations write to abandoned entries that get GC'd, never corrupting the current cycle.

```
type implCacheEntry struct {
    once  sync.Once
    impls map[string][]string
}
```

`computeImplementationsLazy` captures the cache entry + pkgs snapshot under `RLock`, then uses `cache.once.Do(...)` — no key-forgetting, no stale writeback.

## Changes

| File | Change |
|------|--------|
| `index.go` | Add `implCacheEntry` struct; replace `implementations`+`sfGroup` with `*implCacheEntry`; update `newIndexer` and `updateState`; remove `singleflight` import |
| `index_impl.go` | Rewrite `computeImplementationsLazy` with cycle-bound cache; simplify `GetImplementations` |
| `index_impl_test.go` | Update cache invalidation and comments |
| `index_singleflight_test.go` → `index_once_test.go` | Rename; update comments |
| `go.mod` / `go.sum` | Remove `golang.org/x/sync/singleflight` |

## Verification

- `make check-full` — all gates pass (fmt, lint, vet, vulncheck, test, test-race, dead-code, coverage)
- `TestGetImplementations_SingleflightCoalescing -count=20 -race` — 20/20 deterministic
- Net −25 lines of code

## Acknowledgements

Co-authored-by: @thptcnec (cycle-bound cache entry design)

### DIFF
--- a/internal/tools/analysis/index.go
+++ b/internal/tools/analysis/index.go
@@ -6,12 +6,10 @@ package analysis
 import (
 	"context"
 	"go/token"
+	"golang.org/x/tools/go/packages"
 	"path/filepath"
 	"sync"
 	"time"
-
-	"golang.org/x/sync/singleflight"
-	"golang.org/x/tools/go/packages"
 )
 
 // location represents a position in a source file.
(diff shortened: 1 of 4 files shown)


thptcnec (commented):
## Architectural Review: PR 451

The changes perfectly capture the **"Cycle-Bound Cache Entry"** design discussed in Issue #449. 

### What looks great:
1. **Thread Safety**: Replacing the shared `implementations` map and `sfGroup` with `implsCache *implCacheEntry` ensures safe garbage collection of old generational cycles without cross-cycle poisoning.
2. **Locking Clean-up**: Capturing `idx.implsCache` and `idx.pkgs` under a quick `RLock` and then moving all execution into the lock-free `sync.Once` gate eliminates the need for writeback locking entirely. This solves the TOCTOU data race natively.
3. **Removal of External Dependencies**: Trimming out `golang.org/x/sync/singleflight` natively simplifies the module dependencies.

### Findings:
- Impact: **[REFACTOR / TECHNICAL DEBT FIXED]** 
- Assessment: **APPROVED**.

The implementation is highly idiomatic Go and cleanly adheres to the requested architecture and synchronization semantics.

PR #513, opened 2026-05-21
fix(ui): strip ANSI codes in TestHistory_RenderTextFallback assertion (#511)

thptcnec:
## Summary

Fixes flaky test `TestHistory_RenderTextFallback/raw=false_uses_glamour_renderer` that fails during full suite runs due to ANSI color escape sequences emitted by glamour.

## Root Cause

The glamour renderer uses `WithAutoStyle()`, which detects terminal capabilities at render time. When the full test suite runs, prior packages pollute global state (env vars like `TERM`/`COLORTERM`), causing glamour to emit ANSI CSI sequences. These split `"hello world"` into separate color spans, so `strings.Contains(output, "hello world")` fails.

## Fix

Added a `stripANSI` helper that removes all CSI escape sequences before asserting. This decouples the content assertion from terminal formatting state. Production code (`WithAutoStyle`) is untouched — it remains correct.

## Verification

- `go test -count=1 -run TestHistory_RenderTextFallback ./internal/ui/...` → **PASS**

Closes #511

### DIFF
--- a/internal/ui/history_test.go
+++ b/internal/ui/history_test.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"errors"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


thptcnec:
I agree with the approach. Decoupling test assertions from implicit terminal state (like `TERM`/`COLORTERM` env vars affecting glamour) is a necessary and correct fix to eliminate CI flakiness.

**[TECHNICAL DEBT]**
While this works, defining `ansiRE` and `stripANSI` locally here, and having a similar/identical helper in `browser_test.go` (as seen in PR #514), introduces duplicated test utilities across packages.

**Scalable Pattern:**
Extract this utility into a shared test package (e.g., `internal/testutil.StripANSI(s string) string`). This adheres to DRY principles, centralizes regex compilation, and makes UI testing more consistent across the module.

Approved for the immediate flakiness fix, but recommend a follow-up PR to consolidate test helpers.

PR #516, opened 2026-05-21
fix(cli): use capturerOverride for deterministic TTY check in browse test (#515)

thptcnec:
## Summary

Fixes flaky test `TestBrowseCommand_NilInteractorRef_DoesNotPanic` that panics during full suite runs due to TTY detection state pollution — same class of bug as #511 and #512.

## Root Cause

The test relied on the real capturer's `IsTTY(os.Stdout)` returning `false`. When the full test suite runs, prior packages pollute TTY detection global state (terminal env vars, `os.Stdout` state), causing `IsTTY` to return `true`. Execution proceeds past the TTY guard to `GetUnifiedHistoryProvider` on a mock that never had that expectation set up — panic.

## Fix

Switched from `executeBrowseCommand` (cobra pipeline) to direct `c.runBrowse()` with `capturerOverride: &mockCapturerInteractor{isTTY: false}`. This makes the TTY check deterministic regardless of ambient state. Matches the established pattern in `TestBrowseCommand_RunBrowse_PostTTY` in the same file.

## Verification

- `go test -count=1 -run TestBrowseCommand_NilInteractorRef_DoesNotPanic ./internal/cli/...` → **PASS**

Closes #515

### DIFF
--- a/internal/cli/cmd_browse_test.go
+++ b/internal/cli/cmd_browse_test.go
@@ -189,7 +189,17 @@ func TestBrowseCommand_NilInteractorRef_DoesNotPanic(t *testing.T) {
 		Interactor:   nil, // explicit nil
 	}
 
-	err := executeBrowseCommand(cmdCtx, []string{"browse"})
+	// Use capturerOverride to force a deterministic TTY check result.
+	// Without this, prior test packages can pollute TTY detection global
+	// state (os.Stdout, terminal env vars), causing IsTTY to return true
+	// and execution to reach GetUnifiedHistoryProvider on an unprepared
+	// mock — same class of bug as #511/#512.
+	c := &browseCommand{
+		ctx:              cmdCtx,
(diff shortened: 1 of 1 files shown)


thptcnec:
I agree with this approach. 

**[BEST PRACTICE]**
This is an excellent fix. Relying on global state or ambient environment variables (like real `os.Stdout` properties for TTY detection) is a common anti-pattern that leads to flaky tests in Go. 

By injecting the mock (`capturerOverride: &mockCapturerInteractor{isTTY: false}`), you are properly utilizing Dependency Injection to enforce deterministic behavior in your test suite. Bypassing the cobra pipeline (`executeBrowseCommand`) in favor of invoking `runBrowse` directly with an explicitly constructed struct is perfectly acceptable for isolating this unit test.

Approved. It cleanly addresses the root cause without touching production logic.

PR #579, opened 2026-05-26
fix(agenttest): fix data race in MockClock with sync.Mutex (#578)

### LINKED ISSUE #578: Data race in MockClock.Jitter() — unsynchronized slice append causes race detector failures in 8 orchestrator tests
## Summary

The Go race detector flags concurrent access to `MockClock.CalledMethods` in `MockClock.Jitter()` when multiple parallel subtests share the same `MockClock` instance and call `DefaultRetryPolicy.ShouldRetry()` concurrently.

## Reproduction

```bash
go test -race ./tests/integration/agent/orchestrator/ -run "TestDefaultRetryPolicy_Coverage"
```

Produces:

```
WARNING: DATA RACE
Read at 0x00c0000cc520 by goroutine 86:
  agenttest.(*MockClock).Jitter()
      .../internal/agent/agenttest/mock_clock.go:59 +0x3b
  orchestrator.(*DefaultRetryPolicy).ShouldRetry()
      .../internal/agent/orchestrator/engine_types.go:149 +0x126
  ...

Previous write at 0x00c0000cc520 by goroutine 85:
  agenttest.(*MockClock).Jitter()
      .../internal/agent/agenttest/mock_clock.go:59 +0xdc
  orchestrator.(*DefaultRetryPolicy).ShouldRetry()
      .../internal/agent/orchestrator/engine_types.go:149 +0x126
  ...
```

## Root Cause

1. `TestDefaultRetryPolicy_Coverage` creates **one** `MockClock` instance (`c := &agenttest.MockClock{}`) at the test function level (line ~1111 in `turn_engine_test.go`).
2. All subtests (`t.Run(...)`) are marked `t.Parallel()`, so they run concurrently.
3. Both the "Transient error" and "Rate limit error" subtests call `policy.ShouldRetry(c, ...)`, which calls `c.Jitter(...)`.
4. `MockClock.Jitter()` (line 59 of `mock_clock.go`) appends to `m.CalledMethods` — a shared unsynchronized slice:

```go
func (m *MockClock) Jitter(base float64) float64 {
    m.CalledMethods = append(m.CalledMethods, "Jitter")  // RACE
    return base
}
```

5. This is **not** specific to `Jitter`. Every method on `MockClock` (`Now`, `Sleep`, `After`, `NewTicker`, `Jitter`) appends to `CalledMethods` without synchronization.

## Affected Tests (8 failures)

All in `tests/integration/agent/orchestrator`:

- `TestTurnEngine_MaxTurnsLimit`
- `TestDefaultRetryPolicy_Coverage` (subtest: "Rate limit error")
- `TestTurnEngine_ValidatePayloadLimits`
- `TestTurnEngine_RecoveryLogic_TerminalAndContext`
- `TestTurnEngine_BackgroundCostTracking`
- `TestTurnEngine_ToolCallLoopDetection_Table` (subtest: "Calls across exactly 2 turns")
- `TestTurnEngine_ToolCallLoopDetection`
- `TestTurnEngine_MultiStepLoopDetection`

## Proposed Fix

**Option A (preferred — fix the root cause):** Add a `sync.Mutex` to `MockClock` and guard all mutations to `CalledMethods` and `CalledNow`:

```go
type MockClock struct {
    mu            sync.Mutex
    CurrentTime   time.Time
    CalledNow     int
    CalledMethods []string
}

func (m *MockClock) Jitter(base float64) float64 {
    m.mu.Lock()
    m.CalledMethods = append(m.CalledMethods, "Jitter")
    m.mu.Unlock()
    return base
}
```

Apply the same pattern to `Now()`, `Sleep()`, `After()`, and `NewTicker()`.

**Option B (test-only fix — move MockClock into each subtest):** In `TestDefaultRetryPolicy_Coverage`, create a new `MockClock` inside each `t.Run` instead of sharing one. But this only fixes one test — the root type remains unsafe for concurrent use.

## Severity

**[ARCHITECTURAL BLOCKER]** — `make check-full` fails at the `test-race` stage. This blocks any release that requires clean race-detector results.

## Additional Issue

During the same `test-race` run, there are also logged errors about `"event bus is nil or uninitialized"` and `"bus full"` from `TestTurnEngine_RecoveryLogic_TerminalAndContext` — these are likely secondary effects of the race or test isolation issues and should be tracked separately.

## Files Involved

| File | Role |
|------|------|
| `internal/agent/agenttest/mock_clock.go` | Root cause — unsynchronized MockClock |
| `internal/agent/orchestrator/engine_types.go:149` | Caller — `DefaultRetryPolicy.ShouldRetry()` calls `Jitter()` |
| `tests/integration/agent/orchestrator/turn_engine_test.go:1111-1137` | Trigger — parallel subtests sharing one MockClock |

gosharplite:
Closes #578.

## Summary
Added `sync.Mutex` to `MockClock` and guarded all mutations to `CalledMethods` and `CalledNow` in all 5 writer methods (`Now`, `Sleep`, `After`, `NewTicker`, `Jitter`). Added a new concurrency regression test that exercises all methods under `-race`. The fix aligns `MockClock` with existing patterns in the same package (`MockCostTracker`, `MockHistoryManager`).

## Changes
| File | Change |
|------|--------|
| `internal/agent/agenttest/mock_clock.go` | Added `sync.Mutex`, guarded `CalledNow`/`CalledMethods` in 5 methods |
| `internal/agent/agenttest/mock_clock_test.go` | New concurrency regression test (5 goroutines × 20 iterations) |

## Verification
| Check | Status |
|-------|--------|
| `go test -race ./tests/integration/agent/orchestrator/` | PASS — all 8 affected tests |
| `go test -race ./internal/agent/... ./tests/integration/agent/...` | PASS — 15 packages |
| `make check-full` (fmt, tidy, build, lint, arch, vulncheck, test, dead-code) | PASS |
| `test-race` (agent packages) | PASS |
| `test-race` (tools/analysis) | Pre-existing timeout — unrelated to this change |
| Target coverage | N/A — test infrastructure fix |

### DIFF
--- a/internal/agent/agenttest/mock_clock.go
+++ b/internal/agent/agenttest/mock_clock.go
@@ -8,6 +8,7 @@
 package agenttest
 
 import (
+	"sync"
 	"time"
 
 	"github.com/gosharplite/tell-me-go/internal/pkg/clock"
@@ -20,18 +21,23 @@ import (
 // it never blocks. NewTicker() returns a *MockTicker fed from the
 // same After() channel.
 type MockClock struct {
+	mu            sync.Mutex
 	CurrentTime   time.Time
 	CalledNow     int
 	CalledMethods []string
 }
 
 func (m *MockClock) Now() time.Time {
+	m.mu.Lock()
 	m.CalledNow++
 	m.CalledMethods = append(m.CalledMethods, "Now")
-	if m.CurrentTime.IsZero() {
+	t := m.CurrentTime
+	m.mu.Unlock()
+
+	if t.IsZero() {
 		return time.Now()
 	}
-	return m.CurrentTime
+	return t
 }
 
 func (m *MockClock) Since(t time.Time) time.Duration {
@@ -39,24 +45,36 @@ func (m *MockClock) Since(t time.Time) time.Duration {
 }
 
 func (m *MockClock) Sleep(d time.Duration) {
+	m.mu.Lock()
 	m.CalledMethods = append(m.CalledMethods, "Sleep")
 	m.CurrentTime = m.CurrentTime.Add(d)
+	m.mu.Unlock()
 }
 
 func (m *MockClock) After(d time.Duration) <-chan time.Time {
+	m.mu.Lock()
 	m.CalledMethods = append(m.CalledMethods, "After")
+	t := m.CurrentTime.Add(d)
+	m.mu.Unlock()
+
 	c := make(chan time.Time, 1)
-	c <- m.CurrentTime.Add(d)
+	c <- t
 	return c
 }
 
 func (m *MockClock) NewTicker(d time.Duration) clock.Ticker {
+	m.mu.Lock()
 	m.CalledMethods = append(m.CalledMethods, "NewTicker")
+	m.mu.Unlock()
+
 	return &MockTicker{CVal: m.After(d)}
 }
 
 func (m *MockClock) Jitter(base float64) float64 {
+	m.mu.Lock()
 	m.CalledMethods = append(m.CalledMethods, "Jitter")
+	m.mu.Unlock()
+
 	return base
 }
 
--- a/internal/agent/agenttest/mock_clock_test.go
+++ b/internal/agent/agenttest/mock_clock_test.go
@@ -1,0 +1,44 @@
+// Copyright (c) 2026 gosharplite@gmail.com
+// SPDX-License-Identifier: MIT
+
+package agenttest
+
+import (
+	"sync"
+	"testing"
+	"time"
+)
+
+// TestMockClock_RaceDetection verifies that concurrent writes to
+// CalledMethods and CalledNow on MockClock do not trigger the race
+// detector. This test is a precondition for adding sync.Mutex.
+func TestMockClock_RaceDetection(t *testing.T) {
+	m := &MockClock{}
+
+	var wg sync.WaitGroup
+	const goroutines = 5
+	const iterations = 20
+
+	wg.Add(goroutines)
+	for g := 0; g < goroutines; g++ {
+		go func() {
+			defer wg.Done()
+			for i := 0; i < iterations; i++ {
+				// Mix of all mutating methods
+				m.Now()
+				m.Sleep(1 * time.Millisecond)
+				m.After(1 * time.Millisecond)
+				m.NewTicker(1 * time.Millisecond)
+				m.Jitter(1.0)
+			}
+		}()
+	}
+	wg.Wait()
+
+	// CalledMethods should have goroutines*iterations*6 entries
+	// (NewTicker also appends "After" via its internal call to After)
+	expected := goroutines * iterations * 6
+	if len(m.CalledMethods) != expected {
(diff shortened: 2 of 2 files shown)


gosharplite:
## Architect(A) Review — deepseek-pro

**Approved — no gating issues. Ready for Architect(B) review.**

Reviewed `internal/agent/agenttest/mock_clock.go` (+sync.Mutex, 5 guarded methods) and the new regression test in `mock_clock_test.go`.

**Findings:**
- The mutex fix correctly addresses the data race in `CalledMethods`/`CalledNow`. Lock scope is tight; no lock is held across channel operations or re-entrant calls (`NewTicker` → `After` is safe — the lock is released between them).
- The regression test passes under `-race` (5 goroutines × 20 iterations × all 5 mutating methods).
- Verified against existing patterns: consistent with `MockCostTracker` (unexported `mu`), though diverges from `MockHistoryManager` (exported `Mu`).
- 4 minor advisory items noted (mutex visibility, test assertion fragility, doc gap, pre-existing `Since` quirk) — none block merge.

Build, lint, and agent-package tests all pass. No security concerns, no goroutine leaks, no deadlock potential.

gosharplite:
## Architect(B) Review — vertex-pro

### Architectural Review Summary

**Files Reviewed:** 
* `internal/agent/agenttest/mock_clock.go` (Mutex implementation and locking patterns)
* `internal/agent/agenttest/mock_clock_test.go` (Concurrency regression test)

#### Findings & Advisory Items
Following a cross-architect review, the fix for the write-write data race is verified, but we identified a few architectural and technical debt items to keep in mind:

* **[ADVISORY] Unsynchronized Read Access (Downgraded from Gating):** The fields `CalledMethods`, `CalledNow`, and `CurrentTime` remain exported without read synchronization. While this *could* cause a read-write data race, it does not actively trigger one in the current test corpus because tests (like the new regression test) only read these fields after goroutines have quiesced (`wg.Wait()`). For future bulletproofing, consider adding thread-safe getter methods that return a copy of the slice.
* **[ADVISORY] Fragile Lock Ordering / Deadlock Trap:** In `NewTicker()`, the code avoids using `defer m.mu.Unlock()` before calling `m.After()`. If a future maintainer refactors this to use standard `defer` semantics, it will instantly deadlock because `m.After()` also acquires the lock. We highly recommend extracting the mutation into a private helper to clean up the boilerplate and eliminate this trap entirely:
  ```go
  func (m *MockClock) recordCall(method string) {
      m.mu.Lock()
      defer m.mu.Unlock()
      m.CalledMethods = append(m.CalledMethods, method)
  }
  ```
* **[ADVISORY] Flaky Test Architecture:** Sharing a single stateful `MockClock` instance across parallel orchestrator subtests is an anti-pattern. While this PR resolves the *memory-level* race, logical assertions on `CalledMethods` will be inherently non-deterministic due to goroutine scheduling. We recommend a follow-up refactor to ensure each parallel subtest instantiates its own isolated `MockClock`.

#### Final Assessment
After peer confirmation, it is agreed that **no gating issues remain**. The active build-breaking data race is resolved. This PR is ready for human review and merge.

PR #756, opened 2026-06-09
fix(events): add whitebox tests for Flush cancellation paths (#749)

gosharplite:
Closes #749.

## Summary

Added two deterministic whitebox tests to `internal/domain/events/event_bus_lifecycle_whitebox_test.go` covering the previously untested cancellation branches in `SimpleEventBus.Flush`:

- **`TestFlush_CallerContextCancellation`** — covers `<-ctx.Done()` branch (caller context cancellation while flushWaiter is blocked in `cond.Wait()`)
- **`TestFlush_BusShutdownDuringFlush`** — covers `<-b.ctx.Done()` branch (bus internal context cancellation while flushWaiter is blocked)

Both tests use a deterministic three-phase mutex handoff (no `time.Sleep`) ensuring `cancelFlushWaiter` is exercised while `flushWaiter` is verifiably blocked in `cond.Wait()`.

## Verification

| Check | Status |
|-------|--------|
| `make check-full` | PASS (10/10 steps; flaky `TestRunCommand_NonExitErrorWaitPath_SIGKILL` in workspace package is pre-existing) |
| `go test -race ./internal/domain/events/...` | PASS |
| `Flush` coverage | 100% |
| `flushWaiter` coverage | 100% |
| `cancelFlushWaiter` coverage | 100% |
| `events` package coverage | 99.3% |

### DIFF
--- a/internal/domain/events/event_bus_lifecycle_whitebox_test.go
+++ b/internal/domain/events/event_bus_lifecycle_whitebox_test.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"errors"
 	"log/slog"
-	"runtime"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


gosharplite:
**Architect(A) Review — PR #756**

**Scope:** `internal/domain/events/event_bus_lifecycle_whitebox_test.go` (+113 lines), plus structural analysis of `event_bus_lifecycle.go` and `event_bus.go` (SimpleEventBus struct, field visibility, context wiring).

---

### Findings

**Gating Issues:** None.

**Advisory Items (4):**

1. **`runtime.Gosched()` comment is misleading** — The adjacent comment implies scheduler yielding is a correctness requirement. `pendingMu.Lock()` on the next line is the true synchronization barrier; `Gosched` is a harmless performance hint. Clarify the comment so future readers don't mistake this for a `time.Sleep`-style hack.

2. **No explicit `done`-channel-closure verification** — Both tests verify `Flush` returned the expected error but don't independently prove `flushWaiter` exited. A bug that returned `ErrBusClosed` without setting `*cancelled=true` would produce a silent goroutine leak and the test would still pass. The mutex handoff makes this extremely unlikely, but adding `goleak` to the whitebox test package (mirroring `events_test/TestMain`) would close this gap definitively.

3. **`TestFlush_BusShutdownDuringFlush` naming vs. mechanism** — The test name suggests a shutdown scenario but the test calls `bus.cancel()` directly, bypassing `Shutdown()`'s `closed=true` path. This is correct for unit-level branch coverage; the integration path is covered by `TestSimpleEventBus_Flush_BusContextCancellation` in the blackbox suite. Clarify the doc comment to distinguish the two.

4. **Pre-existing coverage gaps** — `WaitStarted` (66.7%) and `signalStarted` (75.0%) have untested nil-receiver branches. Not introduced by this PR; worth a follow-up issue.

---

### Assessment: **PASS — Ready for Architect(B)**

All four acceptance criteria from #749 are satisfied. The three-phase mutex handoff is provably deterministic — `pendingMu.Lock()` in Phase 3 blocks until `flushWaiter` has released the mutex via `cond.Wait()`, establishing a happens-before relationship with zero scheduler dependence. The `-race` detector passes at `-count=3`. Coverage for `Flush`, `flushWaiter`, and `cancelFlushWaiter` is 100%.

No architectural blockers. No concurrency defects. Whitebox field access is legitimate within `package events`. Clean domain-layer isolation maintained.

gosharplite:
### 🏛️ Architect(B) Independent Review — PR #756

**🔍 Initial Finding**
During the initial review, I identified one gating issue: the test implementation relied on non-deterministic synchronization using `runtime.Gosched()`. This is an anti-pattern in Go testing that creates a race condition against the scheduler, leading to flaky and unreliable tests.

**🛠️ Resolution**
The applied fix successfully addresses this by using a **pre-cancellation approach**. By explicitly cancelling the context *before* executing the system under test, the test now algebraically forces the selection of the cancellation branch. This entirely removes the reliance on timing and runtime scheduling, guaranteeing deterministic test execution.

**✅ Final Assessment**
I have verified the updated implementation. There are no remaining gating issues, and the code now fully aligns with idiomatic Go concurrency and testing patterns. 

**Status:** The PR is architecturally sound and is **ready for human merge**.

PR #960, opened 2026-06-20
fix(persistence): fix SQLITE_BUSY with DSN WAL mode and retry (#952)

gosharplite:
Closes #952.

## Summary
Fixes `SQLITE_BUSY (database is locked)` errors when multiple `manage_tasks` mutations are invoked in parallel. The root cause was that `PRAGMA journal_mode=WAL` and `PRAGMA busy_timeout=5000` were set via `ExecContext` on only the initial connection — subsequent pool connections got defaults and failed instantly under write contention.

### Changes
1. **DSN-level PRAGMA configuration** (`sqlite_db.go`): Moved WAL mode + busy_timeout into the DSN connection string (`?_journal_mode=WAL&_busy_timeout=5000`), ensuring every connection from the pool inherits them. Removed the redundant `ExecContext`-based PRAGMA loop.
2. **WAL checkpoint on close** (`state.go`): Added `PRAGMA wal_checkpoint(TRUNCATE)` in `Close()` to prevent unbounded WAL file growth.
3. **Internal retry on SQLITE_BUSY** (`persistence_tools.go`): Added `retryOnBusy` helper with exponential backoff (100ms→200ms→400ms, max 3 retries) wrapping all four write paths (`addTask`, `updateTask`, `deleteTask`, `clearTasks`).
4. **Test DSN alignment** (`sqlite_store_test.go`): Updated `setupSQLite` to use the same DSN PRAGMA pattern as production.

## Verification
| Check | Status |
|-------|--------|
| make check-full | ✅ PASS (all 10 steps) |
| lint | ✅ 0 issues |
| race tests | ✅ All packages clean |
| vulncheck | ✅ No vulnerabilities |
| dead-code | ✅ No dead code |
| Coverage (persistence) | ≥99.6% |
| Coverage (workspace) | ≥98.3% |

### DIFF
--- a/internal/infrastructure/persistence/sqlite_db.go
+++ b/internal/infrastructure/persistence/sqlite_db.go
@@ -23,21 +23,9 @@ var sqlOpenFn = sql.Open
 
 // initSQLiteDB opens the SQLite database and runs migrations.
 func initSQLiteDB(ctx context.Context, dbPath string) (*sql.DB, error) {
-	db, err := sqlOpenFn("sqlite", dbPath)
+	db, err := sqlOpenFn("sqlite", dbPath+"?_journal_mode=WAL&_busy_timeout=5000")
 	if err != nil {
 		return nil, fmt.Errorf("failed to open sqlite db: %w", err)
-	}
-
-	// Apply pragmas for resilience
-	pragmas := []string{
-		"PRAGMA journal_mode=WAL",
-		"PRAGMA busy_timeout=5000",
-	}
-	for _, p := range pragmas {
-		if _, err := db.ExecContext(ctx, p); err != nil {
-			_ = db.Close()
-			return nil, fmt.Errorf("failed to set pragma %q: %w", p, err)
-		}
 	}
 
 	// Run schema migrations
--- a/internal/infrastructure/persistence/sqlite_store_test.go
+++ b/internal/infrastructure/persistence/sqlite_store_test.go
@@ -19,7 +19,12 @@ import (
 
 func setupSQLite(t *testing.T) *sql.DB {
 	t.Helper()
-	db, err := sql.Open("sqlite", ":memory:")
+	// Mirror the production initSQLiteDB DSN parameters (WAL journal mode
+	// and busy_timeout) so that tests exercise the same connection behavior.
+	// While in-memory databases are single-connection and don't benefit from
+	// WAL, this keeps the pattern consistent and prevents accidental
+	// regressions if future tests switch to file-backed databases.
+	db, err := sql.Open("sqlite", ":memory:?_journal_mode=WAL&_busy_timeout=5000")
 	if err != nil {
 		t.Fatalf("Failed to open test database: %v", err)
 	}
--- a/internal/infrastructure/persistence/state.go
+++ b/internal/infrastructure/persistence/state.go
@@ -113,6 +113,15 @@ func (s *sessionState) hydrateInfo(ctx context.Context, storageType string, repo
 
 func (s *sessionState) Close() error {
 	if s.db != nil {
+		// Best-effort WAL checkpoint to merge WAL file back into main DB
+		// and truncate it to zero to reclaim disk space.
+		// A failure here is non-fatal — it only means the WAL file
+		// persists until the next connection checkpoint.
+		if _, err := s.db.Exec("PRAGMA wal_checkpoint(TRUNCATE)"); err != nil {
+			// Logged at debug level since this is a cleanup concern,
+			// not a correctness issue.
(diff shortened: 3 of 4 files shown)


gosharplite:
## Architecture Review: PR #960 — `fix(persistence): fix SQLITE_BUSY with DSN WAL mode and retry (#952)`

### Scope Reviewed

| Area | Files |
|---|---|
| **DSN PRAGMA change** | `internal/infrastructure/persistence/sqlite_db.go` |
| **WAL checkpoint on close** | `internal/infrastructure/persistence/state.go` |
| **Retry logic** | `internal/tools/workspace/persistence_tools.go` |
| **Test DSN alignment** | `internal/infrastructure/persistence/sqlite_store_test.go` |
| **Integration context** | `sqlite_store.go`, `task_service.go`, `ports/repository.go`, `sqlite_db_test.go` |

Also verified: all existing persistence and workspace tests pass (`go test -short`), no import cycles introduced, `go vet` clean.

---

### Findings Summary

**No gating issues.** The PR correctly addresses the root cause of #952 by moving `PRAGMA journal_mode=WAL` and `PRAGMA busy_timeout=5000` into the DSN connection string, ensuring **every** pool connection inherits these settings — not just the initial one.

---

### 5 Advisory Items (non-blocking)

#### 1. String-based error detection is fragile
**`persistence_tools.go:61`** — `strings.Contains(err.Error(), "database is locked")` also matches `SQLITE_LOCKED` (different semantics) and silently breaks if `modernc.org/sqlite` changes error text. Prefer `errors.As` with `*sqlite.Error` and `Code()` comparison. This requires moving the retry into the persistence package (see item 2).

#### 2. `retryOnBusy` is in the wrong architectural layer
**`persistence_tools.go:50-69`** — Retrying SQLITE_BUSY is a persistence concern, but the helper lives in the workspace/tool layer. This forces the string-based detection (workspace can`'t import `modernc.org/sqlite`) and retries at the service boundary rather than around the raw `ExecContext` calls in `sqlite_store.go`.

#### 3. Zero test coverage for the retry function
**`persistence_tools.go:50-69`** — No unit tests exercise `retryOnBusy`. The function`'s six code paths (success, retry-then-success, exhausted retries, non-BUSY error, context cancellation, `select` branch) are all untested. Small, pure function — easy to cover.

#### 4. Misleading test name after PRAGMA loop removal
**`sqlite_db_test.go:174`** — `TestInitSQLiteDB_PragmaFailure` was written for the now-removed `ExecContext` PRAGMA loop. It still passes (exercises the `createTables` error path instead), but the name should be updated to `TestInitSQLiteDB_InvalidPath` or similar.

#### 5. Pre-existing ReadAll-before-write in `UpdateTask`/`DeleteTask`
**`task_service.go:58-67, 94-105`** — Not introduced by this PR, but `UpdateTask` and `DeleteTask` load **all** tasks into memory just to verify a single ID exists (O(N) query). The PR`'s retry wraps only the final `store.Update`/`store.Delete`, not the read-before-write — if a concurrent write removes the task between `Query` and `Update`, the retry will fruitlessly retry. File separately.

---

### Final Assessment

**✅ No gating issues — ready for Architect(B) / human review.**

The three highest-value follow-ups: (1) move `retryOnBusy` to the persistence layer with error-code-based detection, (2) add unit tests for the retry function, (3) rename the stale test. All can be addressed in follow-up PRs without blocking this fix.

gosharplite:
# Architecture & Code Review: PR #960

**Status: APPROVED** ✅

I have completed a comprehensive architectural and code review of the changes introduced in this PR, focusing on system design, maintainability, and concurrency patterns. There are **no gating issues** or architectural blockers.

### 📝 Advisory Findings (Non-Blocking)
The PR is structurally sound, but I recommend addressing the following advisory items either before merging or as fast-follow technical debt:

1. **Unreachable Code [TECHNICAL DEBT]:** There are segments of unreachable code present. These should be pruned to reduce maintenance overhead and prevent future developer confusion.
2. **Write-Only Retry Design [DOCUMENTATION]:** The current retry mechanism relies on a write-only design. Please add explicit architectural documentation (or a lightweight ADR) detailing this constraint and its specific trade-offs so future maintainers understand the bounded context.
3. **Timer Allocation (`time.After`) [REFACTOR]:** There is usage of `time.After` in the code. As a reminder, `time.After` does not release its underlying timer until it fires, which can cause memory leaks if used inside fast-running loops or `select` blocks that exit early. If this is in a hot path, consider refactoring to use `time.NewTimer()` with explicit `Stop()` and drain calls.

### 🏁 Final Assessment
The design adheres to our architectural standards and introduces no scalability bottlenecks. **Approved — no gating issues, ready for human merge.**

PR #978, opened 2026-06-21
fix(telemetry): cache external daily cost to eliminate sync disk I/O in GetDailyCost hot path (#977)

gosharplite:
Closes #977.

## Summary
Replaced synchronous disk I/O in `GetDailyCost` with an O(1) cached-path implementation. Historical (external) daily cost from `global_costs.json` is now cached in `sessionCostTracker` on `Warmup()` or first `GetDailyCost()` call, and refreshed only on UTC-8 date rollover. The hot path in `publishTurnStatus` (called during `PhaseInference` and `PhasePersisting`) no longer acquires `ledgerMu` or reads from the filesystem.

### Changes
- `metrics_tracker.go`: Added `cachedExternalDailyCost`, `cachedDate`, `currentSessionID` fields; extracted `refreshExternalDailyCache` method; simplified `GetDailyCost` to O(1); updated `Warmup()` to eagerly populate the cache
- `metrics_daily_test.go`: Added 3 new tests (cached path, Warmup cache, date rollover refresh)

## Verification
| Check | Status |
|-------|--------|
| make check-full | PASS |
| lint (golangci-lint) | 0 issues |
| vulncheck | No vulnerabilities |
| test (all packages) | PASS |
| test-race | PASS (no races) |
| dead-code | No dead code |
| Test coverage | 99.6% (telemetry package) |
| GetDailyCost coverage | 100.0% |
| Warmup coverage | 100.0% |
| refreshExternalDailyCache coverage | 92.9% |

### DIFF
--- a/internal/infrastructure/telemetry/metrics_daily_test.go
+++ b/internal/infrastructure/telemetry/metrics_daily_test.go
@@ -5,6 +5,7 @@ package telemetry
 
 import (
 	"context"
+	"encoding/json"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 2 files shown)


gosharplite:
**Architectural Review — PR #978** (Architect A — DeepSeek Pro)

**Reviewed:** `metrics_tracker.go` (production), `metrics_daily_test.go` (tests), plus dependent files (`ledger.go`, `metrics.go`, `metrics_cost.go`, `metrics_summary.go`, `middleware.go`) to verify lock ordering and caller contracts.

---

### ✅ No Gating Issues

The core cache decomposition is mathematically correct:

```
refreshExternalDailyCache:  cached = calculateDailyCost(history, now, id) - t.totalCost
GetDailyCost:               return cached + t.totalCost
```

Since `calculateDailyCost` returns `Σ(external_today) + t.totalCost`, the subtraction correctly isolates external cost, and `GetDailyCost` re-adds the caller's live `t.totalCost` — which is the **Source of Truth** for the current session. Three new tests (`CachedPathAfterFirstCall`, `Warmup_PopulatesDailyCostCache`, `DateRolloverRefreshesCache`) validate this end-to-end, including the "delete ledger to prove no disk I/O" technique.

Lock ordering: `t.mu → ledgerMu` is consistent with all existing paths (`m.metricsMu → ledgerMu` in `recordCost` and `getCostSummary`). No cross-object mutex acquisition exists, so no deadlock risk.

Race safety: `GetDailyCost` and `Accumulate` are both serialized on `t.mu`. `cachedExternalDailyCost` is only read under `t.mu` and only mutated under `t.mu` + `ledgerMu`. No data races.

---

### 📝 Advisory Notes (non-blocking)

1. **Dual `currentSessionID` init** — Set in both `ensureInitialized()` (line 199) and `refreshExternalDailyCache()` (line 124). The values are identical (same inputs), but the field doc only mentions `ensureInitialized()`. Consider updating the comment or consolidating to a single init site.

2. **[TECHNICAL DEBT] Implicit coupling with `calculateDailyCost`** — `cachedExternalDailyCost = fullDaily - t.totalCost` depends on the invariant that `calculateDailyCost` always appends `t.totalCost`. If that function is ever refactored, the subtraction silently breaks. A comment on `calculateDailyCost` noting this dependency would protect future maintainers.

3. **Failure-mode semantics change** — `cachedDate` is set to `today` *before* the ledger read, so a transient I/O error poisons the cache for up to ~24h (returns `$0` external cost). Previously, every call retried. This is intentional per the code comment, but worth an ADR since it changes the failure behavior.

4. **Coverage gap** — The `logFile == ""` fast path inside `refreshExternalDailyCache` (lines 94–97) is not explicitly tested. This is the likely source of the 7.1% uncovered portion (92.9% reported).

---

### Verdict

**No gating issues. PR is correct, well-tested, and safe to merge.** Ready for Architect(B) review.

gosharplite:
**Architect(B) Independent Review** (Vertex Pro)

I have completed a second review of PR #978, focusing on the core changes in `internal/infrastructure/telemetry/metrics_tracker.go` and the associated test coverage in `metrics_daily_test.go`. My review specifically evaluated concurrency safety, error handling, and performance characteristics in the `GetDailyCost` hot path.

### Final Assessment
The gating issue regarding transient errors poisoning the daily cache has been successfully addressed. By deferring the `t.cachedDate` assignment to the success path, the system now safely recovers from temporary I/O or JSON parsing failures. The PR correctly achieves the O(1) performance goal for the hot path.

**Verdict: LGTM / Approved** (only non-blocking advisory items remain).

### Advisory Items (Non-Blocking)
To further optimize the hot path and ensure long-term stability, consider addressing the following items either before merging or as a fast-follow:

1. **Hot Path Memory Allocations:** 
   The `GetDailyCost` method still executes `time.FixedZone("UTC-8", -8*3600)` and `now.In(loc).Format("2006-01-02")` on every invocation. This generates minor but repeated heap allocations (a Location struct and a string). *Recommendation:* Calculate and cache the exact `time.Time` for the next UTC-8 midnight rollover during the refresh, and replace the string comparison with an allocation-free `now.After(cachedRolloverTime)` check.
   
2. **Floating-Point Subtraction Risk:**
   Deriving the external cost via `fullDaily - t.totalCost` could theoretically introduce minor floating-point precision artifacts. *Recommendation:* Refactor `calculateDailyCost` to isolate and return only the external sum, avoiding the addition and subsequent subtraction of the in-memory value.

Excellent work on eliminating the synchronous disk I/O bottleneck!

PR #1018, opened 2026-06-21
fix(agenttest): cover untested nil-func and boundary branches in mocks (#1011)

gosharplite:
Closes #1011.

## Summary
Added table-driven tests covering all 8 medium-priority BUSINESS_LOGIC gaps in `internal/agent/agenttest/` mock files:
- `mock_health_test.go`: Nil-func fallthrough tests for CheckAll and CheckComponent
- `mock_history_manager_test.go`: `startIdx < 0` boundary test for GetWindow
- `mock_session_provider_test.go`: Nil-func fallthrough tests for GetSettings, GetInfo, Close, GetHealthChecker
- `mock_ui_renderer_test.go`: Single-threaded and concurrent tests for SwapLogSystemMessageFn

No behavioral changes to mock implementations — only `*_test.go` files modified.

## Verification
| Check | Status |
|-------|--------|
| make check | PASS |
| make test-race | PASS |
| Coverage (agenttest) | 100.0% (up from 98.0%) |
| golangci-lint | 0 issues |

### DIFF
--- a/internal/agent/agenttest/mock_health_test.go
+++ b/internal/agent/agenttest/mock_health_test.go
@@ -176,3 +176,47 @@ func TestMockHealthCheckManager_RaceCondition(t *testing.T) {
 		t.Errorf("calledMethods length: got %d, want %d", len(methods), goroutines*iterations)
 	}
 }
+
+// TestMockHealthCheckManager_NilFuncs exercises the two nil-func fallthrough
+// branches in mock_health.go where CheckAllFunc or CheckComponentFunc is nil.
+func TestMockHealthCheckManager_NilFuncs(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name string
+		call func(m *MockHealthCheckManager)
+	}{
+		{
+			name: "CheckAll_nil_func",
+			call: func(m *MockHealthCheckManager) {
+				got, err := m.CheckAll(context.Background())
+				if got != nil {
+					t.Errorf("CheckAll nil func: got %+v, want nil", got)
+				}
+				if err != nil {
+					t.Errorf("CheckAll nil func: unexpected error %v", err)
+				}
+			},
+		},
+		{
+			name: "CheckComponent_nil_func",
+			call: func(m *MockHealthCheckManager) {
+				got, err := m.CheckComponent(context.Background(), ports.CompLLMProvider)
+				if got != nil {
+					t.Errorf("CheckComponent nil func: got %+v, want nil", got)
+				}
+				if err != nil {
+					t.Errorf("CheckComponent nil func: unexpected error %v", err)
+				}
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			m := &MockHealthCheckManager{}
+			tt.call(m)
+		})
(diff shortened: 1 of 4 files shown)


gosharplite:
## Architecture Review — PR #1018

**Reviewed**: 4 modified `*_test.go` files in `internal/agent/agenttest/`, plus their corresponding source mocks and the `ports.SessionInfo` type. Assessed for correctness, concurrency safety, idiom, and architectural boundaries.

### Findings

**Gating**: None.

**Advisory** (3 items, non-blocking):

1. **`mock_health_test.go`** — `TestMockHealthCheckManager_NilFuncs` verifies return values `(nil, nil)` but does not assert that `checkAllCalls`/`checkComponentCalls` counters and `calledMethods` are still incremented. The mock increments these *before* the nil-func guard (lines 33–35 of `mock_health.go`), and that contract could silently regress without snapshot assertions here.

2. **`mock_session_provider_test.go`** — `GetInfo_nil_func` only checks `Model`, but `ports.SessionInfo` has 5 fields (`Env`, `Paths`, `Model`, `Provider`, `ActiveToolkits`). This mirrors the existing `TestMockSessionProvider_GetInfo` pattern, so it is internally consistent; worth hardening later.

3. **`mock_ui_renderer_test.go`** — `TestMockUIRenderer_SwapLogSystemMessageFn` tests only non-nil → non-nil swaps. The real usage in `session/ui/helpers_test.go:syncBridge` starts with a nil `LogSystemMessageFn`, so a nil-initial → non-nil swap case would be valuable, though not a branch-coverage gap in the mock itself.

### Concurrency

All new tests use `t.Parallel()` correctly. `SwapLogSystemMessageFn_Concurrency` uses proper `go func(i int)` closure capture and `sync.WaitGroup`. The `-race` detector confirms no data races.

### Verdict: **NO_GATING**

All 8 targeted branches are covered. Mock behavior is unchanged. The PR is safe, idiomatic, and ready for Architect(B) review.

gosharplite:
### Independent Architect Review Summary (PR #1018)

**Scope of Review:** 
Independent assessment of system design, test robustness, and concurrency patterns.

**Gating Issues (Resolved / Disputed):**
* ~~**Loop Variable Capture before `t.Parallel()`**: Originally flagged two test files for missing explicit `tt := tt` capture in table-driven subtests.~~
  * **Resolution [DISPUTED]**: The primary Architect correctly pointed out that the project uses `go 1.26.4`. As of Go 1.22, loop variables are instantiated per-iteration, making the explicit `tt := tt` capture obsolete. I concede this point; the tests are structurally sound and this is no longer a gating issue.

**Advisory Items (Non-Blocking / Technical Debt):**
1. **Boolean Visibility (`SwapLogSystemMessageFn` test)**: The assertion visibility around boolean states could be improved. Consider explicitly logging the exact state transition failure rather than a generic boolean mismatch to aid future debugging.
2. **Panic Risk (`GetWindow` test)**: There is a minor panic risk in the test data setup/evaluation (e.g., potential out-of-bounds or nil reference). Recommend hardening this with defensive checks, though it does not block the current release.

**Final Assessment:**
With the parallel loop capture concerns cleared by Go 1.22+ semantics, there are **no remaining architectural blockers**. From my perspective, this PR is safe to proceed.

PR #1161, opened 2026-06-29
refactor: centralize OS path normalization (#1156)

gosharplite:
Closes #1156.

## Summary

Centralizes the duplicated OS path normalization logic that was independently implemented across 5 packages during the Windows compatibility fixes (#1144, #1146, #1148).

## New Package

internal/pkg/filepathutil with three functions:

| Function | Behavior |
|----------|----------|
| Normalize(path) | EvalSymlinks + recursive fallback + ToSlash |
| NormalizePath(path) | EvalSymlinks + recursive fallback (OS-native) |
| NormalizeKey(path) | Normalize + volume strip + ToLower |

## Replacement Sites

| Package | Before | After |
|---------|--------|-------|
| security/paths.go | p.resolveSymlinks() (10 lines) | filepathutil.NormalizePath() |
| workspace/process_executor.go | resolveSymlinks() (10 lines) | filepathutil.NormalizePath() |
| analysis/index_search.go | ToLower(ToSlash(...)) inline | filepathutil.NormalizeKey() |
| analysis/harvest.go | ToSlash + volume strip inline (8 lines) | filepathutil.NormalizeKey() |
| analysistest/mock_security.go | ToSlash + volume strip inline (7 lines) | filepathutil.NormalizeKey() |

### DIFF
--- a/internal/infrastructure/security/paths.go
+++ b/internal/infrastructure/security/paths.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 
 	domain_security "github.com/gosharplite/tell-me-go/internal/domain/security"
+	"github.com/gosharplite/tell-me-go/internal/pkg/filepathutil"
 )
 
 // pathPolicy manages allowed boundaries and validates paths.
@@ -133,7 +134,7 @@ func (p *pathPolicy) ValidatePath(path string, writable bool) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("invalid path: %w", err)
 	}
-	absPath = p.resolveSymlinks(absPath)
+	absPath = filepathutil.NormalizePath(absPath)
 
 	if err := p.isSystemDirectory(absPath); err != nil {
 		return "", err
@@ -172,8 +173,7 @@ func (p *pathPolicy) isSystemDirectory(absPath string) error {
 	// 0. Normalize and resolve input path for internal consistency.
 	// This ensures that short/long names on Windows and symlinks are handled
 	// before prefix matching against system directories and exemptions.
-	absPath = p.resolveSymlinks(absPath)
-	absPath = filepath.ToSlash(absPath)
+	absPath = filepathutil.Normalize(absPath)
 
 	if p.isExemptedDirectory(absPath) {
 		return nil
@@ -322,8 +322,8 @@ func (p *pathPolicy) checkBoundary(target, boundary string) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	realBoundary := p.resolveSymlinks(absBoundary)
-	realTarget := p.resolveSymlinks(target)
+	realBoundary := filepathutil.NormalizePath(absBoundary)
+	realTarget := filepathutil.NormalizePath(target)
 
 	rel, err := filepath.Rel(realBoundary, realTarget)
 	ok := err == nil && !strings.HasPrefix(rel, "..") && !filepath.IsAbs(rel)
@@ -340,16 +340,8 @@ func (p *pathPolicy) tryBoundary(absPath, boundary string) bool {
 	return ok
 }
 
+// resolveSymlinks delegates to filepathutil.NormalizePath for centralized
+// cross-platform symlink resolution with recursive fallback.
 func (p *pathPolicy) resolveSymlinks(path string) string {
-	if realPath, err := filepath.EvalSymlinks(path); err == nil {
-		return realPath
-	}
-
-	dir := filepath.Dir(path)
-	if dir == path || dir == "." {
-		return path
-	}
-
-	resolvedDir := p.resolveSymlinks(dir)
-	return filepath.Join(resolvedDir, filepath.Base(path))
+	return filepathutil.NormalizePath(path)
 }
--- a/internal/pkg/filepathutil/normalize.go
+++ b/internal/pkg/filepathutil/normalize.go
@@ -70,8 +70,14 @@ func NormalizeKey(path string) string {
 // already absolute and on the same filesystem).
 func NormalizeForCompare(path string) string {
 	s := filepath.ToSlash(path)
+	// Strip Windows volume prefix on all platforms.
+	// filepath.VolumeName only detects volumes on the current OS,
+	// so we also check for [A-Za-z]: manually for cross-platform use.
 	if vol := filepath.VolumeName(s); vol != "" {
 		s = s[len(vol):]
+	} else if len(s) >= 2 && s[1] == ':' &&
+		((s[0] >= 'A' && s[0] <= 'Z') || (s[0] >= 'a' && s[0] <= 'z')) {
+		s = s[2:]
 	}
 	return strings.ToLower(s)
 }
--- a/internal/pkg/filepathutil/normalize.go
+++ b/internal/pkg/filepathutil/normalize.go
@@ -70,6 +70,10 @@ func NormalizeKey(path string) string {
 // already absolute and on the same filesystem).
 func NormalizeForCompare(path string) string {
 	s := filepath.ToSlash(path)
+	// On non-Windows platforms, ToSlash does not convert backslashes
+	// because \ is not the OS path separator. Normalize them manually
+	// so that Windows paths (e.g., D:\foo\bar) work cross-platform.
+	s = strings.ReplaceAll(s, "\\", "/")
 	// Strip Windows volume prefix on all platforms.
 	// filepath.VolumeName only detects volumes on the current OS,
 	// so we also check for [A-Za-z]: manually for cross-platform use.
--- a/internal/pkg/filepathutil/normalize.go
+++ b/internal/pkg/filepathutil/normalize.go
@@ -62,3 +62,16 @@ func NormalizeKey(path string) string {
 	}
 	return strings.ToLower(s)
 }
+
+// NormalizeForCompare converts to forward slashes, strips the Windows
+// volume prefix, and lowercases. Unlike NormalizeKey, it does NOT resolve
+// symlinks — suitable when both paths come from the same OS and only
+// format normalization is needed (e.g., comparing two paths that are
+// already absolute and on the same filesystem).
+func NormalizeForCompare(path string) string {
+	s := filepath.ToSlash(path)
+	if vol := filepath.VolumeName(s); vol != "" {
+		s = s[len(vol):]
+	}
+	return strings.ToLower(s)
+}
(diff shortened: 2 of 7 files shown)


gosharplite:
## Review: `refactor: centralize OS path normalization (#1156)`

### Verdict: ⚠️ Changes Requested — one test failure on Linux; otherwise excellent

---

### What Works Well ✅

The `NormalizePath` and `Normalize` replacements are semantically identical to the old code. All affected packages pass `-count=3 -race` on Linux:

| Package | Result |
|---------|--------|
| `internal/pkg/filepathutil` (all 10 subtests) | ✅ PASS |
| `internal/infrastructure/security` | ✅ PASS |
| `internal/tools/workspace` (path tests) | ✅ PASS |
| `internal/tools/analysis/analysistest` | ✅ PASS |
| `internal/tools/analysis` (harvest + deadcode) | ✅ PASS |

The new package is clean, well-documented, and the three-function API (`Normalize`, `NormalizePath`, `NormalizeKey`) covers all use cases.

---

### ❌ Blocking: `TestIsInSearchPath/Win_root` fails on Linux

```
--- FAIL: TestIsInSearchPath/Win_root (0.00s)
    index_search_test.go:70: isInSearchPath("C:/", "C:/a/b.go") = false; want true
```

**Root cause**: `NormalizeKey` calls `Normalize` → `EvalSymlinks`. The old `isInSearchPath` code only did `ToSlash` + `ToLower` — **no filesystem calls**. On Linux, `C:/` is not a real path, so `EvalSymlinks` fails and the recursive fallback misreconstructs it:

```
NormalizeKey("C:/")
 → Normalize("C:/")
   → EvalSymlinks("C:/") → ENOENT
   → Dir("C:/")  → "C:"     ← Unix treats C: as a directory name
   → Base("C:/") → "C:"     ← Same as Dir!
   → Join("C:", "C:") → "C:/C:"  ← doubled component
   → ToLower → "c:/c:"           ← ≠ expected "c:/"
```

On Linux, `filepath.Dir("C:/")` and `filepath.Base("C:/")` both return `"C:"` because the Unix `filepath` package treats `C:` as a regular directory name, not a volume prefix. The reconstruction `Join("C:", "C:")` produces `"C:/C:"` instead of `"C:/"`.

This is not just a test issue — `NormalizeKey` changes the semantics of `isInSearchPath` from pure string manipulation to filesystem-touching path resolution.

### Recommended Fix

Add a symlink-free variant for `isInSearchPath`:

```go
// NormalizeForCompare converts to forward slashes, strips volume prefix,
// and lowercases. Unlike NormalizeKey, it does NOT resolve symlinks —
// suitable when both paths come from the same OS and only format
// normalization is needed.
func NormalizeForCompare(path string) string {
    s := filepath.ToSlash(path)
    if vol := filepath.VolumeName(s); vol != "" {
        s = s[len(vol):]
    }
    return strings.ToLower(s)
}
```

Then in `index_search.go`:

```go
func (idx *indexer) isInSearchPath(targetPath, filePath string) bool {
    targetPath = filepathutil.NormalizeForCompare(targetPath)
    filePath   = filepathutil.NormalizeForCompare(filePath)
    // ... rest unchanged
```

This preserves the exact old behavior while still centralizing the logic. The other `NormalizeKey` call sites (`isInTargetScope`, `mock_security.IsPathSafe`) work with real on-disk paths where `EvalSymlinks` succeeds, so they are fine as-is.

---

### Minor Observation (non-blocking)

`security.pathPolicy.resolveSymlinks` is now a one-line wrapper around `filepathutil.NormalizePath`. Consider removing it entirely in a follow-up and calling `NormalizePath` directly at the two remaining call sites (`ValidatePath` and `checkBoundary`).

gosharplite:
Fix applied. Added NormalizeForCompare - a symlink-free variant that does pure ToSlash + volume strip + ToLower without touching the filesystem. isInSearchPath now uses this instead of NormalizeKey.

NormalizeKey (with EvalSymlinks) remains at the other two call sites (isInTargetScope, mock_security.IsPathSafe) where paths are real on-disk paths and symlink resolution is beneficial.

All 15 filepathutil tests pass, and TestIsInSearchPath/Win_root passes on this Windows machine - should be green on Linux too.

gosharplite:
Nice — `TestIsInSearchPath` passes all 9 subtests now. 

But 3 of the 5 new `NormalizeForCompare` tests fail on Linux:

```
--- FAIL: TestNormalizeForCompare_WindowsRoot (0.00s)
    normalize_test.go:148: NormalizeForCompare("C:/") = "c:/"; want "/"
--- FAIL: TestNormalizeForCompare_StripsVolume (0.00s)
    normalize_test.go:155: NormalizeForCompare = "d:/users/foo/bar.txt"; want "/users/foo/bar.txt"
--- FAIL: TestNormalizeForCompare_WindowsChildPath (0.00s)
    normalize_test.go:162: NormalizeForCompare = "c:/a/b.go"; want "/a/b.go"
```

**Root cause**: `filepath.VolumeName` is platform-dependent. On Linux, `VolumeName("C:/")` returns `""` — it only detects volumes on the current OS. So the volume-stripping block is skipped entirely.

**Fix**: `NormalizeForCompare` needs to detect Windows volumes manually on non-Windows platforms. Add a fallback after the `filepath.VolumeName` check:

```go
func NormalizeForCompare(path string) string {
    s := filepath.ToSlash(path)
    if vol := filepath.VolumeName(s); vol != "" {
        s = s[len(vol):]
    } else if len(s) >= 2 && s[1] == :
        ((s[0] >= A && s[0] <= Z) || (s[0] >= a && s[0] <= z)) {
        // Manual Windows volume detection for cross-platform use
        s = s[2:]
    }
    return strings.ToLower(s)
}
```

This matches `[A-Za-z]:` as a Windows volume on all platforms, not just Windows. Without this, `NormalizeForCompare` works correctly on Windows (where `VolumeName` returns `"C:"`) but not on Linux/macOS.

gosharplite:
## Updated Review (commit `81b0360`)

### ✅ Previously Blocking: `TestIsInSearchPath/Win_root` — FIXED

All 9 subtests pass on Linux now:
```
--- PASS: TestIsInSearchPath (0.00s)
    --- PASS: .../Win_root
    --- PASS: .../Win_child
    --- PASS: .../Win_sibling_rejected
```

---

### ❌ New Blocking: 3 `NormalizeForCompare` tests fail on Linux

`filepath.VolumeName` is **platform-specific** — on Linux/macOS it returns `""` for `C:/` because the OS doesn't recognize Windows volume prefixes. The volume-stripping block is skipped, and the `ToLower` result still contains the drive letter:

| Test | Input | Got (Linux) | Expected |
|------|-------|-------------|----------|
| `TestNormalizeForCompare_WindowsRoot` | `C:/` | `c:/` | `/` |
| `TestNormalizeForCompare_StripsVolume` | `D:\Users\foo\bar.txt` | `d:/users/foo/bar.txt` | `/users/foo/bar.txt` |
| `TestNormalizeForCompare_WindowsChildPath` | `C:\a\b.go` | `c:/a/b.go` | `/a/b.go` |

### Fix

Add a manual `[A-Za-z]:` pattern fallback after the `VolumeName` check in `NormalizeForCompare`:

```go
func NormalizeForCompare(path string) string {
    s := filepath.ToSlash(path)
    // Strip Windows volume prefix on all platforms.
    // filepath.VolumeName only detects volumes on the current OS,
    // so we also check for [A-Za-z]: manually for cross-platform use.
    if vol := filepath.VolumeName(s); vol != "" {
        s = s[len(vol):]
    } else if len(s) >= 2 && s[1] == ':' &&
        ((s[0] >= 'A' && s[0] <= 'Z') || (s[0] >= 'a' && s[0] <= 'z')) {
        s = s[2:]
    }
    return strings.ToLower(s)
}
```

This makes the volume stripping work on all platforms — on Windows, `VolumeName` catches it; on Linux/macOS, the manual `X:` check catches it.

---

### Everything else — ✅

| Package | `-count=3 -race` | 
|---------|-------------------|
| `internal/pkg/filepathutil` (10/15 pass, 3 fail) | ⚠️ |
| `internal/infrastructure/security` | ✅ |
| `internal/tools/workspace` | ✅ |
| `internal/tools/analysis/analysistest` | ✅ |
| `internal/tools/analysis` (harvest + deadcode + IsInSearchPath) | ✅ |
| `make verify-no-test-sleep` | ✅ |

Once the `NormalizeForCompare` volume detection is fixed, this is good to merge.

gosharplite:
Fixed. Added manual [A-Za-z]: pattern as a fallback for filepath.VolumeName which is platform-specific and returns "" on Linux/macOS for Windows paths like C:/.

On Windows: VolumeName("C:/") returns "C:" - stripped by first branch.
On Linux: VolumeName("C:/") returns "" - caught by manual X: check.

All 15 filepathutil tests pass on this Windows machine.

gosharplite:
Nice — `TestNormalizeForCompare_WindowsRoot` passes now. But 2 of 3 backslash tests still fail:

```
--- FAIL: TestNormalizeForCompare_StripsVolume (0.00s)
    NormalizeForCompare = "\users\foo\bar.txt"; want "/users/foo/bar.txt"
--- FAIL: TestNormalizeForCompare_WindowsChildPath (0.00s)
    NormalizeForCompare = "\a\b.go"; want "/a/b.go"
```

The volume (`D:`, `C:`) is stripped correctly now, but the backslashes remain.

**Root cause**: `filepath.ToSlash` only replaces the **OS-specific** path separator. On Linux, the separator is `/`, so `ToSlash("D:\Users\foo")` is a no-op — `\` is not converted. On Windows it works because `\` IS the separator there.

The manual volume check strips `D:` correctly, leaving `\Users\foo\bar.txt`, but the remaining `\` characters are never converted to `/`.

**Fix**: Add an explicit `strings.ReplaceAll` for backslashes after `ToSlash`:

```go
func NormalizeForCompare(path string) string {
    s := filepath.ToSlash(path)
    // On non-Windows platforms, ToSlash does not convert backslashes
    // because \ is not the OS path separator. Normalize them manually
    // so that Windows paths (e.g., D:\foo\bar) work cross-platform.
    s = strings.ReplaceAll(s, "\\", "/")
    if vol := filepath.VolumeName(s); vol != "" {
        s = s[len(vol):]
    } else if len(s) >= 2 && s[1] == ':' &&
        ((s[0] >= 'A' && s[0] <= 'Z') || (s[0] >= 'a' && s[0] <= 'z')) {
        s = s[2:]
    }
    return strings.ToLower(s)
}
```

On Windows, `ToSlash` already converts `\` → `/`, so the `ReplaceAll` is a harmless no-op. On Linux/macOS, it's necessary.

gosharplite:
Fixed. Added explicit strings.ReplaceAll("\\", "/") after ToSlash. On Windows, ToSlash already converts backslashes so ReplaceAll is a harmless no-op. On Linux/macOS where \ is not the OS separator, this ensures Windows paths like D:\foo\bar normalize correctly.

PR #1164, opened 2026-06-29
fix: add Chmod to infra FileSystem, fix AbsError via function injection (#1163)

gosharplite:
Partial fix for #1163.

### Chmod on infra FileSystem
- Added Chmod to infra FileSystem interface, OSFileSystem, plainOSFS, domainFS bridge, and both mock implementations (domain and infra).

### TestHarvestPackageSymbols_AbsError — function injection
- Added resolvePath func field to defaultDeadCodeAnalyzer (default: filepath.Abs)
- isInTargetScope now uses a.resolvePath instead of filepath.Abs directly
- Test updated to inject an error-returning function instead of deleted-CWD technique
- Removed t.Skip on Windows — test now passes on all platforms

## Remaining (5 tests, for future PR)
The other 5 tests need deeper FileSystem injection through production code (Walk, MkdirAll, OpenFile) and will be addressed in a follow-up.

### DIFF
--- a/internal/domain/persistence/mock_fs.go
+++ b/internal/domain/persistence/mock_fs.go
@@ -143,6 +143,10 @@ func (m *mockFileSystem) MkdirAll(ctx context.Context, path string, perm os.File
 	return nil
 }
 
+func (m *mockFileSystem) Chmod(ctx context.Context, name string, mode os.FileMode) error {
+	return nil
+}
+
 func (m *mockFileSystem) Stat(ctx context.Context, name string) (os.FileInfo, error) {
 	m.mu.RLock()
 	defer m.mu.RUnlock()
--- a/internal/infrastructure/persistence/mock_fs_test.go
+++ b/internal/infrastructure/persistence/mock_fs_test.go
@@ -130,6 +130,10 @@ func (m *mockFileSystem) MkdirAll(ctx context.Context, path string, perm os.File
 	return nil
 }
 
+func (m *mockFileSystem) Chmod(ctx context.Context, name string, mode os.FileMode) error {
+	return nil
+}
+
 func (m *mockFileSystem) CreateTemp(ctx context.Context, dir, pattern string) (File, error) {
 	if m.CreateTempFunc != nil {
 		return m.CreateTempFunc(ctx, dir, pattern)
--- a/internal/infrastructure/persistence/os_fs.go
+++ b/internal/infrastructure/persistence/os_fs.go
@@ -26,6 +26,7 @@ type File interface {
 // FileSystem defines the interface for filesystem operations to allow mocking.
 type FileSystem interface {
 	MkdirAll(ctx context.Context, path string, perm os.FileMode) error
+	Chmod(ctx context.Context, name string, mode os.FileMode) error
 	CreateTemp(ctx context.Context, dir, pattern string) (File, error)
 	Rename(ctx context.Context, oldpath, newpath string) error
 	Remove(ctx context.Context, name string) error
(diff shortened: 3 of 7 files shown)


gosharplite:
## Architectural Review — ✅ APPROVE with minor observations

I reviewed the full diff against `dev...fix/1163-vfs-function-injection`, verified the claims against the live codebase, ran `go build ./...`, `go vet`, and the full test suites for all modified packages. All pass.

### ✅ What's correct

**1. Function injection is the right seam pattern.**
The `resolvePath` field on `defaultDeadCodeAnalyzer` follows the existing precedent (`indexer.resolvePath`, `backup.resolvePath`). This is precisely what I recommended in the issue review — reuse the existing function-injection seam for path-resolution cases rather than threading a full VFS. Well executed.

**2. Nil-guard in `harvest.go` is defensive and correct.**
```go
resolvePath := a.resolvePath
if resolvePath == nil {
    resolvePath = filepath.Abs
}
```
The nil guard is technically unreachable through the constructor, but it protects against direct struct-literal construction in tests (a common pattern in this codebase). Minor suggestion: add a one-line comment like `// defense-in-depth: direct struct construction in tests may leave resolvePath nil`.

**3. `t.Parallel()` is now safe.**
No more CWD mutation — the test is deterministic and parallel-safe. Good catch on restoring parallelism.

**4. `Chmod` additions are mechanical and consistent.**
- Infra `FileSystem` interface + `OSFileSystem` (with `fsRetry` wrapper) ✓
- `domainFS` bridge ✓
- Domain mock (`mockFileSystem`) ✓
- Infra mock (`mock_fs_test.go`) ✓
- `plainOSFS` (consistent with existing methods — no `fsRetry`, matching `MkdirAll`) ✓

**5. Import hygiene is correct.**
Only `"runtime"` was removed from `dead_code_test.go`; `"os"` stays because other tests (`TestResolveCrossPackage_NilTypesInfo`, helper `writeModuleFiles`) still use `os.Getwd`/`os.Chdir`/`os.WriteFile`.

---

### [TECHNICAL DEBT] Observations for follow-up PRs

**1. Domain `FileSystem` interface missing `Chmod`.**
The domain mock (`mockFileSystem`) and `domainFS` bridge both have `Chmod`, but `internal/domain/persistence.FileSystem` does not. Go allows structs to have methods beyond the interface, so this compiles and is harmless *now*, but the next PR that needs domain-level `Chmod` (for `Walk`/`MkdirAll`/`OpenFile` injection) must add it to the domain interface. Not a blocker — just calling it out so it doesn't get missed.

**2. `hasTextMatchOutsidePackage` in `dead_code.go:544` still calls `os.ReadFile` directly.**
This is another OS leak in the dead-code path. Not in scope for this PR, but worth tracking for the follow-up that addresses the remaining 5 tests.

**3. `domainFS.Walk` still passes through to `filepath.Walk` directly.**
When the follow-up PR injects `FileSystem` into `complexity.go`/`dependency.go`/`types.go` for `Walk`, the production adapter will need to either truly delegate `Walk` to the inner fs or explicitly document the passthrough. Not an issue now — just linking to the earlier issue comment.

---

### Test results

| Package | Result |
|---|---|
| `./internal/tools/analysis/...` | PASS (21.5s) |
| `./internal/domain/persistence/...` | PASS |
| `./internal/infrastructure/persistence/...` | PASS (3.2s) |
| `go vet ./...` (modified packages) | CLEAN |
| `go build ./...` | CLEAN |

---

### Verdict

**APPROVE.** The PR is a clean partial fix — it addresses exactly one of the six tests using the right architectural seam (function injection), and lays the `Chmod` groundwork for the remaining five. Scope is well-bounded. No regressions. Ship it.

→ The 5 remaining tests (`types_error_test.go` ×3, `refactor_test.go` ×1, `process_executor_security_test.go` ×1) will need deeper `FileSystem` injection through `Walk`/`MkdirAll`/`OpenFile` — but that's correctly deferred to follow-up PRs.

PR #1166, opened 2026-06-29
fix(tests): resolve flaky TestResilience_504Retry hang/failure on Windows

gosharplite:
## Summary

Fixes #1165 — `TestResilience_504Retry` intermittently fails or hangs on Windows.

### [ARCHITECTURAL BLOCKER] stdin deadlock (`tests/e2e/e2e_test.go`)
When `stdin` is `""`, `cmd.Stdin` was not set, causing the child process to inherit the parent's stdin. On Windows, this can block indefinitely waiting for input.  
**Fix:** Always set `cmd.Stdin` — an empty `strings.NewReader("")` provides immediate EOF.

### [TECHNICAL DEBT] Missing HTTP 504 classification coverage (`internal/infrastructure/llm/gemini/gemini_test.go`)
`TestSendChat_ClassifyError_Integration` tested HTTP 503 but not 504. While 504 was correctly classified via `httpStatusToDomain` (status >= 500 → ErrTransient), the gap meant no regression guard.  
**Fix:** Added `sdk_format_504` to `TestClassifyError` and `HTTP 504 → ErrTransient` to `TestSendChat_ClassifyError_Integration`.

### [REFACTOR] Mock server Content-Type (`tests/e2e/resilience_test.go`)
Mock error responses (504, 429) lacked `Content-Type: application/json`, which could cause the genai SDK to parse errors inconsistently.  
**Fix:** Added `w.Header().Set("Content-Type", "application/json")` before writing error body.

## Verification
- All `TestClassifyError` and `TestSendChat_ClassifyError_Integration` test cases pass
- All `TestRecoveryStep` and `TestDefaultRetryPolicy` tests pass
- All `TestResilientClient` tests pass


gosharplite:
## Review — ❌ Changes Requested (1 blocker)

### ✅ Good: stdin deadlock fix (`tests/e2e/e2e_test.go`)

```go
+ } else {
+     cmd.Stdin = strings.NewReader("")
```

Correct. On Windows, a nil `cmd.Stdin` causes the child process to inherit the parent's stdin handle. If the CLI binary ever reads from stdin (e.g., waiting for interactive input), it blocks indefinitely — matching the "hangs until 180s timeout" symptom in the issue. An empty `strings.NewReader("")` returns EOF immediately, preventing the deadlock. This is the likely root cause of the hang.

### ❌ [ARCHITECTURAL BLOCKER] Content-Type header ordering is backwards (`tests/e2e/resilience_test.go`)

```go
w.WriteHeader(http.StatusGatewayTimeout)          // ← sends headers to client
w.Header().Set("Content-Type", "application/json") // ← too late: no-op
```

`http.ResponseWriter.WriteHeader()` sends the HTTP status line and all accumulated headers to the client immediately. Any `Header().Set()` call after `WriteHeader` is silently ignored by Go's `net/http`. The `Content-Type: application/json` header is **never actually transmitted** to the genai SDK.

This means the genai SDK receives a 504 response with **no Content-Type**. The SDK may parse the body differently (e.g., as plain text rather than JSON), which can affect error classification — defeating the purpose of this fix.

**Required fix** — swap the order in both tests (504 and 429):

```go
// BEFORE (wrong):
w.WriteHeader(http.StatusGatewayTimeout)
w.Header().Set("Content-Type", "application/json")

// AFTER (correct):
w.Header().Set("Content-Type", "application/json")
w.WriteHeader(http.StatusGatewayTimeout)
```

Applies at two locations:
- `resilience_test.go` line ~24 (504 test)
- `resilience_test.go` line ~95 (429 test)

### ✅ Good: 504 test coverage (`internal/infrastructure/llm/gemini/gemini_test.go`)

Adding `sdk_format_504` to `TestClassifyError` and `HTTP 504 → ErrTransient` to `TestSendChat_ClassifyError_Integration`. The classification path (`httpStatusToDomain`: status >= 500 → ErrTransient) was already correct; these are pure regression guards. No behavioral change.

### ⚠️ Note: targeting `main` not `dev`

This PR targets `main`. If your merge workflow is feature branches → `dev` → `main`, this should target `dev` instead. If `main` is intentional, ignore this note.

gosharplite:
Good catch — you're absolutely right. `WriteHeader()` immediately flushes headers to the client, so `Header().Set()` after it is silently discarded by `net/http`. The header was never actually transmitted.

Fixed in 5e43819: swapped the order so `Header().Set("Content-Type", "application/json")` precedes `WriteHeader()` at both locations (504 and 429 tests). The genai SDK now actually receives the Content-Type header on error responses.

> **Note on targeting `main`**: intentional — our release workflow merges feature branches directly to `main`.

PR #1210, opened 2026-07-09
perf: reduce analysis test suite from 28s to 9.5s (66% faster)

gosharplite:
## Summary

Reduces `go test ./internal/tools/analysis` from **~28s → ~9.5s** (66% improvement) through four targeted optimizations.

Closes #1209.

### 1. Skip redundant `discoverModulePath` call (`f3573baf`)
Added `knownModulePath` field to `indexer`. When set, `loadPackages` skips the separate `discoverModulePath` → `packages.Load` call. Updated 40+ test call sites across 15 files.

### 2. Shared workspaces for deep tests (`f04d1e3f`)
Replaced 5 standalone workspace creations in `dead_code_deep_test.go` with 2 shared workspaces using the mutex+check pattern. Eliminated ~8s of workspace creation overhead.

### 3. Per-directory locking replaces global `loadMu` (`cb6760c9`)
Replaced the global `var loadMu sync.Mutex` with `var dirLocks sync.Map` keyed by absolute directory. Allows parallel `packages.Load` on different directories while serializing same-directory calls (preserving Windows deadlock safety). Updated all 6 `loadMu` call sites.

### 4. Pre-built indexer state for NoPackages test (`6a44d10`)
`TestFindOrphanedSymbols_NoPackages` no longer creates a real Go workspace — uses a pre-built `&indexer{}` with empty `pkgs` slice.

## Verification

- ✅ All tests pass (`go test -count=1`)
- ✅ Race detector clean (`go test -race -short`)
- ✅ `go build ./...` and `go vet ./...` clean

## Timing

| Phase | Time |
|-------|------|
| Before | 28.0s |
| After Task 1 | ~15.4s |
| After Task 2 | ~14.7s |
| After Task 3 | **9.5s** |
| After Task 4 | **9.5s** |

### DIFF
--- a/internal/tools/analysis/common_test.go
+++ b/internal/tools/analysis/common_test.go
@@ -105,6 +105,7 @@ func getSharedIndexer(tb testing.TB) *indexer {
 		if err != nil {
 			tb.Fatalf("failed to create shared indexer: %v", err)
 		}
+		sharedIdx.knownModulePath = fixtureMod
 		if err := sharedIdx.Refresh(context.Background(), nil); err != nil {
 			tb.Fatalf("failed to refresh shared indexer: %v", err)
 		}
--- a/internal/tools/analysis/context_test.go
+++ b/internal/tools/analysis/context_test.go
@@ -21,6 +21,7 @@ func TestIndexer_Refresh_ContextCancel(t *testing.T) {
 	}
 
 	idx, _ := newIndexer(tmpDir)
+	idx.knownModulePath = "example.com/test"
 
 	// Use a context that cancels very quickly
 	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Nanosecond)
--- a/internal/tools/analysis/dead_code_constructor_propagation_test.go
+++ b/internal/tools/analysis/dead_code_constructor_propagation_test.go
@@ -28,6 +28,7 @@ import (
 	"go/types"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 
@@ -48,6 +49,22 @@ func writeFixture(t *testing.T, tmpDir string, files map[string]string) {
 	}
 }
 
+// readModulePath reads the module path from a go.mod file in tmpDir.
+// Returns "" if go.mod is missing or unreadable.
+func readModulePath(tmpDir string) string {
+	data, err := os.ReadFile(filepath.Join(tmpDir, "go.mod"))
+	if err != nil {
+		return ""
+	}
+	for _, line := range strings.Split(string(data), "\n") {
+		line = strings.TrimSpace(line)
+		if strings.HasPrefix(line, "module ") {
+			return strings.TrimSpace(strings.TrimPrefix(line, "module "))
+		}
+	}
+	return ""
+}
+
 // runAnalyzer is the common harness: build an indexer over tmpDir, run
 // FindOrphanedSymbols, return the resulting report text. Each test then
 // asserts presence/absence of specific symbols in that text.
@@ -55,6 +72,7 @@ func runAnalyzer(t *testing.T, tmpDir string) string {
 	t.Helper()
 	idx, err := newIndexer(tmpDir)
 	require.NoError(t, err)
+	idx.knownModulePath = readModulePath(tmpDir)
 	ctx := context.Background()
 	require.NoError(t, idx.Refresh(ctx, nil))
 
--- a/internal/tools/analysis/dead_code_contracts_test.go
+++ b/internal/tools/analysis/dead_code_contracts_test.go
@@ -159,10 +159,12 @@ func TestIsWellKnownContract_RealStdlib(t *testing.T) {
 		Mode: packages.NeedTypes | packages.NeedTypesInfo | packages.NeedSyntax |
 			packages.NeedDeps | packages.NeedImports | packages.NeedName,
 	}
-	loadMu.Lock()
-	pkgs, err := packages.Load(cfg, pkgPaths...)
-	loadMu.Unlock()
-	require.NoError(t, err)
+	var pkgs []*packages.Package
+	var loadErr error
+	withDirLock(".", func() {
+		pkgs, loadErr = packages.Load(cfg, pkgPaths...)
+	})
+	require.NoError(t, loadErr)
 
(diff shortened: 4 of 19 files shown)


gosharplite:
### Architecture Review

The PR successfully addresses concurrency bottlenecks by replacing a global lock (`loadMu`) with a per-directory lock (`dirLocks`), allowing parallel execution of `packages.Load` for different directories. It also skips redundant module discovery overhead in tests via Dependency Injection (`knownModulePath`). While these changes improve scalability, they introduce state management issues and resource leaks that must be addressed.

#### 🛑 [ARCHITECTURAL BLOCKER] File System Resource Leak in Test State Management
**Current:** 
In `internal/tools/analysis/dead_code_deep_test.go`, shared workspaces (`sharedDeepIdentDir`, `sharedDeepLimDir`) are created via `os.MkdirTemp` inside lazily evaluated package-level helper functions, but they are **never deleted**. This leaks folders onto the OS filesystem with every test run, eventually causing disk exhaustion on CI runners or local machines.
**Scalable:** 
Manage shared test fixture lifecycles properly. Instead of using lazily initialized package-level singletons, leverage `TestMain(m *testing.M)` for global setup/teardown, guaranteeing cleanup via `defer os.RemoveAll()`. Alternatively, utilize `t.TempDir()` (which handles its own cleanup) if the new per-directory lock makes tests fast enough to afford per-test directories.

```go
// Scalable approach using TestMain for shared state
func TestMain(m *testing.M) {
    dir, _ := os.MkdirTemp("", "shared-workspace-*")
    defer os.RemoveAll(dir) // Guarantees cleanup
    
    // Initialize shared workspaces here...
    
    os.Exit(m.Run())
}
```

#### ⚠️ [TECHNICAL DEBT] Unscalable State Management & Memory Leak (`dirLocks`)
**Current:** 
`var dirLocks sync.Map` caches mutexes per directory. However, entries are **never evicted**. In a long-running process (e.g., an LSP daemon) or during test executions where many random temporary directories are evaluated, this creates an unbounded memory leak. 
Furthermore, `dirLocks.LoadOrStore(abs, &sync.Mutex{})` unconditionally allocates a new mutex on the heap for *every* `withDirLock` call, adding unnecessary garbage collection pressure on hot paths.
**Scalable:** 
Avoid package-level global state where possible. If cross-instance synchronization is strictly required for the Windows build cache, you must at least eliminate the unconditional heap allocation. For a truly scalable long-running process, implement an eviction strategy (like an LRU or reference-counted map) to prevent unbounded map growth.

```go
// Current (Allocates on every call)
mu, _ := dirLocks.LoadOrStore(abs, &sync.Mutex{})

// Scalable (Zero allocation on cache hit)
func withDirLock(dir string, fn func()) {
    abs, err := filepath.Abs(dir)
    if err != nil {
        abs = dir
    }
    
    v, ok := dirLocks.Load(abs)
    if !ok {
        v, _ = dirLocks.LoadOrStore(abs, &sync.Mutex{})
    }
    
    mu := v.(*sync.Mutex)
    mu.Lock()
    defer mu.Unlock()
    fn()
}
```

#### 💡 [REFACTOR] Adherence to Separation of Concerns (Test Architecture)
**Current:** 
Tests rely heavily on package-level globals (`sharedDeepIdentMu`, `sharedDeepIdentIdx`, etc.) to share state. This violates the "Avoid Package-Level State" principle, making tests tightly coupled, harder to trace, and more vulnerable to race conditions when utilizing `t.Parallel()`.
**Scalable:** 
Encapsulate shared dependencies in a struct and pass them explicitly.

```go
// Encapsulate test environment
type SharedWorkspace struct {
    Dir     string
    Indexer *indexer
}

// Pass dependencies down explicitly rather than relying on globals
func getSharedDeepIdentIndexer(ws *SharedWorkspace, t *testing.T) {
    t.Helper()
    // ...
}
```

gosharplite:
### Architecture Review (Follow-up)

Thank you for addressing the OS filesystem cleanup and reducing the GC pressure in `withDirLock`. However, there are remaining architectural concerns that were not fully addressed in the latest commit.

#### ⚠️ [TECHNICAL DEBT] Unbounded Map Growth in `dirLocks` remains
**Current:** 
While the recent commit successfully eliminates the unconditional `&sync.Mutex{}` heap allocation on cache hits, the `sync.Map` itself still **never evicts entries**. In a long-running daemon process (e.g., LSP server), this remains an unbounded memory leak as directories are visited.
**Scalable:** 
Implement an eviction strategy. A simple approach is to use a `sync.Pool` of mutexes if cross-directory locking isn't strictly necessary, or implement a reference-counted map where the mutex is deleted from the `sync.Map` when its reference count drops to zero after `packages.Load` completes.

#### 💡 [REFACTOR] Ignored Separation of Concerns (Test Architecture)
**Current:** 
The previous review recommended removing package-level mutable state (`sharedDeepIdentDir`, `sharedDeepIdentMu`, etc.) in `dead_code_deep_test.go` to adhere to the "Avoid Package-Level State" principle. The tests still rely on these globals, tightly coupling the test state to the package and making parallel test execution harder to reason about.
**Scalable:** 
Encapsulate the test workspace dependencies in a struct and pass them explicitly.
```go
// Encapsulate test environment
type SharedWorkspace struct {
    Dir     string
    Indexer *indexer
}
// Initialize in TestMain or a shared sync.Once struct, and pass explicitly.
```

Please address the `[REFACTOR]` item to ensure the test suite remains maintainable and scalable.

gosharplite:
### Architecture Review: Approved ✅

Thank you for the quick follow-up. 

The refactoring to introduce the `sharedWorkspace` struct beautifully resolves the `[REFACTOR]` concern. By encapsulating the directory string, the `*indexer` instance, and the `sync.Mutex` together, the test architecture now properly adheres to Separation of Concerns. This makes the shared state explicit, thread-safe, and self-contained, rather than relying on disjoint, floating package-level variables.

Combined with the previous commits that resolved the `[ARCHITECTURAL BLOCKER]` (disk exhaustion via `TestMain` cleanup) and mitigated the `[TECHNICAL DEBT]` (reducing GC pressure on `dirLocks`), the concurrency improvements to the analysis indexer are now highly scalable and safe to merge.

Great work! 🚀

PR #1212, opened 2026-07-09
perf: reduce e2e test suite from ~17s to ~9.4s (44% faster)

gosharplite:
## Summary

Closes #1211.

Reduces `go test ./tests/e2e/...` from **~16.7s → ~9.4s** (44% faster) by fixing three bottlenecks identified in the investigation.

### 1. Fix `TestStdinPiping` — missing mock server (19.4s → 0.22s)
The test had no `TELL_ME_MOCK_URL`, causing the binary to attempt a real network connection and block on TCP timeout. Added a minimal `httptest.Server` and explicit temp config via `runCommandWithEnvInDir`.

### 2. Replace all `time.Sleep` with polling in history tests
Added `waitForHistoryStable()` helper that polls `os.Stat` at 10ms intervals until the history file size stabilizes. Replaced 6 `time.Sleep` calls across `newHistoryNavEnv`, `TestHistoryNavigationFlags_GoBack`, `TestHistoryNavigationFlags_Retry`, and `TestHistoryOnlyExit`.

### 3. Cache binary build in `TestMain`
Skip `go build` when the binary already exists and is newer than `main.go`, using `ModTime` comparison.

## Per-test impact

| Test | Before | After |
|---|---|---|
| `TestStdinPiping` | 19.4s | 0.22s |
| `TestHistoryNavigationFlags_Retry` | 5.5s | 0.65s |
| `TestHistoryNavigationFlags_GoBack` | 4.5s | 0.56s |
| `TestHistoryNavigationFlags_LastNMessages` | 3.4s | 0.51s |
| `TestHistoryOnlyExit` | 1.3s | 0.37s |

## Remaining

~4.6s is the one-time binary build in `TestMain` (temp dir scoping prevents cross-invocation caching). ~2.8s is `TestBypassArchiving`/`TestSessionArchiving` subprocess spawns for the `--new` archiving path — converting these to in-process DI tests (like `golden_path_test.go`) is a potential follow-up.

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -65,12 +65,26 @@ func TestMain(m *testing.M) {
 	projectRoot = filepath.Dir(filepath.Dir(wd))
 	mainPath := filepath.Join(projectRoot, "cmd", "tell-me-go", "main.go")
 
-	fmt.Printf("Building binary: %s from %s\n", binPath, mainPath)
-	build := exec.Command("go", "build", "-o", binPath, mainPath)
-	if out, err := build.CombinedOutput(); err != nil {
-		fmt.Fprintf(os.Stderr, "Failed to build binary: %v\nOutput: %s\n", err, string(out))
-		_ = os.RemoveAll(tempDir)
-		os.Exit(1)
+	// Skip rebuild if the binary is already up-to-date (faster incremental runs).
+	needsBuild := true
+	if binInfo, err := os.Stat(binPath); err == nil {
+		if srcInfo, err := os.Stat(mainPath); err == nil {
+			if binInfo.ModTime().After(srcInfo.ModTime()) {
+				needsBuild = false
+			}
+		}
+	}
+
+	if needsBuild {
+		fmt.Printf("Building binary: %s from %s\n", binPath, mainPath)
+		build := exec.Command("go", "build", "-o", binPath, mainPath)
+		if out, err := build.CombinedOutput(); err != nil {
+			fmt.Fprintf(os.Stderr, "Failed to build binary: %v\nOutput: %s\n", err, string(out))
+			_ = os.RemoveAll(tempDir)
+			os.Exit(1)
+		}
+	} else {
+		fmt.Printf("Using cached binary: %s\n", binPath)
 	}
 
 	code := m.Run()
@@ -357,12 +371,22 @@ func TestStdinPiping(t *testing.T) {
 		t.Skip("skipping slow E2E test in short mode")
 	}
 
+	// Setup a minimal mock server so the binary does not attempt a real
+	// network connection. The test only needs to verify stdin piping logs.
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	}))
+	defer server.Close()
+
 	homeDir := t.TempDir()
-	env := []string{"TELL_ME_HOME=" + homeDir}
+	configPath := createTempConfig(t, "google", server.URL)
+	env := []string{
+		"TELL_ME_HOME=" + homeDir,
+		"TELL_ME_MOCK_URL=" + server.URL,
+	}
 
 	stdinContent := "This is from stdin"
-	// We check if the stderr shows "Input captured" which is a log in main.go
-	_, stderr, _ := runCommandWithEnv(env, stdinContent, "Prompt from arg")
+	_, stderr, _ := runCommandWithEnvInDir(homeDir, env, stdinContent, "-c", configPath, "Prompt from arg")
 
 	out := stripANSI(stderr)
 	if !strings.Contains(out, "Input captured") {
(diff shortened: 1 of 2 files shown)


gosharplite (commented):
### Architecture Review

Following the Architecture SOP, here is the review of PR #1212:

**1. Analyze**
- **Dependency & Boundaries:** The introduction of per-directory locking (`dirLocks`) inside `index.go` safely mitigates a global bottleneck (`loadMu`) when invoking `packages.Load`. This is a targeted fix that respects the existing module boundaries.
- **God Objects / Global State:** The `dirLocks` variable is a package-level global state (`sync.Map`). While it accurately models the Go build cache contention per directory, it introduces a memory leak risk.

**2. Modularity & 3. Scalability**
- **Test Scalability:** Bypassing the test binary rebuild and injecting `knownModulePath` directly into the `indexer` during tests avoids unnecessary `packages.Load` invocations, greatly improving execution time.
- **Memory Leak in Global Map:** The `var dirLocks sync.Map` grows unbounded. Every new unique directory (e.g., `t.TempDir()` in tests, or varying temporary paths in a long-lived process) adds a new `*sync.Mutex` that is never evicted. In a long-lived service, this constitutes an unscalable state management pattern.

**4. Consistency**
- **Test Stability:** Replacing static `time.Sleep` with `waitForHistoryStable` in `history_flags_test.go` perfectly aligns with robust, non-flaky test practices.

---

### Findings & Feedback

**[TECHNICAL DEBT]** - Unbounded Global State (`sync.Map` memory leak)
*Impact:* The `dirLocks` map stores a pointer to a `sync.Mutex` for every unique directory path ever loaded. Because keys are never removed, long-running processes (or extensive test suites) will slowly leak memory over time.
*Current Pattern:*
```go
var dirLocks sync.Map // map[string]*sync.Mutex

func withDirLock(dir string, fn func()) {
	// ...
	v, _ := dirLocks.LoadOrStore(abs, &sync.Mutex{})
	// Lock is used but the entry remains in the map forever.
}
```
*Scalable Pattern:*
Consider whether a global lock is truly necessary. If an `indexer` instance has exclusive ownership of its `dir` in production, you can replace the global map with a simple `mu sync.Mutex` embedded inside the `indexer` struct. If inter-instance concurrency for the *same* directory is a strict requirement, use a lock cache with eviction (e.g., removing the lock from the map when the lock count reaches 0) or an LRU cache for locks.

**[REFACTOR]** - Test-Only Field in Domain Struct
*Impact:* `knownModulePath` is added to `indexer` strictly to bypass `discoverModulePath` during tests. 
*Current Pattern:* Polluting production structs with test-specific flags.
*Scalable Pattern:* Use Functional Options or interface injection to provide context without hardcoding test behavior into the production type. However, given the immediate performance win, this is acceptable but should be noted for future cleanup.

**Conclusion:**
The performance improvements are excellent and correct. Please address or acknowledge the `sync.Map` memory leak for long-lived processes.

gosharplite:
Thanks for the review.

The feedback on `waitForHistoryStable` and the overall performance direction is appreciated.

Regarding the `[TECHNICAL DEBT]` (`sync.Map` / `dirLocks` in `index.go`) and `[REFACTOR]` (`knownModulePath` in domain struct) — these don't appear in this PR's diff, which only touches `tests/e2e/e2e_test.go` and `tests/e2e/history_flags_test.go`. It looks like those comments may have been intended for a different changeset.

gosharplite (commented):
My sincere apologies! You are absolutely right—I reviewed the entire local branch history instead of isolating the diff for this specific PR.

Let me provide the correct architectural review for **PR #1212**, which focuses on the E2E test suite optimizations:

### Architecture Review (Revised)

**1. Scalability (Build Process & Tests)**
- **Incremental Binary Compilation:** Modifying `TestMain` to conditionally rebuild the CLI binary (`needsBuild` check) is a highly effective optimization for local development. By leveraging file modification timestamps (`ModTime`), the test suite elegantly avoids the heavy `go build` overhead when the source hasn't changed.
- **Mock Server for E2E Tests:** In `TestStdinPiping`, instantiating a local `httptest.NewServer` and passing its URL via `TELL_ME_MOCK_URL` properly isolates the test from the network. This prevents network latency and flakiness from polluting test metrics.

**2. Consistency & Determinism**
- **Eliminating Static Sleeps:** Replacing fixed `time.Sleep(1000 * time.Millisecond)` calls with the `waitForHistoryStable` polling mechanism is a best practice. It transforms a fragile, time-bound assertion into a robust state-bound assertion. The implementation gracefully handles OS-level I/O delays (especially relevant on Windows) without unnecessarily slowing down the test execution.

---

### Findings & Feedback

**[APPROVED] - Clean Architecture for Testing**
*Impact:* The PR successfully reduces E2E execution time by ~44% while increasing the determinism of file-based state checks. 
*Design Check:* The changes strictly adhere to testing best practices (Mocking I/O dependencies, State-based polling over sleep). No architectural blockers or technical debt are introduced.

Excellent work on speeding up the CI and local developer loop!

PR #1219, opened 2026-07-10
fix: loop detector breaks tool_calls→tool_results API contract

thptcnec:
## Bug

Infinite-loop detector corrupts conversation history when the repeating response contains `tool_calls`. The assistant message (with `tool_calls`) is persisted to history, then a `user`-role LoopWarning is injected — but no tool-result messages exist between them. The LLM API rejects the next request with HTTP 400.

## Fix

Skip persisting the response on the loop-break path when it contains `FunctionCall` parts. Only the LoopWarning is appended. Text-only responses are still persisted as before.

## Changes (4 files, +138 / −12)

| File | Change |
|------|--------|
| `internal/agent/orchestrator/middleware.go` | Guard `AddContent` with `!hasToolCalls` check |
| `internal/agent/orchestrator/middleware_test.go` | New unit test: tool-call vs text-only loop-break paths |
| `tests/integration/agent/orchestrator/turn_engine_loop_test.go` | Fix 2 assertions: search all `Parts`, not just `Parts[0]` |
| `tests/integration/agent/orchestrator/turn_engine_test.go` | Fix `assertLoopWarningInjected`: search all `Parts` |

## Verification

- `make check-full` — all gates pass (lint, architecture, vulncheck, tests, race detection)
- Zero coverage gaps in changed files
- `handleLoopBreak` complexity: 5 (acceptable)

### DIFF
--- a/internal/agent/orchestrator/middleware.go
+++ b/internal/agent/orchestrator/middleware.go
@@ -154,7 +154,13 @@ func withLoopDetector() turnMiddleware {
 
 func detectLoop(state *TurnState) bool {
 	// 1. Multi-step loop detection (Text & Tool Calls)
-	rawJSON, _ := json.Marshal(state.Response)
+	// Exclude mutable fields (ID) from the hash to ensure consistent
+	// comparison across turns. AddContent mutates content.ID in-place,
+	// so hashing the raw response would produce different hashes for
+	// structurally identical responses on different turns.
+	sanitized := *state.Response
+	sanitized.ID = ""
+	rawJSON, _ := json.Marshal(&sanitized)
 	h := sha256.Sum256(rawJSON)
 	currentHash := hex.EncodeToString(h[:])
 
@@ -190,34 +196,59 @@ func handleLoopBreak(ctx context.Context, Turn *Turn) (ProcessResult, error) {
 		Level:   "warn",
 	})
 
-	// Only persist the repeating response if it does NOT contain tool calls.
-	// Persisting tool_calls without corresponding tool results violates the
-	// LLM API contract (every assistant tool_calls message must be immediately
-	// followed by matching tool-result messages). Tools are never executed on
-	// the loop-break path, so we skip persistence when tool calls are present.
-	hasToolCalls := (&InferenceStep{}).HasToolCalls(Turn.State.Response)
-	if Turn.State.Response != nil && !hasToolCalls {
-		if err := Turn.CtxManager.AddContent(ctx, Turn.State.Response); err != nil {
-			return ProcessResult{}, err
-		}
+	// Persist the repeating response so the model sees its own mistake in history.
+	if err := Turn.CtxManager.AddContent(ctx, Turn.State.Response); err != nil {
+		return ProcessResult{}, err
 	}
 
-	// INTERCEPTABLE: Append the synthetic warning to history.
-	// We use 'user' role as it's the most common way to provide feedback to the LLM.
+	if err := injectSyntheticLoopFeedback(ctx, Turn); err != nil {
+		return ProcessResult{}, err
+	}
+
+	// RECOVERY: End the current Turn gracefully and signal the Run() loop
+	// to continue to a new generation Turn.
+	Turn.State.Response = nil
+	Turn.State.ToolResponse = nil
+	Turn.State.HasToolCalls = true
+
+	return ProcessResult{NextPhase: PhaseComplete}, nil
+}
+
+// injectSyntheticLoopFeedback appends corrective feedback to history
+// that satisfies the LLM API contract after a loop is detected.
+//
+// For tool-call loops: synthetic "tool"-role FunctionResponse entries
+// are paired with each FunctionCall, satisfying the tool_calls→tool_results
+// adjacency requirement while preserving the model's chain-of-thought text.
+//
+// For text-only loops: a "user"-role warning is appended as before.
+func injectSyntheticLoopFeedback(ctx context.Context, Turn *Turn) error {
+	if Turn.State.HasToolCalls && Turn.State.Response != nil {
+		for _, part := range Turn.State.Response.Parts {
+			if part.FunctionCall != nil {
+				synthetic := &llm.Content{
+					Role: "tool",
+					Parts: []*llm.Part{{
+						FunctionResponse: &llm.FunctionResponse{
+							ID:       part.FunctionCall.ID,
+							Name:     part.FunctionCall.Name,
+							Response: map[string]interface{}{"error": LoopWarning},
+						},
+					}},
+				}
+				if err := Turn.CtxManager.AddContent(ctx, synthetic); err != nil {
+					return err
+				}
+			}
+		}
+		return nil
+	}
+
 	warning := &llm.Content{
 		Role:  "user",
 		Parts: []*llm.Part{{Text: LoopWarning}},
 	}
-	if err := Turn.CtxManager.AddContent(ctx, warning); err != nil {
-		return ProcessResult{}, err
-	}
-
-	// RECOVERY: End the current Turn gracefully and signal the Run() loop to continue to a new generation Turn.
-	Turn.State.Response = nil
-	Turn.State.ToolResponse = nil
-	Turn.State.HasToolCalls = true // Trick shouldStopRunning into starting a new Turn
-
-	return ProcessResult{NextPhase: PhaseComplete}, nil
+	return Turn.CtxManager.AddContent(ctx, warning)
 }
 
 func isDuplicateResponse(currentHash string, recentHashes []string) bool {
--- a/internal/agent/orchestrator/middleware.go
+++ b/internal/agent/orchestrator/middleware.go
@@ -160,7 +160,10 @@ func detectLoop(state *TurnState) bool {
 	// structurally identical responses on different turns.
 	sanitized := *state.Response
 	sanitized.ID = ""
-	rawJSON, _ := json.Marshal(&sanitized)
+	rawJSON, err := json.Marshal(&sanitized)
+	if err != nil {
+		return false // skip loop detection if response can't be serialized
+	}
 	h := sha256.Sum256(rawJSON)
 	currentHash := hex.EncodeToString(h[:])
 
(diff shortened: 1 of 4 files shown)


thptcnec:
Thanks for tackling this critical LLM API contract violation. While the current approach successfully prevents the HTTP 400 crash, it introduces a couple of architectural regressions that need to be addressed before we can merge.

### 1. [ARCHITECTURAL BLOCKER] Cross-Component Coupling & State Management
In `handleLoopBreak`, the PR inline-instantiates `InferenceStep` just to access a utility method. This creates unnecessary coupling between the middleware and a specific pipeline step, and ignores the engine's existing state management.

**Why:** The `InferenceStep` already accurately evaluates and populates `Turn.State.HasToolCalls` *before* the loop detection middleware evaluates the state. 

**Current vs. Scalable:**
```go
// Current (Code Smell)
hasToolCalls := (&InferenceStep{}).HasToolCalls(Turn.State.Response)
if Turn.State.Response != nil && !hasToolCalls {
    // ...
}

// Scalable
if Turn.State.Response != nil && !Turn.State.HasToolCalls {
    // ...
}
```
*(Note: Making this change will require updating the mock `TurnState` in `TestHandleLoopBreak_ToolCallResponseNotPersisted` to explicitly set `HasToolCalls: true`.)*

### 2. [TECHNICAL DEBT] History Corruption via Message Dropping
By selectively dropping the model's response when it contains `tool_calls`, we are corrupting the conversation history. We lose any parallel text reasoning the model generated (e.g., chain-of-thought preceding the tool call), and the subsequently injected `user` loop warning becomes disconnected from the causal action that triggered it.

**Why:** We must satisfy the strict LLM API contract (every `tool_calls` needs a matching `tool` result), but dropping the `tool_calls` entirely to achieve this is an anti-pattern. 

**Scalable Solution:**
Instead of dropping the response and appending a `user` warning, **persist the full model response** (preserving its mistake) and inject a **`tool`** response containing the warning. This natively satisfies the LLM API contract while keeping the history intact.

```go
// 1. ALWAYS persist the repeating response (revert the new guard)
if err := Turn.CtxManager.AddContent(ctx, Turn.State.Response); err != nil {
    return ProcessResult{}, err
}

// 2. Inject a mock 'tool' response instead of a 'user' warning
warning := &llm.Content{
    Role:  "tool",
    Parts: []*llm.Part{{
        FunctionResponse: &llm.FunctionResponse{
            Name:     "system_feedback", // Or extract the attempted tool name
            Response: map[string]interface{}{"error": LoopWarning},
        },
    }},
}
```

**Next Steps:**
Please refactor `handleLoopBreak` to use the existing `Turn.State.HasToolCalls` property, and explore the `tool` role injection instead of dropping the model's output. Let me know if you want to pair on the `tool` response injection logic.

thptcnec:
## Updates addressing review feedback

### 1. [ARCHITECTURAL BLOCKER] Cross-Component Coupling
✅ Resolved. Now uses `Turn.State.HasToolCalls` (already populated by `InferenceStep.updateState`) instead of inline `(&InferenceStep{}).HasToolCalls()`.

### 2. [TECHNICAL DEBT] History Corruption via Message Dropping
✅ Resolved. New design:
- **Always** persists the model response (preserves chain-of-thought text)
- For tool-call loops: injects synthetic `tool`-role `FunctionResponse` entries (one per `FunctionCall`) with matching `ID`/`Name` and `Response: {"error": LoopWarning}`
- For text-only loops: injects `user`-role warning (unchanged)
- Satisfies `tool_calls→tool_results` API contract natively

### Additional fixes discovered during implementation

| # | Bug | Fix |
|---|-----|-----|
| 3 | `detectLoop` false negatives: `AddContent` mutates `content.ID` in-place, producing different hashes for identical responses across turns | Sanitize ID before hashing |
| 4 | `toolResponseCleaner` strips `FunctionCall` parts with empty `ID` | Added `llm.NewID()` to integration test fixtures |

### Complexity & Coverage

| Metric | Before | After |
|--------|--------|-------|
| `handleLoopBreak` complexity | 8 | **3** |
| `injectSyntheticLoopFeedback` (new, standalone) | — | **6** |
| Coverage gaps in changed code | 1 HIGH | **0** |
| `make check-full` (incl. -race) | — | ✅ all gates green |

thptcnec:
Excellent work on the refactor. I have reviewed the updated commits against the Architecture SOP and can confirm that all previous blockers have been resolved:

1. **[Cross-Component Coupling] - Resolved**: Removing the inline `InferenceStep` instantiation and utilizing the pre-evaluated `Turn.State.HasToolCalls` cleanly respects the separation of concerns between the middleware and execution layers.
2. **[History Corruption] - Resolved**: The new `injectSyntheticLoopFeedback` logic is the exact scalable pattern we needed. By preserving the model's response and dynamically synthesizing `tool`-role `FunctionResponse` entries that map perfectly to the attempted `ID` and `Name`, we natively satisfy the strict LLM API contract while fully preserving the model's chain-of-thought text.

The additional fixes (sanitizing the mutable `ID` for deterministic hashing in `detectLoop` and updating the integration test fixtures) are also solid improvements for stability.

The PR now passes all architectural checks. Ready to merge.

thptcnec:
## Architectural Review of PR 1219

Following the Architecture SOP, I have analyzed the current state of PR #1219. While cross-component coupling has been resolved, critical structural and idiomatic issues remain that must be addressed before merging.

### 1. [ARCHITECTURAL BLOCKER] LLM API Contract Violation (Tool Result Grouping)
**Analysis:** In `injectSyntheticLoopFeedback`, the code loops over `Turn.State.Response.Parts` and calls `AddContent` for *each* `FunctionCall`. This creates multiple sequential `tool` messages (N messages for N tool calls). Strict LLM providers (e.g., Anthropic, Gemini) require that all tool results corresponding to a single turn's tool calls be batched into a *single* `tool` message containing multiple parts. Sending sequential `tool` role messages breaks role alternation rules and will trigger HTTP 400s.

**Current vs. Scalable:**
*Current (N+1 State Modification):*
```go
for _, part := range Turn.State.Response.Parts {
    if part.FunctionCall != nil {
        synthetic := &llm.Content{ Role: "tool", Parts: /* 1 part */ }
        Turn.CtxManager.AddContent(ctx, synthetic) // Called N times
    }
}
```

*Scalable (Atomic Batch Injection):*
```go
var syntheticParts []*llm.Part
for _, part := range Turn.State.Response.Parts {
    if part.FunctionCall != nil {
        syntheticParts = append(syntheticParts, &llm.Part{
            FunctionResponse: &llm.FunctionResponse{
                ID:       part.FunctionCall.ID,
                Name:     part.FunctionCall.Name,
                Response: map[string]interface{}{"error": LoopWarning},
            },
        })
    }
}
if len(syntheticParts) > 0 {
    return Turn.CtxManager.AddContent(ctx, &llm.Content{
        Role:  "tool",
        Parts: syntheticParts,
    })
}
```

### 2. [TECHNICAL DEBT] Silent Failure on JSON Marshal
**Analysis:** In `detectLoop`, `json.Marshal(&sanitized)` explicitly swallows the error with the blank identifier (`_`). If marshalling fails (e.g., due to unsupported types like channels/funcs injected dynamically in the underlying maps), it yields a `nil` byte slice. This results in a static SHA-256 hash for all failed marshals, causing catastrophic false-positive loop detections.

**Current vs. Scalable:**
*Current:*
```go
rawJSON, _ := json.Marshal(&sanitized)
h := sha256.Sum256(rawJSON)
```

*Scalable (Idiomatic Go):*
```go
rawJSON, err := json.Marshal(&sanitized)
if err != nil {
    // Log the error and fail safely (skip loop detection for this turn)
    return false 
}
h := sha256.Sum256(rawJSON)
```

### 3. [REFACTOR] Redundant Nil Checks vs. Potential Panics
**Analysis:** 
- `detectLoop` performs a blind dereference: `sanitized := *state.Response`. If `state.Response` is ever `nil`, this panics. 
- Conversely, `injectSyntheticLoopFeedback` defensively checks `if Turn.State.HasToolCalls && Turn.State.Response != nil`. 

**Actionable:** Ensure `state.Response != nil` is strictly verified *before* the dereference in `detectLoop` logic. The redundant `nil` check in `injectSyntheticLoopFeedback` can then be removed since the upstream invariant is guaranteed.

thptcnec:
## Updates addressing third review

### 1. [ARCHITECTURAL BLOCKER] Tool Result Grouping — ✅ Resolved
`injectSyntheticLoopFeedback` now collects all `FunctionResponse` parts into a single `tool`-role `AddContent` call instead of N sequential calls. This:
- Satisfies strict LLM role alternation rules (single tool message per turn)
- Eliminates fragile reliance on `AddContent` same-role merge side-effect
- Is more efficient (1 call instead of N)

### 2. [TECHNICAL DEBT] Silent `json.Marshal` Error — ✅ Resolved
`detectLoop` now checks the `json.Marshal` error and returns `false` on failure instead of silently hashing `nil`, which would cause false-positive loop detections for all unsupported types.

### Pipeline
`make check-full` — all gates green including `-race`.

thptcnec:
## Final Architectural Review

I have reviewed the updated commits against the Architecture SOP. Excellent work resolving the issues.

### 1. Tool Result Grouping - ✅ Resolved
The refactoring of `injectSyntheticLoopFeedback` perfectly addresses the LLM API contract violation. By collecting all `FunctionResponse` parts and injecting them atomically in a single `tool`-role message, we ensure strict compliance with role alternation rules (like Anthropic/Gemini) and avoid relying on unpredictable same-role merge side-effects.

### 2. Silent JSON Marshal Error - ✅ Resolved
`detectLoop` now gracefully handles the `json.Marshal` error. Failing safely (returning `false`) prevents the catastrophic false-positive loop detections that would have occurred when unsupported types (like dynamically injected channels/functions) caused `json.Marshal` to yield a `nil` byte slice.

### 3. Redundant Nil Checks vs. Potential Panics - ✅ Structurally Safe
While the redundant `nil` check remains in `injectSyntheticLoopFeedback`, tracing the middleware execution path confirms that `withLoopDetector()` explicitly checks `Turn.State.Response == nil` *before* ever calling `detectLoop(Turn.State)`. Therefore, the blind dereference `sanitized := *state.Response` in `detectLoop` is structurally protected from panics.

The PR now fully aligns with our scalability and API compliance standards. Great work. Ready to merge.

thptcnec:
### Architectural Review: Approved ✅

**Analysis & Modularity**
This PR elegantly resolves a complex edge-case regarding strict LLM role-adjacency (where  responses must immediately follow  tool calls). 

* **Clean Boundaries:** Extracting  ensures the orchestrator handles the LLM API constraints locally without leaking transport-specific rules into the core context manager.
* **Deterministic Hashing:** The shallow copy approach () to strip the mutable uid=504(johndoe) gid=20(staff) groups=20(staff),12(everyone),61(localaccounts),79(_appserverusr),80(admin),81(_appserveradm),98(_lpadmin),702(com.apple.sharepoint.group.2),704(com.apple.sharepoint.group.4),701(com.apple.sharepoint.group.1),703(com.apple.sharepoint.group.3),33(_appstore),100(_lpoperator),204(_developer),250(_analyticsusers),395(com.apple.access_ftp),398(com.apple.access_screensharing),399(com.apple.access_ssh),400(com.apple.access_remote_ae),705(com.apple.sharepoint.group.5) before JSON marshaling is a very clean, modular way to achieve stable loop-detection hashes.
* **Robustness:** Adding the serialization guard () ensures non-serializable payloads fail open rather than crashing the Turn Engine.

**Feedback / Minor Refactor**
**[REFACTOR] (Test Maintainability)**
I noticed that the verification loop to find the injected warning logic is duplicated across  and  in three separate places:
```go
for _, msg := range window {
    if msg.Role == "user" { ... }
    if msg.Role == "tool" { ... }
}
```
*Recommendation:* Extract this duplicated assertion block into a shared test helper function (e.g., `assertContainsLoopWarning(t *testing.T, history []*llm.Content)`) utilizing `t.Helper()`. This aligns with our Go Testing Patterns SOP and makes it easier to update if the warning structure ever changes in the future.

Overall, excellent work satisfying the model provider constraints cleanly. Safe to merge!

thptcnec:
### Final Review: Approved ✅

The requested refactoring of the test assertions has been successfully implemented. 

* The new `assertLoopWarningInHistory` helper greatly improves the maintainability of the test suite. 
* Properly utilizing `t.Helper()` adheres perfectly to standard Go testing best practices outlined in our SOP.
* The test logic is now DRY and will be much easier to manage if the synthetic warning structure evolves.

Great work! 🚀

PR #1223, opened 2026-07-10
feat: add -o/--tui-output flag for progress TUI dashboard (#1216)

thptcnec:
## Summary

Adds a `-o` / `--tui-output` flag that replaces the linear log output with a minimal two-line Bubble Tea TUI dashboard during agent turns.

```
╭─ Turn 1 - architect-johndoe
[07:38:58] Payload: ~123906/1000000 tokens - architect-johndoe - deepseek-v4-pro
```

## Changes

| Package | Files | Change |
|---|---|---|
| `internal/ui/tui/progress/` | `model.go`, `model_test.go` | New: Bubble Tea model with `stateIdle → stateThinking → stateRendering` state machine, channel-based event delivery, and 7 Update tests |
| `internal/cli/` | `chat_command.go` | New `-o`/`--tui-output` flag, threaded into `ChatCommand` |
| `internal/agent/` | `chat.go`, `service.go` | `TUIOutput` field threaded through `ChatCommand` → `RunParams` |
| `internal/agent/session/` | `session_manager.go`, `interfaces.go`, `export_test.go` | `setupUIRendering` branch: TUI path creates channel, subscribes, launches `tea.Program`; `tuiCleanup` closes channel after agent shutdown for clean terminal exit |

## Key design decisions

- **Channel-based event delivery**: Domain events flow into model via buffered channel (cap 64, non-blocking send with drop)
- **Shutdown**: `chatAgent.Shutdown` → `tuiCleanup()` closes channel → `waitForEvent` returns `tea.Quit()` → terminal restored
- **Ctrl+C**: Handled in `Update`, returns `tea.Quit` directly
- **Composes with `-i`**: Both TUI prompt and progress dashboard work together

## Testing

- 7 unit tests for `Update` state transitions and commands
- End-to-end verified: clean terminal exit with cursor restored
- `go build ./...`, `go vet ./...` clean

PR #1224, opened 2026-07-11
feat: AI response display, markdown rendering, and turn metrics in progress TUI (#1216)

thptcnec:
## Summary

Extends the `-o` progress TUI dashboard to display AI response text, rendered through glamour markdown, plus post-call token/metrics lines matching the default linear renderer exactly.

## Changes

| Commit | What |
|---|---|
| `0b8d6cd` | Response text display — `ResponseEvent` text stored and rendered below header |
| `075d048` | Markdown rendering — glamour `TermRenderer` injected via `mdRender func(string)string` |
| `23f3666` | Post-call metrics + final summary — `IsPostCall`/`IsFinal` branches with `formatMetricsLine` and `formatFinalLine` |
| `e3a6d20` | Formatting parity — ⠿ prefix, exact token count, session timing, 4-cost format |
| `3bb8e56` | More parity — timestamp on metrics line, per-turn average, single closing paren, separator |
| `fdc288f` | Σ character fix |

## Output parity with default renderer

```
╭─ Turn 3 - architect-johndoe
[08:43:10] Payload: ~192657/1000000 tokens - architect-johndoe - deepseek-v4-pro

  Pushed. 5 commits:                                                          
  ... (markdown-rendered response)

[08:43:10] Payload: 192657/1000000 tokens - architect-johndoe - deepseek-v4-pro
[08:43:10] [deepseek-pro] M: 145 H: 192512 C: 109 ($0.0009) [2.92s (ΣT: 0.00s)] / 11.84s (3.95)
╰─⠿ Ready ($0.0009 $0.0025 $0.1781 $4.6068 M: 131333 H: 18503680 99.3% O: 61917)
```

## Architecture

- `mdRender func(string)string` injected from `renderer.go` — no glamour in `model.go`
- `postCallStatus *events.TurnStatus` stored for metrics line
- `formatMetricsLine` / `formatFinalLine` helpers for clean View composition

## Testing

22 tests across 3 layers (Update, View, Integration) — all pass.

### DIFF
--- a/internal/ui/tui/progress/model.go
+++ b/internal/ui/tui/progress/model.go
@@ -33,8 +33,10 @@ type model struct {
 	maxTokens    int
 	timestamp    time.Time
 	err          error
-	responseText string // accumulated AI response text
-	mdRender     func(string) string // optional markdown renderer
+	responseText    string             // accumulated AI response text
+	mdRender        func(string) string // optional markdown renderer
+	postCallMetrics *llm.Metrics       // non-nil when TurnStatusEvent has IsPostCall
+	finalCostLine   string             // rendered "Ready (...)" line from IsFinal
 }
 
 // NewModel creates a new progress model that consumes events from the given
@@ -89,6 +91,13 @@ func (m *model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		m.sessionName = msg.Status.Mode
 		m.modelName = msg.Status.Model
 		m.currentState = stateRendering
+
+		if msg.Status.IsPostCall && msg.Status.Metrics != nil {
+			m.postCallMetrics = msg.Status.Metrics
+		}
+		if msg.Status.IsFinal {
+			m.finalCostLine = formatFinalLine(msg.Status)
+		}
 		return m, m.waitForEvent()
 
 	case events.ResponseEvent:
@@ -122,6 +131,63 @@ func extractResponseText(content *llm.Content) string {
 	return sb.String()
 }
 
+// formatMetricsLine renders a single-line post-call metrics summary.
+func formatMetricsLine(m *llm.Metrics) string {
+	if m == nil {
+		return ""
+	}
+	miss := m.PromptTokens - m.CachedTokens
+	var parts []string
+
+	display := m.Provider
+	if display == "" {
+		display = m.Model
+	}
+	if display != "" {
+		parts = append(parts, fmt.Sprintf("[%s]", display))
+	}
+
+	parts = append(parts, fmt.Sprintf("M: %d H: %d C: %d", miss, m.CachedTokens, m.ResponseTokens))
+
+	if m.ThinkingTokens > 0 {
+		parts = append(parts, fmt.Sprintf("Th: %d", m.ThinkingTokens))
+	}
+
+	if m.Cost > 0 {
+		parts = append(parts, fmt.Sprintf("($%.4f)", m.Cost))
+	}
+
+	totalLatency := m.Duration + m.ToolDuration
+	parts = append(parts, fmt.Sprintf("[%.2fs (∑T: %.2fs)]", totalLatency, m.CumulativeToolDuration))
+
+	return strings.Join(parts, " ")
+}
+
+// formatFinalLine renders the "Ready" summary line when IsFinal is true.
+func formatFinalLine(status events.TurnStatus) string {
+	var parts []string
+	parts = append(parts, "Ready")
+
+	if status.TaskCost > 0 {
+		parts = append(parts, fmt.Sprintf("($%.4f)", status.TaskCost))
+	}
+	if status.SessionCost > 0 {
+		parts = append(parts, fmt.Sprintf("$%.4f", status.SessionCost))
+	}
+	if status.DailyCost > 0 {
+		parts = append(parts, fmt.Sprintf("$%.4f", status.DailyCost))
+	}
+
+	hitRate := 0.0
+	if total := status.TotalM + status.TotalH; total > 0 {
+		hitRate = float64(status.TotalH) / float64(total) * 100
+	}
+	parts = append(parts, fmt.Sprintf("M: %d H: %d %.1f%% O: %d",
+		status.TotalM, status.TotalH, hitRate, status.TotalO))
+
+	return "╰─ " + strings.Join(parts, " ")
+}
+
 // View renders the progress model as a two-line display with optional response text.
 func (m *model) View() string {
 	ts := m.timestamp.Format("15:04:05")
@@ -131,6 +197,12 @@ func (m *model) View() string {
 	out := header + "\n" + info
 	if m.responseText != "" {
 		out += "\n" + m.responseText
+	}
+	if m.postCallMetrics != nil {
+		out += "\n" + formatMetricsLine(m.postCallMetrics)
+	}
+	if m.finalCostLine != "" {
+		out += "\n" + m.finalCostLine
 	}
 	return out + "\n"
 }
--- a/internal/ui/tui/progress/model.go
+++ b/internal/ui/tui/progress/model.go
@@ -167,7 +167,7 @@ func formatMetricsLine(m *llm.Metrics, startTime time.Time, timestamp time.Time,
 	}
 
 	totalLatency := m.Duration + m.ToolDuration
-	timing := fmt.Sprintf("[%.2fs (∑T: %.2fs)]", totalLatency, m.CumulativeToolDuration)
+	timing := fmt.Sprintf("[%.2fs (ΣT: %.2fs)]", totalLatency, m.CumulativeToolDuration)
 	if !startTime.IsZero() {
 		sessionDur := time.Since(startTime).Seconds()
 		if turns > 0 {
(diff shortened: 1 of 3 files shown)


thptcnec:
### Architectural Review

This PR successfully implements markdown rendering while honoring Clean Architecture by injecting `glamour` via the `mdRender` dependency rather than coupling it directly into the domain models. However, there are critical state management and event routing issues that violate the established architectural patterns from PR #1223.

#### 1. [ARCHITECTURAL BLOCKER] Goroutine Leak & State Machine Corruption
**File:** `internal/ui/tui/progress/model.go`
**Issue:** The `Update()` loop falls back to `return m, m.waitForEvent()` in the `default:` switch block and for unhandled keys/errors. Bubble Tea executes every returned `tea.Cmd` in a new background goroutine. When native framework messages arrive (e.g., `tea.WindowSizeMsg`, mouse events, ignored keystrokes), the model spawns an *additional* concurrent reader on `m.eventCh`. 

With multiple concurrent readers on a single channel, Go's scheduler distributes incoming events randomly across goroutines. This breaks the sequential delivery of domain events, allowing a `TurnStatusEvent` to hit the Bubble Tea loop before its preceding `TurnStarted`, corrupting the state machine.
**Directive:** Wrap domain events in a distinct type to ensure exactly one reader exists at any given time.
```go
// 1. Wrap the domain event
type domainEventMsg events.Event 

func (m *model) waitForEvent() tea.Cmd {
    return func() tea.Msg {
        e, ok := <-m.eventCh
        if !ok { return tea.Quit() }
        return domainEventMsg(e)
    }
}

// 2. Consume and loop ONLY on your explicit wrapper
func (m *model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
    switch msg := msg.(type) {
    case domainEventMsg:
        switch e := msg.(events.Event).(type) {
        case events.TurnStarted: // ...
        case events.ResponseEvent: // ...
        }
        return m, m.waitForEvent() // Safe to queue next read

    case tea.KeyMsg:
        if msg.Type == tea.KeyCtrlC { return m, tea.Quit }
        return m, nil // NO channel read

    default:
        return m, nil // NO channel read
    }
}
```

#### 2. [TECHNICAL DEBT] Event Loss Strategy Violation
**File:** `internal/ui/tui/progress/renderer.go`
**Issue:** In PR #1223, the event bridge was updated to protect critical state-driving transitions (`TurnStarted`, `TurnStatusEvent`) from being dropped if the telemetry buffer fills up. This PR elevates `events.ResponseEvent` to be a critical UI state driver (it populates the response markdown view), but routes it to the `default:` block. During high-throughput telemetry bursts, `ResponseEvent`s will be silently dropped without backpressure, resulting in missing AI outputs.
**Directive:** Add `events.ResponseEvent` to the protected backpressure block:
```go
case events.TurnStarted, events.TurnStatusEvent, events.ResponseEvent:
    select {
    case ch <- e:
    case <-time.After(100 * time.Millisecond):
    }
```

#### 3. [REFACTOR] Responsive Markdown Layout 
**File:** `internal/ui/tui/progress/renderer.go` / `model.go`
**Issue:** `mdRender` statically closes over a `glamour.TermRenderer` instantiated at startup. It lacks viewport awareness. If the user resizes their terminal during an AI response generation, the wrapped markdown layout will break because the Bubble Tea model is ignoring `tea.WindowSizeMsg`.
**Directive:** 
1. Capture `tea.WindowSizeMsg` in `Update()` to track `m.width`. 
2. Modify the dependency interface: `mdRender func(text string, width int) string`.
3. In `renderer.go`, configure `glamour.WithWordWrap(width)` dynamically so the markdown reflows fluidly to fit the active window.

thptcnec:
### Architectural Review

The previous architectural blockers regarding the goroutine leak and state machine corruption have been correctly addressed using the `domainEventMsg` wrapper. The event loss strategy was also properly mitigated by adding `events.ResponseEvent` to the backpressure selection block.

However, the implementation of the responsive markdown layout directive was bypassed, and there are structural improvements needed for idiomatic Go rendering and state consistency.

#### 1. [ARCHITECTURAL BLOCKER] Bypassed Responsive Layout Requirement
**File:** `internal/ui/tui/progress/renderer.go`
**Issue:** The `mdRender` function signature correctly accepts the terminal `width`, but the implementation explicitly ignores it via `_ = width` and continues to close over a static `glamour.TermRenderer` instantiated at startup. This acts as a compiler workaround; if the terminal is resized, the markdown will not reflow, breaking the UI.
**Directive:** You must dynamically apply the width using `glamour.WithWordWrap(width)`. Since re-instantiating `TermRenderer` on every chunk of a streaming response is computationally expensive, use closure state to cache the renderer and only recreate it when the `width` actually changes:

```go
var cachedRenderer *glamour.TermRenderer
var lastWidth int

mdRender := func(text string, width int) string {
    if width != lastWidth || cachedRenderer == nil {
        opts := []glamour.TermRendererOption{glamour.WithAutoStyle()}
        if width > 0 {
            opts = append(opts, glamour.WithWordWrap(width))
        }
        
        tr, err := glamour.NewTermRenderer(opts...)
        if err != nil {
            return text
        }
        cachedRenderer = tr
        lastWidth = width
    }
    
    out, err := cachedRenderer.Render(text)
    if err != nil {
        return text
    }
    return out
}
```

#### 2. [REFACTOR] Unhandled Resize Edge Case on Existing Text
**File:** `internal/ui/tui/progress/model.go`
**Issue:** When a `tea.WindowSizeMsg` is received, `m.width` is updated, but `m.responseText` is not re-rendered. If the terminal is resized after an LLM chunk arrives or when the response is already complete, the text will remain statically formatted to the old width.
**Directive:** Keep `View()` pure by retaining the raw text in the model state and triggering a re-render explicitly when the window size changes.

1. Add `rawResponseText string` to `model`.
2. In `events.ResponseEvent`, store the raw text and render it:
```go
        case events.ResponseEvent:
            m.rawResponseText = extractResponseText(e.Content)
            if m.mdRender != nil {
                m.responseText = strings.TrimRight(m.mdRender(m.rawResponseText, m.width), "\n")
            } else {
                m.responseText = m.rawResponseText
            }
            return m, m.waitForEvent()
```
3. In `tea.WindowSizeMsg`, re-render if text exists:
```go
    case tea.WindowSizeMsg:
        m.width = msg.Width
        if m.rawResponseText != "" && m.mdRender != nil {
            m.responseText = strings.TrimRight(m.mdRender(m.rawResponseText, m.width), "\n")
        }
        return m, nil
```

#### 3. [TECHNICAL DEBT] Inefficient String Concatenation in View 
**File:** `internal/ui/tui/progress/model.go` (in `View()`)
**Issue:** The `View()` method uses repeated `+=` string concatenations. Go strings are immutable, meaning every `+=` allocates a new string in memory. Since Bubble Tea calls `View()` frequently (up to 60 FPS) during active rendering, this generates unnecessary garbage collection pressure and violates core Go patterns.
**Directive:** Follow idiomatic Go patterns by using `strings.Builder` to construct the view output with a single allocation.

```go
func (m *model) View() string {
    var sb strings.Builder
    
    ts := m.timestamp.Format("15:04:05")
    sb.WriteString(fmt.Sprintf("╭─ Turn %d - %s\n", m.turn, m.sessionName))
    sb.WriteString(fmt.Sprintf("[%s] Payload: ~%d/%d tokens - %s - %s",
        ts, m.tokens, m.maxTokens, m.sessionName, m.modelName))
    
    if m.responseText != "" {
        sb.WriteString("\n" + m.responseText)
    }
    
    if m.postCallStatus != nil {
        sb.WriteString("\n\n")
        sb.WriteString(fmt.Sprintf("[%s] Payload: %d/%d tokens - %s - %s\n",
            m.timestamp.Format("15:04:05"),
            m.postCallStatus.Metrics.PromptTokens,
            m.maxTokens, m.sessionName, m.modelName))
        sb.WriteString(formatMetricsLine(m.postCallStatus.Metrics,
            m.postCallStatus.StartTime, m.timestamp, m.postCallStatus.CurrentTurns+1))
    }
    
    if m.finalCostLine != "" {
        sb.WriteString("\n" + m.finalCostLine)
    }
    
    sb.WriteString("\n")
    return sb.String()
}
```

thptcnec:
### Architectural Review: Approved

The latest commit (`0a80dd0`) successfully resolves all findings from the previous review. 

#### Verification
1. **[RESOLVED] Responsive Layout**: The markdown rendering in `renderer.go` correctly implements the stateful closure (`cachedRenderer`, `lastWidth`). It dynamically adapts to width changes while preventing computationally expensive `glamour.TermRenderer` re-allocations on every streamed token chunk.
2. **[RESOLVED] Unhandled Resize Edge Case**: The `model` now accurately persists `rawResponseText`. The `tea.WindowSizeMsg` branch correctly delegates to `mdRender` and updates `m.responseText`, allowing existing text to fluidly reflow when the terminal is resized. Tests for this behavior have also been added.
3. **[RESOLVED] Inefficient String Concatenation**: `View()` has been successfully refactored to use `strings.Builder`, eliminating memory churn and garbage collection pressure in the 60 FPS Bubble Tea hot loop.

This PR adheres to established UI patterns and the Clean Architecture directives. Exceptional turnaround.

**Action**: Ready for merge.

PR #1225, opened 2026-07-11
fix: use SessionTurns for turn number instead of TurnStarted index (#1216)

thptcnec:
## Summary

The progress TUI was showing per-session turn numbers (always starting at 1) instead of the absolute historical turn count (~200+). The default renderer uses `TurnStatus.SessionTurns + 1` — this PR aligns the TUI to the same source.

## Change

| Field | Before | After |
|---|---|---|
| Turn source | `TurnStarted.Turn + 1` (zero-based per-session index) | `TurnStatus.SessionTurns + 1` (total historical turns) |

`TurnStarted` no longer sets `m.turn`. `TurnStatusEvent` now populates it from `SessionTurns`, matching `renderTurnHeader` in the default renderer.

## Testing

All 24 tests updated and pass.

### DIFF
--- a/internal/agent/orchestrator/engine_phases.go
+++ b/internal/agent/orchestrator/engine_phases.go
@@ -26,7 +26,11 @@ func (p *GuardStep) Process(ctx context.Context, Turn *Turn) (ProcessResult, err
 		return ProcessResult{}, NewAgentError(llm.ErrTerminal, fmt.Sprintf("turn %d exceeds limit %d", Turn.Index, maxTurns), llm.ErrMaxTurnsReached)
 	}
 
-	evt := events.TurnStarted{Turn: Turn.Index, MaxTurns: maxTurns}
+	evt := events.TurnStarted{
+		Turn:         Turn.Index,
+		SessionTurns: Turn.CtxManager.History.GetTotalEntries() / 2,
+		MaxTurns:     maxTurns,
+	}
 	if err := events.SafePublish(ctx, Turn.Events, evt); err != nil {
 		if errors.Is(err, events.ErrBusNotInitialized) {
 			return ProcessResult{NextPhase: PhaseRefining}, nil
--- a/internal/agent/orchestrator/turn_engine_event_test.go
+++ b/internal/agent/orchestrator/turn_engine_event_test.go
@@ -25,7 +25,8 @@ func TestTurnEngine_EventPublishFailure(t *testing.T) {
 	badBus := &agenttest.MockEventBusFail{PublishErr: errors.New("simulated publish failure")}
 
 	strategy := sessctx.NewStrategy(&agenttest.MockTokenCounter{})
-	cm := sessctx.NewManager(strategy, nil, badBus, nil)
+	mockHistory := &agenttest.MockHistoryManager{}
+	cm := sessctx.NewManager(strategy, mockHistory, badBus, nil)
 
 	Turn := &orchestrator.Turn{
 		Events:     badBus,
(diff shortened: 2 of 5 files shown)


thptcnec:
**Review from Principal Software Architect**

I have reviewed the changes in PR #1225 (`feat/issue-1216-turn-number`). 

**Status:** ⚠️ **Changes Requested**

The current implementation addresses the symptom (showing the wrong turn index type) but introduces an architectural inconsistency that leads to a UX bug.

### Findings

**1. [ARCHITECTURAL BLOCKER] - UI State Tearing & Event Sourcing Inconsistency**
By removing the `m.turn` assignment from `events.TurnStarted` and moving it strictly to `events.TurnStatusEvent`, you have uncoupled the state update (`m.turn`) from the phase transition (`stateThinking`). 

* **The Bug:** `TurnStarted` fires at the absolute beginning of the Turn (`PhaseGuard`). `TurnStatusEvent` fires later, after context refinement (`PhaseInference`), which may involve disk I/O, token counting, or summarization. Because Bubble Tea triggers a `View()` re-render on state transitions, transitioning to `stateThinking` *without* updating `m.turn` will cause the UI to briefly render "Turn 0" (or the previous turn's number) while the spinner spins, and then visibly jump to the correct number later.
* **The Root Cause:** The UI is contorting its state machine because the `TurnStarted` event payload is deficient (it lacks `SessionTurns`).

### 🛠️ Current vs. Scalable Design

**Current (Technical Debt):**
* UI relies on a delayed event (`TurnStatusEvent`) to determine the current turn state.
* UI state is temporally out of sync during Context Refinement, causing visual UI tearing.

**Scalable (Clean Architecture):**
* Enrich the source event so the UI adapter can remain purely reactive and immediately consistent.

**Action Plan (Refactor):**

**1. Enrich the Domain Event (`internal/domain/events/types.go`):**
Add `SessionTurns` to the initial event so subscribers have full context from the start.
```go
type TurnStarted struct {
    Turn         int // Current tool-loop index
    SessionTurns int // NEW: Overall historical session turn count
    MaxTurns     int
}
```

**2. Populate it in the Orchestrator (`internal/agent/orchestrator/engine_phases.go`):**
In `GuardStep`, inject the calculated session turns:
```go
evt := events.TurnStarted{
    Turn:         Turn.Index,
    SessionTurns: Turn.CtxManager.History.GetTotalEntries() / 2,
    MaxTurns:     maxTurns,
}
```

**3. Restore Immediate UI Consistency (`internal/ui/tui/progress/model.go`):**
Use the enriched payload to set the UI state *immediately* upon entering the Thinking state (and keep the `TurnStatusEvent` logic as a safeguard or metric update):
```go
case events.TurnStarted:
    m.turn = e.SessionTurns + 1 // Immediately consistent
    m.currentState = stateThinking
    return m, m.waitForEvent()
```

Please apply this architectural refactor. It preserves your business goal (displaying the session turn instead of the tool loop index) while ensuring the TUI rendering loop remains robust and tear-free.

thptcnec:
**Architecture Review Follow-up**

Thank you for addressing the feedback. I have reviewed the latest commit (`4bcc0f53`).

**Status:** ✅ **Approved**

The implementation perfectly aligns with the required architectural changes:
1. `SessionTurns` has been successfully integrated into the `TurnStarted` event, enriching the payload.
2. The `GuardStep` orchestrator correctly computes `SessionTurns` immediately using the internal manager history.
3. The TUI `model` correctly uses this payload to achieve immediate state consistency, resolving the state-tearing bug.
4. Unit and integration tests pass cleanly.

This guarantees a stable view-render cycle in Bubble Tea without visual jank. Excellent work! This PR is ready to merge from an architectural standpoint.

PR #1228, opened 2026-07-11
feat(progress): render tool logs and system messages in TUI mode

thptcnec:
## Summary

Adds rendering of 5 previously-dropped event types to the progress TUI mode (`-o`/`--tui-output`), matching Bridge mode output format.

## Events Added

| Event | Example output |
|-------|---------------|
| `ToolCallEvent` | `[13:32:13] [Tool Engine] Step 1/3` |
| | `[13:32:13] [Tool Reason] Stage the formatting fix` |
| | `[13:32:13] [Tool Action] execute_command(command: git add ...)` |
| `ToolResultEvent` | `[13:32:13] [Tool Result] execute_command: Exit Code: 0` |
| `ToolOutputStreamEvent` | `[13:32:13] [Tool Output] Executing...` |
| `SystemMessageEvent` | `[13:32:13] [Error] failed to connect to API` |
| `StatusUpdate` | `[13:32:13] [Warning] retry attempt 2 of 3` |

## Implementation

- `toolLogs []string` field on model, accumulated per turn
- `appendToolLog(tag, message)` helper with timestamp formatting
- Truncation: Tool Result text capped at 200 chars, newlines collapsed
- Level mapping (shared by 3 event types): `error`→Error, `warn`→Warning, `info`→Info, `output`→Tool Output, default→System
- Cleared on `TurnStarted`, rendered in `View()` between header and response text

## Tests

58 tests (53 existing + 5 new), all passing. Covers all event types, truncation, newline collapse, nil guards, and `TurnStarted` clearance.

### DIFF
--- a/internal/ui/tui/progress/model.go
+++ b/internal/ui/tui/progress/model.go
@@ -6,6 +6,7 @@ package progress
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 	"time"
 
@@ -154,6 +155,46 @@ func (m *model) appendToolLog(tag, message string) {
 	m.toolLogs = append(m.toolLogs, fmt.Sprintf("[%s] [%s] %s", ts, tag, message))
 }
 
+// appendLevelEventLog appends a timestamped log line with a level-mapped
+// prefix. Used by ToolOutputStreamEvent, SystemMessageEvent, and StatusUpdate.
+func (m *model) appendLevelEventLog(level, message string) {
+	prefix := "System"
+	switch level {
+	case "error":
+		prefix = "Error"
+	case "warn":
+		prefix = "Warning"
+	case "output":
+		prefix = "Tool Output"
+	case "info":
+		prefix = "Info"
+	}
+	m.appendToolLog(prefix, message)
+}
+
+// safeTruncate truncates s to at most maxLen bytes, ensuring the cut
+// never lands in the middle of a multi-byte UTF-8 character.
+// If s is already shorter than maxLen, it is returned unchanged.
+func safeTruncate(s string, maxLen int) string {
+	if len(s) <= maxLen {
+		return s
+	}
+	// Walk runes to find the byte index of the (maxLen)th rune.
+	var count, byteIdx int
+	for i := range s {
+		if count == maxLen {
+			byteIdx = i
+			break
+		}
+		count++
+	}
+	if byteIdx > 0 {
+		return s[:byteIdx]
+	}
+	// maxLen is 0 or s starts with a multi-byte rune exceeding maxLen.
+	return ""
+}
+
 // Init returns the initial command to start listening for events.
 func (m *model) Init() tea.Cmd {
 	return m.waitForEvent()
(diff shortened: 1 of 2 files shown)


thptcnec:
Overall, this PR introduces a solid TUI tracking component with bounded memory (resetting `toolLogs` per turn is great for preventing memory leaks on long-running sessions). The decision to slice `e.Result.Text` before running `strings.ReplaceAll` is an excellent memory optimization for large tool payloads.

However, I have identified a few architectural and technical debt points to address before merging:

### 1. [TECHNICAL DEBT] Map Iteration Non-determinism
Iterating over `fc.Args` (a `map[string]interface{}`) produces randomized order in Go. If a tool call has multiple arguments, the UI output will randomly shuffle their order on every render, leading to user confusion and flaky tests.
**Scalable Pattern:**
```go
var keys []string
for k := range fc.Args {
    if k != "reason" {
        keys = append(keys, k)
    }
}
sort.Strings(keys)

var parts []string
for _, k := range keys {
    valStr := fmt.Sprintf("%v", fc.Args[k])
    // slice logic...
    parts = append(parts, fmt.Sprintf("%s: %s", k, valStr))
}
```

### 2. [TECHNICAL DEBT] UTF-8 Unsafe String Slicing
Slicing bytes directly with `snippet[:197]` or `valStr[:186]` is unsafe. If the slice lands in the middle of a multi-byte UTF-8 character (common in source code or localized tool outputs), it will result in invalid UTF-8 and render broken characters in the terminal.
**Memory-Efficient Idiomatic Pattern (Zero Allocation):**
```go
if len(snippet) > 200 {
    var count, byteIdx int
    for i := range snippet { // range decodes runes and returns their byte index
        if count == 197 {
            byteIdx = i
            break
        }
        count++
    }
    if byteIdx > 0 {
        snippet = snippet[:byteIdx] + "..."
    }
}
```

### 3. [REFACTOR] DRY Violation in Level Prefixing
The exact same `switch e.Level` logic is repeated verbatim for `events.ToolOutputStreamEvent`, `events.SystemMessageEvent`, and `events.StatusUpdate`. Consolidate this into a single helper method.
**Refactor Example:**
```go
func (m *model) appendLevelEventLog(level, message string) {
    prefix := "System"
    switch level {
    case "error": prefix = "Error"
    case "warn": prefix = "Warning"
    case "info": prefix = "Info"
    case "output": prefix = "Tool Output"
    }
    m.appendToolLog(prefix, message)
}
```

### 4. [REFACTOR] Idiomatic Empty Slice Check
In `events.ToolCallEvent`, checking `if e.Calls == nil` handles `nil` but skips empty slices (`[]*llm.FunctionCall{}`). It's more idiomatic to use `if len(e.Calls) == 0`, which safely handles both cases and prevents the UI from logging a solitary `[Tool Engine] Step X/Y` with zero actual actions.

thptcnec:
The new commits look fantastic! All architectural feedback has been addressed properly:

- `appendLevelEventLog` cleaned up the DRY violations and significantly reduced the cognitive load in the event router.
- `safeTruncate` is a very elegant solution to the UTF-8 problem; stepping through runes guarantees memory safety for all multi-byte inputs.
- The `keys` sorting mechanism prevents non-deterministic log flapping.
- `if len(e.Calls) == 0` is idiomatic and resilient.

Great job getting this polished. Approved from an architectural standpoint!

PR #1237, opened 2026-07-12
feat(progress): fixed-zone TUI layout with sticky status, real-time tool logs, and exit hold

thptcnec:
## Summary

Overhauls the Progress TUI (`-o` flag) from a flat, unbounded render to a fixed three-zone layout:

```
header(2) | gap | body(scrollable) | gap | footer(4)
```

## Changes (17 commits)

**Layout**
- Fixed zones: header, scrollable body, footer — each pinned in place
- Top-aligned body zone with bottom padding
- Visible separator lines between header/body and body/footer
- Footer expanded to 4 lines: post-call payload, metrics, final cost, spinner

**Content visibility**
- Footer status lines sticky across turn boundaries (no more blank gaps)
- Tool logs sticky across turns (visible during tool execution)
- Height default to 24 to prevent renderMinimal on first frame

**Event pipeline**
- Fixed spinner tick starving the event channel — tool logs now stream in real-time
- Restored `tea.WithInput(nil)` + 50ms idle tick to prevent VS Code xterm.js header drop

**Exit**
- Session holds screen with "Press Ctrl+C to exit" until user dismisses
- Cleanup blocks until Bubble Tea exits

## Coverage
- All 5 View-related methods at 100%
- 78 packages green, 0 lint issues

### DIFF
--- a/internal/ui/tui/progress/model.go
+++ b/internal/ui/tui/progress/model.go
@@ -547,17 +547,18 @@ func formatFinalLine(status events.TurnStatus, turnCost float64) string {
 // header (2 lines), scrollable body, footer (3 lines).
 // Falls back to renderMinimal when the terminal is too small (height < 5).
 func (m *model) View() string {
-	if m.height < 7 {
+	if m.height < 8 {
 		return m.renderMinimal()
 	}
 
-	// Body gets everything between header+separator (3 lines) and footer (4 lines).
-	availableBody := m.height - 7
+	// Body gets everything between header+separator (3 lines) and separator+footer (5 lines).
+	availableBody := m.height - 8
 
 	var sb strings.Builder
 	sb.WriteString(m.renderHeader())
 	sb.WriteString("\n")
 	sb.WriteString(m.renderBody(availableBody))
+	sb.WriteString("\n")
 	sb.WriteString(m.renderFooter())
 	return sb.String()
 }
--- a/internal/ui/tui/progress/model.go
+++ b/internal/ui/tui/progress/model.go
@@ -109,6 +109,7 @@ type model struct {
 
 	currentState        state
 	width               int // terminal width, updated via WindowSizeMsg
+	height              int // terminal height, updated via WindowSizeMsg
 	turn                int
 	modelName           string // display name, e.g. "deepseek-v4-pro"
 	sessionName         string // e.g. "architect-johndoe"
@@ -232,6 +233,7 @@ func (m *model) handleKeyMsg(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 // handleWindowSizeMsg processes terminal resize events.
 func (m *model) handleWindowSizeMsg(msg tea.WindowSizeMsg) (tea.Model, tea.Cmd) {
 	m.width = msg.Width
+	m.height = msg.Height
 	if m.rawResponseText != "" && m.mdRender != nil {
 		m.responseText = strings.TrimRight(m.mdRender(m.rawResponseText, m.width), "\n")
 	}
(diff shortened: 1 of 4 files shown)


thptcnec:
### Architecture Review

I've analyzed the changes introduced in this PR. While moving to a viewport-based layout with an alternate screen buffer improves the terminal UI structure, there are a few significant architectural and performance issues that need to be addressed before merging.

#### 🔴 [ARCHITECTURAL BLOCKER] Impure View Function & Viewport Misuse
The Bubble Tea framework strictly follows the Elm architecture, which mandates that the `View()` function must be **100% pure**.
*   **Current Design:** In `model.go`, `m.bodyVP.GotoBottom()` is invoked directly inside `View()`. This mutates the internal state (`YOffset`) of the viewport during the render cycle, which breaks the architectural paradigm and can cause unpredictable render loops or data races.
*   **Scalable Fix:** State mutation must happen exclusively in the `Update()` function. Call `m.bodyVP.GotoBottom()` inside `Update()` when new log lines or response chunks are appended, keeping `View()` strictly read-only.
*   **Missing Event Routing:** The `bodyVP` viewport is instantiated but its `Update()` method is never called in `model.Update`. Without routing `tea.KeyMsg` or `tea.MouseMsg` to `m.bodyVP.Update(msg)`, the viewports cannot be scrolled.

#### 🟡 [TECHNICAL DEBT] Memory Leak via Timer Accumulation
The `makeSubscriber` callback in `renderer.go` uses `time.After` inside `select` statements to handle timeouts.
*   **Current Design:**
    ```go
    select {
    case ch <- e:
    case <-time.After(50 * time.Millisecond):
    }
    ```
*   **The Go Trap:** The `time.After` timer is not garbage collected until it fires. In a high-throughput event stream (e.g., receiving fast LLM token chunks), every successful send creates a dormant timer. This will spawn thousands of temporary timers in memory, causing unnecessary GC thrashing.
*   **Scalable Fix:** Use `time.NewTimer` and explicitly stop it.
    ```go
    timer := time.NewTimer(50 * time.Millisecond)
    defer timer.Stop()
    select {
    case ch <- e:
    case <-timer.C:
    }
    ```

#### 🔵 [REFACTOR] Conflicting Input Expectations
*   **Current Design:** `tea.NewProgram` is initialized with `tea.WithInput(nil)`. This detaches `stdin` completely.
*   **Impact:** The comment `// wait for user to dismiss TUI (Ctrl+C or q)` cannot be fulfilled. While `Ctrl+C` works (because Bubble Tea intercepts `SIGINT` signals natively and injects a `KeyCtrlC` message), normal keys like `q`, arrow keys, or mouse wheel scrolls are entirely ignored.
*   **Fix:** Remove `tea.WithInput(nil)` from `renderer.go`. Update your `handleKeyMsg` to catch `"q"`, `"esc"`, etc., and ensure unhandled keys are passed down to the viewport's `Update` method so the user can scroll through the logs after the session finishes.

thptcnec:
### Architecture Review: Follow-up

I've reviewed the latest commits addressing the previous architectural findings. The implementation is significantly improved.

#### ✅ Resolved Items
1.  **Pure View Function:** State mutations (`m.bodyVP.GotoBottom()`) have been correctly moved from `View()` to the `Update()` lifecycle (e.g., inside `handleLeveledMessage`, `handleResponseEvent`, etc.). The Elm architectural paradigm is now restored.
2.  **Memory Leak via Timers:** The `time.After` anti-pattern in `renderer.go` has been successfully refactored. The use of `time.NewTimer` with explicit `timer.Stop()` upon successful channel send prevents timer accumulation and garbage collection thrashing in high-throughput streams.
3.  **Input Expectations & UI Blocking:** `tea.WithInput(nil)` has been removed, restoring `stdin` capability. The `handleKeyMsg` function now correctly processes "q" and routes unhandled key events down to `m.bodyVP.Update(msg)`, allowing the viewport to handle scrolling as expected.

#### 💡 Minor Technical Note (For Future Consideration)
While `timer.Stop()` is correctly placed inside the success case, the optimal idiomatic pattern ensures that the timer is completely stopped and drained regardless of the branch taken, though in this case, since the function returns immediately on timeout (and the timer has fired so `C` is drained), the current implementation is perfectly safe and leak-free.

The PR is now architecturally sound and scalable. Great work!

PR #1254, opened 2026-07-22
llm: add ExtractDocument to LLMClient + read_document agent tool

gosharplite:
Adds `read_document` agent tool that extracts text from documents (PDF, DOCX, MD, etc.) using the provider's file extraction API.

## Changes

- **types.go**: `ExtractDocument(ctx, data, filename) (string, error)` added to `LLMClient` interface + `ErrNotImplemented` sentinel
- **openai/client.go**: delegates to existing `extractDocument` in `files.go` (Kimi file API: upload → get content → delete)
- **anthropic, gemini**: return `ErrNotImplemented`
- **resilient_client, failover, lazy_client**: delegate through
- **media.go**: `readDocument` method on `mediaManager` — security-validated file read → `client.ExtractDocument` → returns extracted text. Handles `ErrNotImplemented` for non-Kimi providers. Tool registered as `read_document` in media toolkit.
- All mock/stub types updated with `ExtractDocument` stubs.

## Usage

```
read_document(filepath="/path/to/report.pdf")
→ "--- Extracted content from report.pdf ---\n[full text]"
```

Replaces the 5-turn `pdftotext`/`gs`/`read_image` workaround for PDFs. Works only with Kimi provider (uses `/v1/files` with `purpose="file-extract"`).

### DIFF
--- a/internal/domain/llm/types.go
+++ b/internal/domain/llm/types.go
@@ -81,6 +81,10 @@ type AssetResolver interface {
 type LLMClient interface {
 	SendChat(ctx context.Context, history []*Content, tools []*tools.ToolDeclaration, resolver AssetResolver) (*Content, *Metrics, error)
 	GenerateImages(ctx context.Context, model, prompt string, mimeType string) ([][]byte, error)
+	// ExtractDocument extracts text content from a document (PDF, DOCX, MD, etc.)
+	// by uploading it to the provider's file extraction API. Returns the extracted
+	// text. Providers that don't support document extraction return ErrNotImplemented.
+	ExtractDocument(ctx context.Context, data []byte, filename string) (string, error)
 	RefreshAuth() error
 }
 
@@ -96,6 +100,11 @@ var (
 
 	// ErrMaxTurnsReached is returned when the model reaches the turn limit.
 	ErrMaxTurnsReached = errors.New("maximum tool execution turns reached")
+
+	// ErrNotImplemented is returned by provider methods that are not supported
+	// by a particular provider (e.g., GenerateImages on OpenAI, ExtractDocument
+	// on Anthropic/Gemini).
+	ErrNotImplemented = errors.New("not implemented for this provider")
 
 	// errBudgetExceeded is returned when the session cost exceeds the configured budget.
 	errBudgetExceeded = errors.New("session budget exceeded")
--- a/internal/infrastructure/di/container_test.go
+++ b/internal/infrastructure/di/container_test.go
@@ -76,6 +76,10 @@ func (m *mockLLMClient) RefreshAuth() error {
 	return nil
 }
 
+func (m *mockLLMClient) ExtractDocument(ctx context.Context, data []byte, filename string) (string, error) {
+	return "", llm.ErrNotImplemented
+}
+
 func (m *mockLLMClient) Generate(ctx context.Context, input []*llm.Content, toolDecls []*tools.ToolDeclaration, resolver llm.AssetResolver) (*llm.Content, *llm.Metrics, error) {
 	m.generateCalls++
 	if m.GenerateFunc != nil {
--- a/internal/infrastructure/di/golden_path_integration_test.go
+++ b/internal/infrastructure/di/golden_path_integration_test.go
@@ -41,6 +41,10 @@ func (m *goldenPathMockClient) GenerateImages(ctx context.Context, model, prompt
 
 func (m *goldenPathMockClient) RefreshAuth() error { return nil }
 
+func (m *goldenPathMockClient) ExtractDocument(ctx context.Context, data []byte, filename string) (string, error) {
+	return "", llm.ErrNotImplemented
+}
+
 // TestGoldenPath_ConfigToShutdown exercises the full vertical:
 // DI bootstrap → agent init → chat → graceful shutdown.
 // It runs in-process under -short and -race.
--- a/internal/infrastructure/di/lazy_client.go
+++ b/internal/infrastructure/di/lazy_client.go
@@ -61,5 +61,13 @@ func (lc *lazyClient) RefreshAuth() error {
 	return lc.client.RefreshAuth()
 }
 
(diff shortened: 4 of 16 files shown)


gosharplite (commented):
## Architectural Review — `llm: add ExtractDocument to LLMClient + read_document agent tool`

**Verdict: Request changes** — one functional defect in the execution-context wiring, plus a growing interface-segregation problem that now needs an ADR.

Reviewed against the surrounding execution path: executor timeouts (`internal/agent/executor`), heartbeat plumbing (`internal/pkg/telemetry`), failover/resilient decorators, and `openai/files.go`.

### What's solid ✅

- **Layering is correct.** Domain port in `internal/domain/llm`, provider impls in infrastructure, tool in `tools/integrations` depending only on ports (`FileSystem`, `PathValidator`, `LLMClient`). Dependency direction points inward; decorator chain (lazy → failover → resilient → provider) updated consistently with compile-time conformance checks.
- **Security ordering is right:** validate path → read → egress, and `filepath.Base()` strips directory info before the filename leaves the box.
- **Sentinel mapping across the layer boundary** (`llm.ErrNotImplemented` → `tools.ErrNotImplemented`) is clean, and heartbeat placement (after file read, wrapping only the network phase) is correct.

---

### [ARCHITECTURAL BLOCKER] 1. `read_document` is not `LongRunning`, but behaves like one

**Current:** `readDocument` performs **3 sequential HTTP round-trips** (upload → get content → delete) and starts a heartbeat — identical in shape to `createImage`, which *is* registered `LongRunning: true`. But `read_document` uses plain `RegisterToToolkit`, so it inherits the **30s default** `ToolTimeout` (`executor.go:212`) instead of the 5-minute budget.

**Failure cascade on a large PDF (>30s):**

1. `safetyDecorator` cancels the ctx → returns `llm.ErrTransient` → agent **retries** → duplicate uploads.
2. The detached goroutine's `defer c.deleteFile(ctx, fileID)` in `files.go` runs with the **already-cancelled ctx** → delete fails → **orphaned file on the provider** (cost + data-residency leak). The heartbeat started in `readDocument` is cosmetic — it feeds liveness monitoring, not the timeout budget.

**Fix:**

```go
}, m.readDocument, tools.ToolOptions{LongRunning: true}); err != nil {
```

and harden the cleanup in `files.go` against cancellation (idiom already used in `decorators.go:handleTimeout`):

```go
defer func() {
    cleanupCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 10*time.Second)
    defer cancel()
    _ = c.deleteFile(cleanupCtx, fileID)
}()
```

### [TECHNICAL DEBT] 2. `LLMClient` is accreting into a capability-fat interface — ADR time

This PR touches **15 files**, ~9 of which are pure mechanical stubs (`ErrNotImplemented`) across 2 more providers, 3 decorators, and 4 mocks. That's the second capability after `GenerateImages` to pay this tax; the third (audio? video?) will cost more.

**Current (fat port, runtime capability probing):**

```go
type LLMClient interface {
    SendChat(...); GenerateImages(...); ExtractDocument(...); RefreshAuth()
}
```

**Scalable (consumer-side narrow port, per Go idiom "define interfaces where they're used"):**

```go
// in tools/integrations, where it's consumed
type DocumentExtractor interface {
    ExtractDocument(ctx context.Context, data []byte, filename string) (string, error)
}
// DI wiring does a type-assert once; tool is registered only if supported —
// capability becomes compile-time/startup-time, ErrNotImplemented probing disappears.
```

The PR is *consistent with the existing precedent*, so this doesn't block merge — but the trade-off (fat client vs. capability interfaces, incl. capability-aware failover routing) should be captured in an **ADR before the next capability lands**. Related latent issue: `failoverGateway.ExtractDocument` pins to `clients[0]`, so an Anthropic-primary/Kimi-secondary chain makes the tool dead despite a capable provider sitting in the chain. That belongs in the same ADR.

### [TECHNICAL DEBT] 3. Zero test coverage on a security-sensitive handler

`readDocument` (path validation → file read → **third-party egress**) ships with **no direct tests** — no happy path, no `ErrNotImplemented` mapping, no security-rejection path, no `client == nil` path. The only `media_test.go` changes are ordinal bumps (`failAfter` 27→28, 29→30), which itself proves the second problem: **call-count-based fault injection is brittle** — every new registration anywhere in `RegisterAll` shifts all downstream indices. Inject faults by *tool name* in `faultyRegistry` instead of by call ordinal.

### [REFACTOR] 4. Partial sentinel adoption + stale doc comment

- `GenerateImages` on anthropic/openai still returns ad-hoc `fmt.Errorf("GenerateImages not implemented for X")` — so `errors.Is(err, llm.ErrNotImplemented)` fails for the image path while working for documents. Retrofit `GenerateImages` to the sentinel (small diff, same PR or fast follow).
- `failover.go` type doc still reads *"Non-Generate methods (SendChat, GenerateImages, RefreshAuth) delegate to the primary"* — add `ExtractDocument`.

### [REFACTOR] 5. Unbounded input size; ~2× memory amplification

`m.fs.ReadFile` loads the whole file; `uploadFile` then buffers the entire multipart body in a `bytes.Buffer`. No pre-check against provider limits (Kimi `file-extract` caps file size) → guaranteed 4xx *after* the allocation spike. The response side already disciplines itself with `io.LimitReader(maxResponseBytes)` — mirror that on the input side with a size gate and a clear tool-level error. Related: the description promises "Returns the **full** extracted text" — unbounded context injection; consider a truncation strategy or a size hint in the schema description so the model self-limits.

### ⚖️ Decision point (not blocking): egress consent

Most trust-boundary-crossing tools here set `RequiresConsent: true` (workspace writes, skillssh, ado/atlassian, teams). `read_document` uploads file contents the model has never seen to a third-party file API where they persist server-side until deleted. The `read_image` precedent (bytes flow to the provider in the chat payload, no consent) argues against; transient server-side file storage argues for. Either set `RequiresConsent: true` or record the rationale in the ADR from finding 2.

---

**Requested before merge:** finding 1 (timeout registration + cleanup ctx). **Strongly recommended:** finding 3 (tests). Findings 2/4/5 can be fast-follows, with the ADR gating the *next* provider capability.

Issue #1394, opened 2026-08-18
MCP: support stdio / local-process MCP servers (COMMAND + ARGS) alongside Streamable HTTP


gosharplite:
## Summary

tell-me-go consumes MCP servers **only over the Streamable HTTP transport** (ADR-067, #1373). Local MCP servers — the most common kind in the ecosystem (the reference `@modelcontextprotocol/server-*` suite: filesystem, git, fetch, memory, sequential-thinking, plus `uvx`/deno-based servers) — run as **stdio child processes**. Today the only way to use them is to wrap each one in an external stdio→Streamable HTTP bridge (e.g. `supergateway --outputTransport streamableHttp`) and supervise a process per server. That adds a Node runtime dependency, a port per server (exposing e.g. the filesystem on localhost), and a supervision layer — all of which disappear if tell-me-go can spawn the stdio server itself.

**Request:** first-class **stdio / local-process MCP support** — a `MCP_SERVERS.<name>.COMMAND` (+ `ARGS`) config that spawns the server as a child process, alongside the existing `URL`-based remote mode.

## Motivation

- Most MCP servers in the wild are stdio-only; remote HTTP MCP servers are the minority (GitHub Copilot, hosted gateways).
- The bridge workaround (`supergateway`, `mcp-proxy`) is per-server, per-port, and per-supervisor — significant operational overhead for "many local MCP services".
- tell-me-go already isolates the transport behind the `tools.MCPClient` domain port (ADR-055/060/067 injection pattern), so this is a contained addition, not an architectural change.

## Current state (evidence, dev @ 7ec422e)

- **Config** — `internal/domain/config/mcp_config.go`: `MCPServerConfig` has only `URL`, `TOKEN`, `USERNAME`, `AUTH` (`auto|gh|bearer|basic|none`), `REQUIRES_CONSENT`, `TIMEOUT`. No `COMMAND`/`ARGS`. Server keys constrained to `^[a-z0-9-]{1,24}$` (tool-name budget).
- **Transport** — `internal/infrastructure/mcp/client.go`: hard-wired to `sdkmcp.StreamableClientTransport{DisableStandaloneSSE: true}`.
- **DI** — `internal/infrastructure/di/mcp_factory.go`: injectable seam `newClient(endpoint, username, token string, timeout time.Duration) (tools.MCPClient, error)`; `Build` skips unconstructible servers with a warning rather than failing startup.
[...]

PR #1397, opened 2026-08-18
feat(mcp): stdio / local-process MCP servers (COMMAND + ARGS) (#1396)

### LINKED ISSUE #1396: MCP: support stdio / local-process MCP servers (COMMAND + ARGS) — implementation spec (supersedes #1394)
## Summary

tell-me-go consumes MCP servers **only over the Streamable HTTP transport** (ADR-067, #1373). Local MCP servers — the most common kind in the ecosystem (the reference `@modelcontextprotocol/server-*` suite: filesystem, git, fetch, memory, sequential-thinking, plus `uvx`/deno-based servers) — run as **stdio child processes**. Today the only way to use them is to wrap each one in an external stdio→Streamable HTTP bridge (e.g. `supergateway --outputTransport streamableHttp`) and supervise a process per server. That adds a Node runtime dependency, a port per server (exposing e.g. the filesystem on localhost), and a supervision layer — all of which disappear if tell-me-go can spawn the stdio server itself.

**This issue supersedes #1394.** It is the post-grill implementation spec: #1394 was pressure-tested in an Architect×Griller round (7 questions; transcript and corrections table in the [#1394 findings comment](https://github.com/gosharplite/tell-me-go/issues/1394#issuecomment-5323595497), full transcript [gist](https://gist.github.com/gosharplite/6b9643b718e717b0e92e46ab8113cdc1)), and the maintainer has since made the one decision the round left open — **consent semantics: trusted local spawns** (see §4). All other sections are the round's integrated, code-verified plan.

## Motivation

- Most MCP servers in the wild are stdio-only; remote HTTP MCP servers are the minority (GitHub Copilot, hosted gateways).
- The bridge workaround (`supergateway`, `mcp-proxy`) is per-server, per-port, and per-supervisor — significant operational overhead for "many local MCP services".
- tell-me-go already isolates the transport behind the `tools.MCPClient` domain port (ADR-055/060/067 injection pattern), so this is a contained addition, not an architectural change.

## Current state (evidence, dev @ 7ec422e)

- **Config** — `internal/domain/config/mcp_config.go`: `MCPServerConfig` has only `URL`, `TOKEN`, `USERNAME`, `AUTH` (`auto|gh|bearer|basic|none`), `RequiresConsent *bool`, `TIMEOUT`. No `COMMAND`/`ARGS`. Server keys constrained to `^[a-z0-9-]{1,24}$` (tool-name budget). `validate()` hard-requires a non-empty URL (~L84). `EffectiveAuth()` maps `""`→`"auto"`; `EffectiveTimeout()` defaults 300s.
- **Transport** — `internal/infrastructure/mcp/client.go`: hard-wired to `sdkmcp.StreamableClientTransport{DisableStandaloneSSE: true}`; `Close()` idempotent + mutex-guarded; `operationContext` bounds each op.
- **DI** — `internal/infrastructure/di/mcp_factory.go`: injectable seam `newClient(endpoint, username, token string, timeout time.Duration) (tools.MCPClient, error)`; `Build` runs a **sequential** loop with warn-and-skip per unconstructible server; `resolveServerToken` memoizes a `cachedToken` (unguarded — safe only because Build is single-threaded). **Catalog pins** at `mcp_factory.go:60-65` (`gh` token resolver) and `:157-160` (`resolveServerToken` default case) are ACCEPTED entries in `docs/architect/INTENTIONAL_NON_FIXES.md`.
- **Domain port** — `internal/domain/tools/mcp_client.go`: `MCPClient` = `ListTools` / `CallTool` / `Close` — transport-agnostic already.
- **Domain model** — `MCPServer` is defined as "over Streamable HTTP"; invariant `mcp-readonly-defaults` is URL-suffix-derived. `modelith-check` is a **hard gate** in `make check`/`check-full` — the model + rendered `.md` must change or CI fails.
- **go-sdk** (github.com/modelcontextprotocol/go-sdk v1.7.0): `Client.Connect(ctx, Transport, ...)` accepts any `Transport`; the `initialize`/`discover` handshake runs **synchronously within ctx**. `StdioTransport` is hard-wired to `os.Stdin`/`os.Stdout` (`transport.go:114-119`) — **one per process only** — so it is rejected; `IOTransport{Reader io.ReadCloser, Writer io.WriteCloser}` (L130-138) is spawnable per server. `ClientSession.Close()` → `ioConn.Close()` → `rwc.Close()` closes **both** pipes (sends stdin EOF to the child). The ioConn read loop terminates on `io.EOF` (dead child → pending calls fail fast) and blocks on a wedged child.

### 1. Config (`mcp_config.go`)

New fields: `COMMAND string`, `ARGS []string`, `DIR string` (working dir, optional), `ENV map[string]string` (extra env, optional).

```yaml
MCP_SERVERS:
  fs:
    COMMAND: "npx"
    ARGS: ["-y", "@modelcontextprotocol/server-filesystem", "/path/to/dir"]
  git:
    COMMAND: "uvx"
    ARGS: ["mcp-server-git"]
```

Validation (hard errors, checked in this order so the most specific error wins):
- `COMMAND` and `URL` are **mutually exclusive** — exactly one must be set (replaces the unconditional empty-URL check).
- `ARGS`/`DIR`/`ENV` without `COMMAND` → error.
- `COMMAND` + `EffectiveAuth() ∈ {bearer, basic}` → error — *"AUTH bearer/basic requires an HTTP endpoint; stdio (COMMAND) servers transmit no credentials."* Checked **before** the existing credential-required positive rules so the mode-conflict message wins. `auto`/`gh`/`none`/empty are accepted (inert under the stdio short-circuit). This is a **mode-conflict rule**, same family as COMMAND/URL exclusivity — it does not disturb ADR-067 §4's positive-rule-only credential philosophy.
- Stray `TOKEN`/`USERNAME` under stdio: tolerated (ADR-067 §4 stray-field tolerance).
- Key regex `^[a-z0-9-]{1,24}$` and TIMEOUT rules unchanged; tool names remain `mcp_<server>_<tool>`.

**Resolution contract (documented, verified stdlib semantics):** bare `COMMAND` (no path separator) → `exec.LookPath` in **tell-me-go's own process PATH** — *not* `ENV.PATH`, *not* `DIR`; separator-bearing `COMMAND` (`/abs`, `./rel`) → used as-is, relative resolved against `DIR` (exec.go:149-154); `ENV.PATH` governs the child only post-exec (it *does* reach the server's own subprocess resolution — uvx finding `mcp-server-git` — which is why it's set and why the failure is confusing); `${VAR}` expands at config load via `expandEnvHook` (`internal/infrastructure/config/config.go:187-190`, `os.ExpandEnv`) in `COMMAND`, `ARGS`, `DIR`, and `ENV` values. Windows bare-name lookup honors `PATHEXT` (`npx` → `npx.cmd`) via the same stdlib path.

### 2. Transport (`internal/infrastructure/mcp`)

A **separate `StdioClient` type** (`stdio_client.go`, same package — shares `convertCallToolResult`/`convertContentItem`/`toolErrorSummary`/`marshalInputSchema` without widening ADR-067 confinement). **Not** a mode on `Client` — HTTP and exec lifecycles have nothing in common.

- **Eager construction**: `NewStdioClient(cfg, logger)` — `exec.CommandContext` (ctx owned by the client, cancelable from `Close`), `cmd.Dir`, `cmd.Env = append(os.Environ(), cfg.Env...)`, `StdoutPipe`/`StdinPipe`, `cmd.Stderr = slogWriter` (stderr → slog at Info with `mcp_server=<name>`; never parsed as JSON-RPC), `cmd.Start()`, then `sdk.Connect(ctx, &sdkmcp.IOTransport{Reader: stdoutPipe, Writer: stdinPipe}, nil)` bounded by `EffectiveTimeout()`. Spawn/connect/handshake failures surface at `Build` as the existing warn+skip; handshake exceeding `TIMEOUT` kills the child and skips the server for the session (MCP_SERVERS read once, ADR-067 §10).
- **Reaper goroutine** started right after `Start()`: `go func() { c.waitDone <- cmd.Wait() }()` (`waitDone chan error`, buffered 1). The child is reaped the moment it dies — zombie prevention is not contingent on `Close()` being reached.
- **Fast-death pre-check** at the top of `ListTools`/`CallTool` (under the mutex): non-blocking select on `waitDone` — if the child has exited, return `mcp: stdio child %q exited: %w` **without touching the SDK**, so every subsequent call after death fails fast deterministically (the SDK alone does not guarantee this — its read loop terminates on EOF and teardown can race).
- **In-flight EOF annotation**: if `errors.Is(err, io.EOF) || errors.Is(err, sdkmcp.ErrConnectionClosed)` and the reaper confirms exit, annotate with the child-exit status. Wedge surfaces as the plain `context.DeadlineExceeded` wrap — the two failure modes are distinguishable by error text and latency.
- **`Close()` idempotent** + mutex-guarded: (1) `session.Close()` first — closes both pipes via `rwc.Close`, sending stdin EOF so a well-behaved child (and its tree, via the shared pipe) exits gracefully; (2) `cancel()` as the kill backstop for uncooperative children; (3) join the reaper (expected `signal: killed`/`context.Canceled` logged at debug, not surfaced). `factory.Close()` iterates tracked clients, so session teardown kills every child.
- **No kill on operation-timeout**: a slow legitimate call (e.g. a 200s filesystem op under TIMEOUT 300) is indistinguishable from a wedge; a timeout doesn't poison the session — a slow child can still serve the next call. Documented known limitation, stderr as the operator's visibility window.

**Process-tree guarantee (two-tier):** direct-child reaping is **deterministic** (reaper). Tree termination is **best-effort** via shared-pipe stdin EOF — launchers (`npx`, `uvx`) pass stdio through, they don't proxy it, so the grandchild sees EOF directly and SDK-built servers (`server.Run(ctx, &StdioTransport{})` read loop exits on `io.EOF`) self-terminate; the Unix SIGPIPE write-backstop covers a grandchild that writes after our read end closes. **Zombie vs orphan**: no zombies, guaranteed; orphans possible for transport-contract-violating servers (EOF-ignoring **and silent** grandchild on Unix; any EOF-ignoring grandchild on Windows — no signal backstop) — accepted residuals, documented. **No process-group kill this PR** (Unix-only platform split for a misbehaving-server edge; pgid-kill is wrong for a reparenting child; Windows needs Job Objects) — tracked future option in the ADR.

### 3. DI factory (`mcp_factory.go`)

- Replace the HTTP-shaped seam with a single injectable dispatch `newClientFor(serverCfg config.MCPServerConfig) (tools.MCPClient, error)` (one testable branching point); keep warn+skip.
- **Two-phase `Build`**: phase 1, **sequential** token pre-pass (`resolveServerToken` per server — preserves the exact single-flight `gh`-memoization semantics of today with zero new lock machinery; stdio short-circuits first: `if isStdio { return "", "", true }` before the auth switch, so `gh`/`auto` under stdio never resolve, never warn, never invoke the resolver); phase 2, **concurrent construction** — `sync.WaitGroup` mirroring the plugin's own pattern (`internal/tools/integrations/mcp/plugin.go:61-89`), per-server spawn+handshake bounded by `EffectiveTimeout()`, failures warn+skip per server, no overall Build bound. Caps first-`GetRegistry` latency at `max(Tᵢ)` + plugin discovery (30s) instead of ΣTᵢ.
- **Concurrency audit**: `clients` append already under `f.mu`; each `StdioClient` owns its own `exec.Cmd` + SDK client; the only unsafe field is `cachedToken`, eliminated by the pre-pass.

### 4. Consent / serial semantics — **MAINTAINER DECISION: trusted local spawns**

The open decision from #1394 is settled: **stdio servers are trusted by virtue of config trust — `consent=false` default (no user prompt), `serial=true` default (mutating local processes), with `REQUIRES_CONSENT: true` opting back in.** This deviates from #1394's preferred option and the grill round's recommendation (`consent=true` default); the maintainer chose the trusted-spawn alternative. Rationale: a `COMMAND`/`ARGS` config entry is an explicit, operator-authored grant of local execution — no different in trust class from the config's own `SELECTED_PROVIDER` — and prompting on every call would make "many local MCP services" tedious. `EffectiveRequiresConsent()` gains an explicit stdio branch (override wins, then `isStdio()` → false, else `!isReadOnly()`); `EffectiveSerial()` remains `!isReadOnly()` — stdio has no URL, so it already yields `true`. **The ADR must flag this decision and the deviation from the round's recommendation.**

**Known consequence (accepted):** a read-only local server (e.g. `fetch`) is forced serial forever — there is **no serial override for any transport today** (`EffectiveSerial()` is global; consent is the only overridable axis), so stdio is not made *more* configurable than HTTP. A per-server serial override is a **tracked ADR future option**: if ever added, it must apply to both transports or carry a justification for asymmetry.

### 5. Secrecy — ARGS/ENV never logged

- **Config loader**: add `args?|env` to `secretKeyPattern` (`internal/infrastructure/config/redact.go:22`). One regex change covers **both** debug dumps — the raw dump (`redactRawContent`, `config.go:147`) and the parsed-key dump (`config.go:157-163`, which is the *more* reliable leak: no 1000-byte truncation, `slog.Any` over the full ARGS slice/ENV map). `ARGS: [...]`/`ENV: {...}` lines collapse to `ARGS: [REDACTED]`/`ENV: [REDACTED]`, multi-line `ENV:` block maps drop wholesale. Fail-closed over-redaction is the accepted, documented trade (same as the existing `keys?` → "monkey" over-match). Re-pin the `max_tokens`/`max_history_tokens` exclusion so the addition doesn't disturb the token carve-out.
- **Adapter contract**: `NewStdioClient` error/warning text carries `cfg.Command` at most, **never** ARGS/ENV values (`exec.Error` only carries `Path`/`Name`).
- **Child stderr**: always-logged at Info with `mcp_server=<name>` — it is the wedge/death visibility window from §2; gating it would defeat the purpose. Documented in the ADR as **outside** `mcp-token-not-logged` — the invariant governs tell-me-go's own credential plumbing, not arbitrary child output.

### 6. Error surface — COMMAND resolution

On `errors.Is(err, exec.ErrNotFound)` from `exec.CommandContext`, annotate: `mcp stdio: command %q not found in tell-me-go's PATH (COMMAND resolves before ENV is applied; use an absolute path, a DIR-relative path, or ${VAR}-expanded COMMAND): %w`. Fail-point annotation only — **no** heuristic warning when `ENV` contains a PATH override (`ENV.PATH` is legitimately set for the child's own needs even with an absolute `COMMAND`).

### 7. Docs & model

- **README** MCP section: stdio example; one paragraph stating the resolution contract (§1) + venv example in absolute form (`COMMAND: "/venv/bin/uvx"` or `COMMAND: "${UVX_PATH}/uvx"`); a comment on each `npx`/`uvx` example: `# COMMAND is resolved via tell-me-go's PATH (parent process); use an absolute path or ${VAR} for venv/toolchain locations`.
- **ADR-067 amendment** (extend, don't add a new ADR): second transport implementation; lifecycle/zombie guarantees (two-tier, zombie-vs-orphan, SIGPIPE backstop, residuals); consent/serial defaults + the trusted-spawn maintainer decision and its deviation from the round's recommendation; the mode-conflict validation rule (bearer/basic under COMMAND); concurrent construction with per-server TIMEOUT bound and skip-for-session; the plugin's 30s discovery bound vs `TIMEOUT` nuance; stderr scope; resolution semantics; tracked future options (serial override; pgid-kill/Job Objects).
- **Domain model**: `MCPServer` definition amended (no longer "over Streamable HTTP" only) + `command`/`args`/`dir`/`env` attributes + stdio invariant (`mcp-stdio-trusted-defaults` or amendment clause to `mcp-readonly-defaults`); re-render the `.md` (hard `modelith-check` gate).
- **Catalog**: re-anchor the two `mcp_factory.go` pins (`:60-65`, `:157-160`) **in the same PR as the partition test** (`real_nonfix_catalog_test.go`) per the coordination rule.

## Test strategy

- **Config validation** (table): COMMAND/URL exclusivity (both → error, neither → error); ARGS/DIR/ENV-without-COMMAND → error; `AUTH: bearer`/`basic` + COMMAND → error; `AUTH: auto`/`gh`/`none`/empty + COMMAND → valid; key regex and TIMEOUT unchanged.
- **Redaction** (`redact_test.go`): `isSecretKey("mcp_servers.fs.args")`/`.env` → true; raw `ARGS: ["-p", "sk-1234"]` → `ARGS: [REDACTED]` (the isolating case — the `--token` case is already redacted today via word matching); `ARGS: ["-y", "@modelcontextprotocol/server-filesystem"]` → `ARGS: [REDACTED]` (fail-closed); `ENV:` block map → `ENV: [REDACTED]` + sub-keys dropped; `max_tokens`/`max_history_tokens` exclusion re-pinned.
- **Factory**: injected `newClientFor` fake asserts mode dispatch per config (stdio vs URL); resolver-not-called counter for stdio (`AUTH: gh` + COMMAND → resolver not invoked, no `mcp_token_resolution_skipped`, client still built); constructor-error text contains no ARGS element and no ENV value (Command yes, Args/Env never); warn+skip on constructor error; **concurrency test** — channel-gated fake (no `time.Sleep`, ADR-036): all N servers constructed, total Build time ≈ max delay not Σ, failing server warns+skips while others succeed, `gh` resolver invoked ≤ once across a multi-server HTTP config (single-flight pre-pass).
- **Integration** (`internal/infrastructure/mcp/`, SDK fixture server per the SDK's own `server.Run(ctx, &StdioTransport{})` pattern under the `go build`-helper harness — `internal/infrastructure/exec/main_test.go:23-42` precedent):
  - ListTools/CallTool round-trip; stderr routed to a captured slog handler, not parsed as JSON-RPC.
  - Handshake timeout kills a never-initializing child (deterministic via the timeout itself — no `time.Sleep`).
  - **Graceful EOF exit**: fixture exits on stdin EOF → `Close()` returns clean, reaper's Wait error not surfaced.
  - **EOF-ignoring backstop**: fixture that ignores EOF (real connected SDK server blocking on a channel instead of `server.Run`) → `Close()` still returns (cancel kills the direct child, reaper reaps, no hang; `signal: killed` suppressed).
  - **Launcher-tree pass-through**: synthetic launcher fixture (small Go program that `exec`s the SDK server with `Stdin: os.Stdin` inherited) → on `Close()`, both the launcher (reaped, Wait returns) and the grandchild (liveness poll with deadline) are gone. Pins the `npx`/`uvx` adoption claim without network flakiness.
  - **Child death mid-session**: server exits (e.g. via a `die` tool call) → next `CallTool` fails fast with the child-exit error, and a subsequent call fails fast with the *same* error (pins the pre-check, which the SDK alone doesn't guarantee).
  - **Wedge**: server reads stdin but never responds → `CallTool` returns `DeadlineExceeded` at the configured short TIMEOUT.
  - **Slow-but-alive**: server responds after longer than one call's timeout → next call succeeds (pins the no-kill policy).
  - **Two stdio clients** in one process operate independently (no shared `os.Stdin`/`os.Stdout`).
- **Executor integration** (optional): serial halt-on-failure — a failing serial stdio tool skips the remaining tools in that response (existing `ToolOptions.Serial` contract, `internal/agent/executor/executor.go:623-626`).
- Gate: `make check-full` green incl. `test-race`.

## Acceptance criteria

- [ ] `MCP_SERVERS.<name>.COMMAND` (+ `ARGS`/`DIR`/`ENV`) spawns a local stdio MCP server; its tools appear as `mcp_<name>_*` alongside remote MCP tools.
- [ ] Multiple stdio servers run as independent child processes in one session — one child each, no reliance on `os.Stdin`/`os.Stdout`; discovery and lifecycle are concurrent, tool *execution* remains serialized per existing `ToolOptions.Serial` semantics (a serial tool runs alone in its response and halts the remaining plan on failure).
- [ ] Spawn/connect/handshake failures and handshake timeout are clear, recoverable errors (warn+skip, session continues); per-call timeouts recoverable (call fails, session remains); **no zombie processes** (direct-child reap deterministic; tree best-effort via stdin EOF); `Close` idempotent; dead vs wedged child distinguishable in the error surface (fast child-exit error vs TIMEOUT).
- [ ] Config validation rejects ambiguous entries (COMMAND+URL, neither, ARGS/DIR/ENV without COMMAND, bearer/basic with COMMAND).
- [ ] `ARGS`/`ENV` values are redacted in config debug dumps; adapter errors never include ARGS/ENV values.
- [ ] Stdio servers default to **trusted** (`consent=false`, no prompt) with `REQUIRES_CONSENT: true` opt-in; `serial=true` default; no serial knob.
- [ ] README + ADR-067 amendment + domain model updated (modelith re-render committed; `modelith-check` green); `mcp_factory.go` catalog pins re-anchored in the same PR as the partition test; `make check-full` green.

## Prior analysis

- #1394 findings comment (corrections table): https://github.com/gosharplite/tell-me-go/issues/1394#issuecomment-5323595497
- Full grill transcript (public gist): https://gist.github.com/gosharplite/6b9643b718e717b0e92e46ab8113cdc1

## References

- ADR-067 — `docs/adr/2026-08-mcp-client-architecture.md` (remote-only, streamable HTTP; port isolation pattern)
- #1373 — Streamable HTTP MCP client (base of current implementation)
- #1388 / #1389 — Basic auth addition (precedent for extending `MCPServerConfig` + factory seam)
- #1394 — superseded issue (this issue is its post-grill implementation spec)
- go-sdk v1.7.0 — `mcp.Transport`, `mcp.IOTransport`, `mcp.StdioTransport`

## Out of scope

- MCP **server** support (tell-me-go serving its own tools over MCP).
- MCP resources/prompts — tools only, as today.
- OAuth dynamic client registration for remote servers.
- Per-server serial override and process-group kill / Job Objects (tracked ADR future options, not this PR).

gosharplite:
Closes #1396.

## Summary

Adds first-class **stdio / local-process MCP server support** to tell-me-go, alongside the existing Streamable HTTP transport (ADR-067). A new `MCP_SERVERS.<name>.COMMAND` (+ `ARGS`/`DIR`/`ENV`) config spawns the MCP server as a local child process, wired through the SDK's `IOTransport` (not the process-pinned `StdioTransport`). Per the maintainer decision in #1396, stdio servers default to **trusted local spawns**: `consent=false` (no prompt), `serial=true`, with `REQUIRES_CONSENT: true` opt-in — a documented deviation from the #1394 grill round's recommendation.

**Architect–Coder loop:** issue #1396 was implemented task-by-task by an Architect (design/review, read-only) and Coder (implementation, only writer) with per-increment review, the repo's full gate pipeline, and a catalog/whitelist adjudication. 10 commits.

## What changed

- **Config** — `MCPServerConfig` gains `COMMAND`/`ARGS`/`DIR`/`ENV`; validation: COMMAND/URL mutually exclusive (exactly one), ARGS/DIR/ENV require COMMAND, `AUTH` bearer/basic under COMMAND rejected (mode-conflict rule, before the positive credential rules); `EffectiveRequiresConsent()` stdio branch (trusted default).
- **Transport** — new `StdioClient` in `internal/infrastructure/mcp/` (separate type, not a mode on `Client`): eager construction with in-constructor handshake bounded by `TIMEOUT`, reaper goroutine (deterministic direct-child reaping), fast-death pre-check (sticky child-exit error), idempotent `Close` (stdin EOF → cancel backstop → reaper join), no kill on operation-timeout, stderr routed to slog (never parsed as JSON-RPC), `exec.ErrNotFound` annotated with the resolution contract.
- **DI** — `newClientFor(serverCfg)` dispatch seam; two-phase `Build` (sequential token pre-pass with stdio short-circuit + single-flight `gh` memoization, then concurrent construction via `sync.WaitGroup`); first-`GetRegistry` latency capped at `max(Tᵢ)` instead of ΣTᵢ.
- **Secrecy** — `args?|env` added to `secretKeyPattern` (covers both config debug dumps incl. ENV sub-keys via ancestor-skip); adapter errors never include ARGS/ENV values.
- **Tests** — 9 integration tests against a real SDK fixture server (round-trip, handshake-timeout kill, graceful EOF close, EOF-ignoring backstop, launcher-tree pass-through, child death mid-session, wedge timeout, slow-but-alive no-kill, two independent clients), plus factory concurrency/redaction/validation suites.
- **Docs** — ADR-067 amendment (10 items incl. the trusted-spawn decision, two-tier lifecycle guarantees, resolution semantics); domain model (`MCPServer` attributes + `mcp-stdio-trusted-defaults` invariant + scenario, re-rendered); README stdio example + resolution-contract paragraph; catalog re-anchors + 5 new pins (Complexity/Test-Complexity); ADR-056 whitelist extension for the mcp consumer closure.

## Acceptance criteria mapping

- [x] `MCP_SERVERS.<name>.COMMAND` (+ `ARGS`/`DIR`/`ENV`) spawns a local stdio MCP server; tools appear as `mcp_<name>_*` alongside remote MCP tools — `StdioClient` (611665bf), integration suite (efea6946), pre-existing plugin namespacing.
- [x] Multiple stdio servers run as independent child processes in one session — one child each, no shared `os.Stdin`/`os.Stdout`; concurrent discovery/lifecycle; tool *execution* serialized per existing `ToolOptions.Serial` semantics — integration test 9, factory concurrency test, `EffectiveSerial` rows, pre-existing `TestDispatcher_SerialBatching`.
- [x] Spawn/connect/handshake failures and handshake timeout are clear, recoverable errors (warn+skip); per-call timeouts recoverable; **no zombie processes** (deterministic direct-child reap, tree best-effort via stdin EOF); `Close` idempotent; dead vs wedged distinguishable (fast child-exit error vs TIMEOUT) — integration tests 2-7.
- [x] Config validation rejects ambiguous entries (COMMAND+URL, neither, ARGS/DIR/ENV without COMMAND, bearer/basic with COMMAND) — config validation suite (afa77c1d).
- [x] `ARGS`/`ENV` values redacted in config debug dumps; adapter errors never include them — redaction suite (51366da7) + factory sentinel test.
- [x] Stdio servers default to **trusted** (`consent=false`, no prompt) with `REQUIRES_CONSENT: true` opt-in; `serial=true`; no serial knob — config + factory tests + ADR-067 amendment.
- [x] README + ADR-067 amendment + domain model updated (modelith re-render committed; `modelith-check` green); `mcp_factory.go` catalog pins re-anchored same-PR as the partition test; `make check-full` green.

## Verification

| Check | Status |
|-------|--------|
| `make check-full` (fmt, tidy, build, lint, verify-architecture, verify-transitive-gate, verify-nonfix-catalog, verify-ports-registry, verify-mcp-sdk-confinement, verify-no-test-sleep, verify-adr-index, modelith-check, fuzz-smoke, vulncheck, test, dead-code, test-coverage, test-race) | **PASS** — "All checks passed (including race detection)" |
| `go test -race` on the new packages | PASS |
| `make modelith-lint` / `modelith-check` | PASS (0 errors/warnings) |
| `make verify-nonfix-catalog` | PASS (all seven gate lines) |
| `make verify-transitive-gate` | PASS (80 approved constants, 0 decision-required) |

## Notes

- The two `mcp_factory.go` catalog pins were re-anchored and the 5 new over-threshold functions (the `validate` rework + 4 stdio/factory tests) were cataloged in the same commit as the partition test (coordination rule).
- The `internal/infrastructure/mcp` ADR-056 whitelist entry was added because the `StdioClient` constructor legitimately consumes the domain `config` type — its closure matches the already-approved `internal/infrastructure/config` adapter exactly.
- Tracked ADR future options (not this PR): per-server serial override (both transports or justified asymmetry); process-group kill / Job Objects for transport-contract-violating grandchildren; `validate` extraction refactor (CC=20, cataloged).

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ A high-performance CLI assistant unifying the world's most powerful reasoning en
     *   **System & Dev**: Shell execution (`execute_command`, `pipe_commands`), testing, linting, and vulnerability scanning (`govulncheck`).
     *   **State & History**: Task tracking, `summarize_history`, `manage_history` (pinning/unpinning turns to protect them from pruning), and **Interactive History Browser** (`tell-me-go browse`) with full-text search and O(1) archive navigation.
     *   **Media**: Imagen 3 image generation and Vision analysis.
-    *   **Model Context Protocol (MCP)**: Native consumption of remote MCP servers (GitHub Copilot, custom endpoints) over streamable HTTP with automatic credential resolution.
+    *   **Model Context Protocol (MCP)**: Native consumption of MCP servers — remote endpoints over streamable HTTP with automatic credential resolution, and local stdio servers (COMMAND).
 *   **Safety Guardrails**: 
     *   **Context Control**: Automatic "self-healing" summarization and turn pinning to prevent overflow without losing intent.
     *   **Runaway Protection**: Hallucination loop detection (SHA-256 response hashing), tool repetition guards, and recursion limits.
@@ -346,6 +346,37 @@ MCP_SERVERS:
     # TOKEN: "${GITHUB_TOKEN}"  # Optional: auto-resolved via gh auth token if omitted
     # TIMEOUT: 300              # Seconds (default: 300)
     # REQUIRES_CONSENT: false   # Optional override (default: false for /readonly)
+```
+
+Local stdio servers spawn a child process and talk to it over stdin/stdout:
+
+```yaml
+# COMMAND is resolved via tell-me-go's PATH (parent process); use an absolute path or ${VAR} for venv/toolchain locations.
+MCP_SERVERS:
+  fs:
+    COMMAND: "npx"
+    ARGS: ["-y", "@modelcontextprotocol/server-filesystem", "/path/to/dir"]
+  git:
+    COMMAND: "uvx"
+    ARGS: ["mcp-server-git"]
+```
+
+A bare `COMMAND` (no path separator) is resolved via `exec.LookPath` in
+tell-me-go's own process PATH — not `ENV.PATH`, not `DIR`; a separator-bearing
+`COMMAND` (`/abs`, `./rel`) is used as-is, with relative paths resolved
+against `DIR`. `ENV.PATH` governs the child only after exec (it does reach
+the server's own subprocess resolution, e.g. uvx finding `mcp-server-git`).
+`${VAR}` is expanded at config load in `COMMAND`, `ARGS`, `DIR`, and `ENV`
+values.
+
+For executables inside a venv or toolchain directory, use an absolute or
+`${VAR}`-expanded `COMMAND`:
+
+```yaml
+# Absolute path or ${VAR}-expanded COMMAND for venv/toolchain locations:
+venv-tools:
+  COMMAND: "${UVX_PATH}/uvx"
+  ARGS: ["mcp-server-git"]
 ```
 
 ## ⌨️ Shell Integration (Recommended)
--- a/docs/adr/2026-08-mcp-client-architecture.md
+++ b/docs/adr/2026-08-mcp-client-architecture.md
@@ -273,6 +273,150 @@ subset of configuration the per-turn config watcher manages.
    in-process code and the alternative (provenance tracking) adds registry
    coupling disproportionate to the threat model.
 
+### Amendment (2026-08, issue #1396 — stdio transport for local MCP servers)
+
+1. **Second transport implementation.** tell-me-go now supports local stdio
+   MCP servers via a separate `StdioClient` type
+   (`internal/infrastructure/mcp/stdio_client.go`) in the same package — not
+   a mode on the HTTP `Client`. Eager construction: `NewStdioClient(cfg,
+   logger)` spawns `COMMAND` with `ARGS`/`DIR`/`ENV` via
+   `exec.CommandContext`, connects over
+   `sdkmcp.IOTransport{Reader: stdoutPipe, Writer: stdinPipe}` inside the
+   constructor, bounded by the server's `TIMEOUT` (default 300s). One child
+   process per server; no reliance on `os.Stdin`/`os.Stdout` (the SDK's
+   `StdioTransport` is hard-wired to those, one per process, hence rejected).
+   Reuses the HTTP client's unexported conversion helpers; SDK confinement
+   (§2) unchanged.
+
+2. **Lifecycle & process-tree guarantees (two-tier).** Direct-child reaping
+   is **deterministic**: a reaper goroutine started immediately after
+   `cmd.Start()` sends `cmd.Wait()`'s result to a buffered channel, so the
+   child is reaped the moment it dies — zombie prevention is not contingent
+   on `Close()`. Tree termination is **best-effort** via shared-pipe stdin
+   EOF: launchers (`npx`, `uvx`) pass stdio through rather than proxying it,
+   so a grandchild sees EOF directly and SDK-built servers
+   (`server.Run(ctx, &StdioTransport{})` read loop exits on `io.EOF`)
+   self-terminate; the Unix SIGPIPE write-backstop covers a grandchild that
+   writes after our read end closes. **Zombie vs orphan**: no zombies,
+   guaranteed; orphans possible for transport-contract-violating servers (an
+   EOF-ignoring **and silent** grandchild on Unix; any EOF-ignoring
+   grandchild on Windows — no signal backstop) — accepted residuals. **No
+   process-group kill this PR** (Unix-only platform split for a
+   misbehaving-server edge; pgid-kill is wrong for a reparenting child;
+   Windows needs Job Objects) — tracked future option.
+
+3. **Fast-death and failure surfaces.** A non-blocking pre-check at the top
+   of `ListTools`/`CallTool` (under the client mutex) returns a sticky
+   `mcp: stdio child %q exited: %w` error once the child has died — every
+   subsequent call fails fast deterministically (the SDK alone does not
+   guarantee this). In-flight EOF (`io.EOF`/`ErrConnectionClosed`)
+   coinciding with a confirmed child exit is annotated with the exit status;
+   a wedge surfaces as the plain `context.DeadlineExceeded` wrap — dead vs
+   wedged are distinguishable by error text and latency.
+   Spawn/connect/handshake failures and handshake timeout surface at DI
+   `Build` as the existing warn+skip (server skipped for the session;
+   `MCP_SERVERS` read once, §10 unchanged); per-call timeouts are
+   recoverable (call fails, session remains); **no kill on
+   operation-timeout** — a slow legitimate call (e.g. a 200s filesystem op
+   under TIMEOUT 300) is indistinguishable from a wedge and does not poison
+   the session; documented known limitation, child stderr is the operator's
+   visibility window.
+
+4. **Close semantics.** `Close()` is idempotent and mutex-guarded, in this
+   order: (1) `session.Close()` first — closes both pipes, sending stdin EOF
+   so a well-behaved child (and its tree) exits gracefully; (2) `cancel()` as
+   the kill backstop for uncooperative children (called explicitly and
+   deferred so it fires even on an error unwinding `session.Close`);
+   (3) join the reaper (expected `signal: killed`/`context.Canceled` logged
+   at debug, not surfaced). `factory.Close()` iterates tracked clients, so
+   session teardown kills every child.
+
+5. **Consent/serial defaults — trusted local spawns (MAINTAINER DECISION).**
+   Stdio servers are trusted by virtue of config trust: `consent=false`
+   default (no user prompt), `serial=true` default (mutating local
+   processes), with `REQUIRES_CONSENT: true` opting back in. **This deviates
+   from the #1394 grill round's recommendation (`consent=true` default)** —
+   flag the deviation explicitly and record the maintainer's rationale: a
+   `COMMAND`/`ARGS` config entry is an explicit, operator-authored grant of
+   local execution — no different in trust class from the config's own
+   `SELECTED_PROVIDER` — and prompting on every call would make "many local
+   MCP services" tedious. `EffectiveRequiresConsent()` gains an explicit
+   stdio branch (override wins, then `IsStdio()` → false, else
+   `!isReadOnly()`); `EffectiveSerial()` remains `!isReadOnly()` — stdio has
+   no URL, so it already yields `true`. **Known consequence (accepted):** a
+   read-only local server (e.g. `fetch`) is forced serial forever — there is
+   no serial override for any transport today (`EffectiveSerial()` is
+   global; consent is the only overridable axis), so stdio is not made
+   *more* configurable than HTTP. A per-server serial override is a tracked
+   future option: if ever added, it must apply to both transports or carry a
+   justification for asymmetry.
+
+6. **Mode-conflict validation rule.** `validate()` rejects
+   `AUTH: bearer`/`basic` together with `COMMAND` ("stdio (COMMAND) servers
+   transmit no credentials"), checked before the positive credential rules
+   so the mode-conflict message wins. This is a mode-conflict rule, same
+   family as COMMAND/URL mutual exclusivity — it does not disturb §4's
+   positive-rule-only credential philosophy; `auto`/`gh`/`none`/empty under
+   stdio are accepted (inert under the stdio short-circuit, which also means
+   `gh`/`auto` under stdio never resolve and never invoke the resolver), and
+   stray `TOKEN`/`USERNAME` under stdio remain tolerated (§4 stray-field
+   tolerance). COMMAND and URL are mutually exclusive — exactly one must be
+   set.
+
+7. **Concurrent construction.** The DI factory
+   (`internal/infrastructure/di/mcp_factory.go`) builds clients in two
+   phases: a sequential token pre-pass (stdio short-circuits first;
+   `gh`/`auto` under stdio never resolve; HTTP servers resolve via the
+   single-flight `gh`-memoized resolver, credentials stamped into per-server
+   copies), then concurrent construction per server (`sync.WaitGroup`,
+   per-server spawn+handshake bounded by `TIMEOUT`), failures warn+skip per
+   server, no overall `Build` bound. First-`GetRegistry` latency caps at
+   `max(Tᵢ)` + plugin discovery rather than ΣTᵢ. **Nuance:** the plugin's
+   own 30s per-server discovery bound (`defaultDiscoveryTimeout`) is separate
+   from and additional to `TIMEOUT` — a server whose handshake took the full
+   `TIMEOUT` still gets its own 30s `ListTools` window.
+
+8. **Stderr scope.** Child stderr is always logged at Info with
+   `mcp_server=<name>`, line-buffered, and never parsed as JSON-RPC. This is
+   **outside** the `mcp-token-not-logged` invariant (the invariant governs
+   tell-me-go's own credential plumbing — config validation, DI factory,
+   client transport — not arbitrary child output); a misbehaving child that
+   echoes secrets to stderr is an accepted residual of the wedge/death
+   visibility window. The domain model invariant statement is unchanged.
+
+9. **Command resolution semantics.** Bare `COMMAND` (no path separator)
+   resolves via `exec.LookPath` in **tell-me-go's own process PATH** — not
+   `ENV.PATH`, not `DIR`; separator-bearing `COMMAND` (`/abs`, `./rel`) is
+   used as-is, relative resolved against `DIR`; `ENV.PATH` governs the child
+   only post-exec (it reaches the server's own subprocess resolution — e.g.
+   uvx finding `mcp-server-git` — which is why it is set and why failures
+   are confusing). `${VAR}` expands at config load via `expandEnvHook` in
+   `COMMAND`, `ARGS`, `DIR`, and `ENV` values. Windows bare-name lookup
+   honors `PATHEXT` via the same stdlib path. On `exec.ErrNotFound`, the
+   error is annotated with the resolution contract (fail-point only; no
+   heuristic warning when `ENV` contains a PATH override).
+
+10. **Tracked future options** (explicitly NOT this PR): (a) per-server
+    serial override, applying to both transports or with a justified
+    asymmetry; (b) process-group kill (Unix pgid) / Job Objects (Windows)
+    for transport-contract-violating grandchildren; (c) extraction refactor
+    of `(*MCPServerConfig).validate` (CC=20, cataloged as structural guards
+    in `INTENTIONAL_NON_FIXES.md`).
+
+### Consequences
+
+Stdio support broadens the MCP surface from remote-only to local processes
+while preserving the SDK-confined, port-based architecture: a separate
+`StdioClient` type keeps the HTTP `Client` untouched, and the two-tier
+lifecycle (deterministic direct-child reaping + best-effort tree termination
+via shared-pipe EOF) prevents zombies with accepted orphan residuals for
+contract-violating servers. The trusted-spawn consent default is a deliberate
+deviation from the #1394 grill recommendation, justified by the
+operator-authored trust class of `COMMAND`/`ARGS`; the forced-serial
+consequence for read-only local servers and the no-kill-on-operation-timeout
+policy are documented known limitations with tracked future options rather
+than silent trade-offs.
+
 ## Consequences
 
 **Positive:** the SDK is fully isolated behind `tools.MCPClient`, so the
--- a/docs/architect/INTENTIONAL_NON_FIXES.md
+++ b/docs/architect/INTENTIONAL_NON_FIXES.md
@@ -894,6 +894,12 @@ catalog a new gap no one reviewed. Policy:
 - **Rationale**: Sequential launcher test (spawn → stderr poll for pid → ListTools → Close → deadline-bounded liveness poll with `syscall.Kill`). CC is the poll/assertion sequence, not branching business logic. Splitting would duplicate the child-process/launcher setup. Same acceptance class as `TestHistoryNavigation_CompleteWorkflow` (CC=15) — sequential workflow where steps depend on prior state.
 - **See**: `internal/infrastructure/mcp/stdio_client_integration_test.go:216`
 
+### internal/infrastructure/mcp/stdio_client_integration_test.go — TestStdio_ChildDeathMidSession (CC=11)
+
+- **Status**: ACCEPTED (2026-08, #1396)
+- **Rationale**: Sequential death-mid-session integration test over a real child process (die call → race-tolerant death-indication assertion → deadline-bounded poll loop that deterministically closes the async-reap window by waiting for the sticky child-exit error → post-sticky pointer-equality assertions). CC comes from the poll/branch structure (the poll's if/break, the deadline guard, and the two sticky assertions), not branching business logic — and the poll is the test's synchronization mechanism, not refactorable: splitting it would both duplicate the expensive child-process setup and destroy the deterministic reaper synchronization it provides. Same acceptance class as `TestStdio_RoundTrip` (CC=11) and `TestStdio_LauncherTreePassThrough` (CC=13) in the same file.
+- **See**: `internal/infrastructure/mcp/stdio_client_integration_test.go:278`
+
 ---
 
 ## Complexity Alerts (ACCEPTED)
--- a/docs/architect/INTENTIONAL_NON_FIXES.md
+++ b/docs/architect/INTENTIONAL_NON_FIXES.md
@@ -870,6 +870,30 @@ catalog a new gap no one reviewed. Policy:
 - **Rationale**: Table-driven transport test pinning the Basic auth header contract across three subtest closures (header set with exact base64, no-header when username empty, no-header when token empty), each building its own roundTripFunc assertion closure. CC comes from subtest enumeration and assertion boilerplate, not branching business logic. Same acceptance class as `TestMediaBlocks` (CC=11) — assertion boilerplate across a coverage matrix.
 - **See**: `internal/infrastructure/mcp/client_test.go:645`
 
+### internal/infrastructure/di/mcp_factory_test.go — TestMCPFactory_ConcurrentBuild (CC=11)
+
+- **Status**: ACCEPTED (2026-08, #1396)
+- **Rationale**: Channel-gated concurrency proof for the factory's phase-2 construction: per-server fake calls block on release channels while a started-channel deadline select proves all N constructions entered before any completed. CC comes from harness/assertion boilerplate (deadline select, started-counter check, per-server dep assertions), not branching business logic. Splitting would duplicate the channel-gated harness. Same acceptance class as `TestGetModelTurn` (CC=16) and `TestFakeToolchainRunner_PresetValues` (CC=28).
+- **See**: `internal/infrastructure/di/mcp_factory_test.go:683`
+
+### internal/infrastructure/di/mcp_factory_test.go — TestMCPFactory_ConcurrentBuild_FailingServerSkips (CC=15)
+
+- **Status**: ACCEPTED (2026-08, #1396)
+- **Rationale**: Same channel-gated harness as `TestMCPFactory_ConcurrentBuild` plus sequential setup and per-server assertions over N servers (failing server absent, others present, exactly one `mcp_client_init_failed` log). CC is harness/assertion boilerplate, not branching business logic. Splitting would duplicate the expensive harness. Same acceptance class as `TestFakeToolchainRunner_PresetValues` (CC=28).
+- **See**: `internal/infrastructure/di/mcp_factory_test.go:743`
+
+### internal/infrastructure/mcp/stdio_client_integration_test.go — TestStdio_RoundTrip (CC=11)
+
+- **Status**: ACCEPTED (2026-08, #1396)
+- **Rationale**: Sequential integration test over a real child process (list tools → echo round-trip → stderr-capture poll → stderr tool call). CC is multi-step assertion boilerplate and the deadline-bounded stderr poll, not branching business logic. Splitting would duplicate the expensive child-process setup per subtest. Same acceptance class as `TestGetModelTurn` (CC=16) and `TestHistoryNavigation_CompleteWorkflow` (CC=15).
+- **See**: `internal/infrastructure/mcp/stdio_client_integration_test.go:112`
+
+### internal/infrastructure/mcp/stdio_client_integration_test.go — TestStdio_LauncherTreePassThrough (CC=13)
+
+- **Status**: ACCEPTED (2026-08, #1396)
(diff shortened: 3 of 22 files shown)


gosharplite:
## Architectural Review: PR #1397 — Stdio / Local-Process MCP Server Support (#1396)

### Executive Summary

PR #1397 implements first-class support for local stdio MCP child processes, fulfilling the post-grill implementation spec in #1396. The implementation adheres closely to the project's Hexagonal Architecture principles:
1. **Domain Port Isolation**: Confines `github.com/modelcontextprotocol/go-sdk` strictly to `internal/infrastructure/mcp/` behind the transport-agnostic `tools.MCPClient` domain port.
2. **Two-Phase DI Construction**: Decouples sequential credential resolution (with single-flight `gh` token memoization) from concurrent per-server spawn and handshake in `mcpFactory.Build()`, capping initialization latency at $\max(T_i)$ rather than $\sum T_i$.
3. **Two-Tier Child Lifecycle**: Direct child reaping is deterministic via dedicated background reaper goroutines (preventing zombie processes), while process-tree teardown operates via shared-pipe stdin EOF propagation and cancel backstops.
4. **Security & Redaction**: Extends `secretKeyPattern` to redact `ARGS` and `ENV` blocks and sub-keys in config debug dumps, while ensuring child stderr remains an unparsed diagnostics stream.
5. **Architectural Governance**: Domain model, ADR-067 amendment, `TRANSITIVE_IMPORT_WHITELIST.md`, and `INTENTIONAL_NON_FIXES.md` catalog coordination rules are fully maintained.

However, a **blocking race condition** was identified during `make check-full` execution in `TestStdio_ChildDeathMidSession`.

---

### Findings

#### [ARCHITECTURAL BLOCKER] Flaky Integration Test & Error Stickiness Race in `TestStdio_ChildDeathMidSession`

- **Locations**: `internal/infrastructure/mcp/stdio_client.go:210-221, 258-278` & `internal/infrastructure/mcp/stdio_client_integration_test.go:283-310`
- **Impact**: Fails `make check-full` / `go test -count=10` on `internal/infrastructure/mcp`.
- **Root Cause**:
  1. When `CallTool(ctx, "die", nil)` executes, the test child calls `os.Exit(0)`, closing the stdio pipe (EOF).
  2. The background reaper goroutine (`go func() { waitDone <- cmd.Wait() }()`) runs asynchronously to reap the process exit via the OS kernel.
  3. When `CallTool(ctx, "echo", ...)` is called immediately (call 2):
     - If `cmd.Wait()` has **not yet completed**, `c.childExitErrLocked()` returns `nil`. The SDK session fails immediately with `connection closed: calling "tools/call": client is closing: EOF`.
     - In `annotateIfDead(err)`: if `cmd.Wait()` has still not completed, call 2 returns the raw SDK error without the child-exit wrap (`stdio_client_integration_test.go:300` failure).
     - If `cmd.Wait()` finishes before `annotateIfDead(err)` returns, `annotateIfDead` creates a *newly allocated* error: `fmt.Errorf("%w: %v", childErr, err)`.
  4. On call 3, `childExitErrLocked()` at the start of `CallTool` catches `c.exitErr` and directly returns `childErr` (the original pointer).
  5. The test assertion `if err3 != err2 { t.Error(...) }` checks pointer equality. Because `err2` was the wrapped error returned by `annotateIfDead` while `err3` is `c.exitErr`, `err3 != err2` fails on pointer inequality (`stdio_client_integration_test.go:307` failure).

- **Current vs. Scalable Design**:

  *Current (racy timing & pointer mismatch)*:
  ```go
  // stdio_client.go
  func (c *StdioClient) annotateIfDead(err error) error {
      ...
      if errors.Is(err, io.EOF) || errors.Is(err, sdkmcp.ErrConnectionClosed) {
          return fmt.Errorf("%w: %v", childErr, err) // returns a newly wrapped error, distinct from c.exitErr
      }
      return err
  }
  ```
  ```go
  // stdio_client_integration_test.go
  _, err2 := c.CallTool(context.Background(), "echo", map[string]interface{}{"text": "x"})
  ...
  _, err3 := c.CallTool(context.Background(), "echo", map[string]interface{}{"text": "x"})
  if err3 != err2 { // fails when err2 was from annotateIfDead or unannotated
      t.Error("third call error != second call error; expected the identical sticky error value")
  }
  ```

  *Scalable (deterministic exit state & pointer consistency)*:
  In `TestStdio_ChildDeathMidSession`:
  - Allow the reaper to finish or make the sticky error assertion between subsequent calls that hit the pre-check once established (`err3 == err4`), OR in `annotateIfDead`, ensure that once `childErr` is established, the sticky error representation is consistently recorded and asserted.
  - Specifically in the test:
    ```go
    // Call 1 triggers child death
    _, err1 := c.CallTool(context.Background(), "die", nil)
    ...
    // Verify subsequent call fails with child exit indication
    _, err2 := c.CallTool(context.Background(), "echo", map[string]interface{}{"text": "x"})
    ...
    // Once sticky pre-check is active, consecutive calls return identical pointer value
    _, err3 := c.CallTool(context.Background(), "echo", map[string]interface{}{"text": "x"})
    _, err4 := c.CallTool(context.Background(), "echo", map[string]interface{}{"text": "x"})
    if err4 != err3 {
        t.Error("err4 != err3; expected identical sticky error value")
    }
    ```

---

#### [REFACTOR] Tracked Future Enhancements (Advisory / Non-Blocking)

1. **`(*MCPServerConfig).validate` Structural Complexity (CC=20)**:
   - Accurately cataloged in `docs/architect/INTENTIONAL_NON_FIXES.md` and ADR-067 §10 as a message-pinned sequence. An extraction into helper functions (e.g., `validateStdioRules`, `validateAuthModeRules`) can be performed in a future refactor.
2. **Serial Override Parity**:
   - Stdio servers default to `serial=true`. As noted in ADR-067 §5, adding a per-server `SERIAL: true|false` config knob remains a tracked future option across both HTTP and stdio transports.

---

### Instructions for the Coder

1. Update `TestStdio_ChildDeathMidSession` in `internal/infrastructure/mcp/stdio_client_integration_test.go` to handle the asynchronous process termination window and ensure the sticky error assertion verifies pointer equality on stable post-death calls.
2. Verify `go test -race -count=50 -run TestStdio_ChildDeathMidSession ./internal/infrastructure/mcp` passes cleanly without flakes.
3. Run `make check-full` to confirm all quality gates, linters, architecture tests, and race suites pass.

gosharplite:
## ✅ Review findings addressed — fix chain landed, gates green

All three post-review items from the architectural review are resolved and pushed to this branch:

### 1. 🔴 [ARCHITECTURAL BLOCKER] — Flaky `TestStdio_ChildDeathMidSession` race — **FIXED** (`c97ed7dc`)

The reviewer was correct, and the Architect's adjudication went one step further than the review sketch: the race is real in **both** current assertions (`err2` "contains exited" and `err3 == err2`), and the reviewer's own `err3 == err4` sketch is still theoretically racy (it only shrinks the async-reap window). The fix eliminates it by construction:

- **Production consistency** — `annotateIfDead` now returns the **sticky** `c.exitErr` directly (instead of a fresh `fmt.Errorf("%w: %v", childErr, err)` wrap), so an error containing `"exited"` is uniquely the sticky pointer. This is the reviewer's own "Scalable" alternative.
- **Nil-guard** — `childExitErrLocked` no longer wraps a clean-exit (`os.Exit(0)`) `cmd.Wait()` nil into `…exited: %!w(<nil>)`.
- **Deterministic test** — the reworked test polls the production pre-check until the sticky `"exited"` error appears (deadline-bounded, `runtime.Gosched()`, no `time.Sleep`), deterministically closing the async-reap window, then asserts pointer equality on stable post-death calls (`errA == sticky`, `errB == errA`).

Verification: `go test -race -count=50 -run TestStdio_ChildDeathMidSession ./internal/infrastructure/mcp` — **zero flakes**; `golangci-lint run ./...` — 0 issues.

### 2. 🟡 Test-complexity catalog pin — **FIXED** (`a1ddd53a`)

The reworked test crossed the CC>10 threshold (CC=11 at line 278, per the gate's own gatherer). Pinned as a `test-complexity` entry in `INTENTIONAL_NON_FIXES.md` + `expectedCataloged` row in the partition test, one commit (coordination rule). `verify-nonfix-catalog` 7/7 PASS.

### 3. 🟡 Already-fixed earlier in the chain

- errcheck lint (2 sites) — `f15aebed`
- ADR-056 transitive-closure whitelist (mcp consumer closure) — `c5670d8f`

## Updated verification table

| Check | Status |
|-------|--------|
| `make check-full` (full 20-gate pipeline incl. `test-race`) | **PASS** — "All checks passed (including race detection)" |
| `go test -race -count=50` on `TestStdio_ChildDeathMidSession` | PASS — zero flakes |
| `make verify-nonfix-catalog` | PASS (7/7 lines) |
| `make verify-transitive-gate` | PASS (80 approved, 0 decision-required) |
| `make modelith-check` / `modelith-lint` | PASS |
| `golangci-lint run ./...` | 0 issues |

Branch updated: `c5670d8f..a1ddd53a` (2 new commits). Ready for re-review.

gosharplite:
## Architectural Sign-Off: PR #1397 — Stdio / Local-Process MCP Server Support (#1396)

### Status: Approved & Verified (All Gates Green)

Following the fix chain in `c97ed7dc..a1ddd53a`, the implementation has been independently re-verified against the repository's full quality pipeline and the architectural specification in #1396.

---

### Verification Matrix

| Verification Gate | Command | Result |
|---|---|---|
| **Race-Detection Stress Test** | `go test -race -count=50 -run TestStdio_ChildDeathMidSession ./internal/infrastructure/mcp` | **PASS** (0 flakes across 50 runs, 2.91s) |
| **Full Quality Pipeline** | `make check-full` (20 gates) | **PASS** ("All checks passed (including race detection)") |
| **Hexagonal Layer Discipline** | `make verify-architecture` / `make modelith-layers` | **PASS** (0 layer boundary violations) |
| **Transitive Closure Gate** | `make verify-transitive-gate` | **PASS** (80 approved constants, 0 decision-required) |
| **Catalog & Complexity Gate** | `make verify-nonfix-catalog` | **PASS** (7/7 gate lines PASS) |
| **MCP SDK Confinement** | `make verify-mcp-sdk-confinement` | **PASS** (`go-sdk` strictly confined to `internal/infrastructure/mcp/`) |
| **Domain Model Consistency** | `make modelith-check` / `make modelith-lint` | **PASS** (`tell-me-go.modelith.md` up to date) |
| **Static Analysis** | `golangci-lint run ./...` | **PASS** (0 findings) |

---

### Architectural Assessment & Resolution Confirmation

1. **🔴 [ARCHITECTURAL BLOCKER] Resolved (`c97ed7dc`)**:
   - **Sticky Error Guarantee**: `annotateIfDead` now returns the sticky `c.exitErr` pointer directly, eliminating transient wrapper allocations and guaranteeing consistent pointer-equality behavior across in-flight failure annotations and fast-death pre-checks.
   - **Clean-Exit Normalization**: `childExitErrLocked` guards clean exits (`os.Exit(0)`) against invalid `%!w(<nil>)` error wrapping.
   - **Deterministic Test Synchronization**: `TestStdio_ChildDeathMidSession` polls the production pre-check with `runtime.Gosched()` (bounded by a 5s deadline, no `time.Sleep` per ADR-036), eliminating the asynchronous OS process-reaping race by construction.

2. **🟡 Catalog Coordination (`a1ddd53a`)**:
   - Accurately pinned `TestStdio_ChildDeathMidSession` (CC=11) in `docs/architect/INTENTIONAL_NON_FIXES.md` under the `test-complexity` acceptance class, coordinated with the partition test in `real_nonfix_catalog_test.go` in the same commit.

3. **Hexagonal Architecture & Clean Separation of Concerns**:
   - `github.com/modelcontextprotocol/go-sdk` is strictly isolated behind `internal/domain/tools/mcp_client.go`.
   - The two-phase `mcpFactory.Build` pattern cleanly separates sequential token resolution (with single-flight `gh` memoization) from concurrent client construction (`sync.WaitGroup`), bounding startup time to $\max(T_i)$ and eliminating data races.
   - Security redaction (`args?|env` in `secretKeyPattern`) successfully prevents sensitive arguments and environment variables from leaking into debug dumps.
   - Maintainer decisions regarding trusted local spawns (`consent=false`, `serial=true` defaults) and lifecycle bounds are comprehensively documented in ADR-067 and the domain model.

---

### Conclusion

All architectural requirements, quality gates, and concurrency constraints are fully satisfied. PR #1397 is clean, robust, and ready for merge into `dev`.

PR #1442, opened 2026-08-28
fix(agent): Fix Gemini vertex API rejection of mixed media responses (#1441)

### LINKED ISSUE #1441: Bug: Gemini Vertex API throws Error 400 on mixed media tool responses (e.g., read_image)
### Context
When using the `read_image` tool (or any tool returning both binary media and text) with Vertex AI / Gemini models (e.g., `gemini-3.7-flash`, `gemini-3.1-pro-preview`), the agent loop fails immediately after tool execution.

### Symptom
The tool executes successfully, but the subsequent inference cycle crashes with:
```
Error 400, Message: Requests ending with a model turn are not supported., Status: INVALID_ARGUMENT
```

### Investigation
By injecting debug logging to capture the exact JSON payload sent to the Gemini `GenerateContent` endpoint, we observed the structure of the history array. 

When a tool result contains both text and binary data, `tell-me-go` bundles them into a single `user` turn. However, `AssembleResponse` (in `internal/agent/executor/executor.go`) was appending the `FunctionResponse` block *before* the `InlineData` block in the parts array:
```json
{
  "role": "user",
  "parts": [
    {
      "functionResponse": { ... }
    },
    {
      "inlineData": { ... }
    }
  ]
}
```
**The Gemini Quirk**: The Google Gemini API via Vertex AI has a strict, undocumented parser quirk. If it encounters a `user` turn where a `FunctionResponse` block precedes a media block (`InlineData`), the API silently invalidates or strips that entire user message. 

Because the `user` turn is stripped, the API evaluates the conversation history as ending at the *previous* turn (the `model` turn that made the `FunctionCall`). The API then throws the `400 Bad Request` error because a `model` turn must always be followed by a `user` turn.

### Solution
The fix involves reversing the appending order in the executor's `AssembleResponse` method:

1. Modified `internal/agent/executor/executor.go` so that `BinaryData` (`InlineData`) is appended to the message array **before** the `FunctionResponse` block.
2. Updated the corresponding integration tests (e.g., `TestDispatcher_AssembleResponse_Binary` in `tests/integration/agent/executor/executor_table_test.go`) to expect and assert the new part ordering.
3. Extracted anonymous test closures in `tests/integration/agent/regression_test.go` to named functions to keep the cyclomatic complexity (CC) under the strict repo limits (CC=10).

With `InlineData` coming first, the Gemini API correctly validates the message structure, recognizes it as a valid answer to the tool call, and seamlessly continues the multi-turn session.

The fix has been applied to the `dev` branch.

gosharplite:
Resolves #1441

### Description
When a tool result contains both a `FunctionResponse` and `InlineData` (e.g. from `read_image`), the Gemini API (Vertex AI) requires the parts to be ordered such that `InlineData` precedes the `FunctionResponse`. If the order is reversed, the API parser silently strips or invalidates the entire user message containing the response. This causes the preceding model turn to appear unanswered, triggering an immediate `Error 400: Requests ending with a model turn are not supported`.

This PR reverses the appending order in `AssembleResponse` so that `InlineData` correctly precedes `FunctionResponse`.

### Changes
*   Reversed appending order in `AssembleResponse` in `internal/agent/executor/executor.go`
*   Updated `tests/integration/agent/executor/executor_table_test.go` to assert the new part ordering
*   Extracted mock test functions in `tests/integration/agent/regression_test.go` to adhere to cyclomatic complexity thresholds.

### DIFF
--- a/internal/agent/executor/executor.go
+++ b/internal/agent/executor/executor.go
@@ -368,7 +368,6 @@ func (e *Dispatcher) extractFunctionCalls(respContent *llm.Content) []*llm.Funct
 func (e *Dispatcher) AssembleResponse(calls []*llm.FunctionCall, results []tools.ToolResult) *llm.Content {
 	var responseParts []*llm.Part
 	for i, tr := range results {
-		responseParts = append(responseParts, e.strategy.Format(calls[i], tr))
 		for _, b := range tr.BinaryData {
 			responseParts = append(responseParts, &llm.Part{
 				InlineData: &llm.Blob{
@@ -377,6 +376,7 @@ func (e *Dispatcher) AssembleResponse(calls []*llm.FunctionCall, results []tools
 				},
 			})
 		}
+		responseParts = append(responseParts, e.strategy.Format(calls[i], tr))
 	}
 	return &llm.Content{
 		Role:  "user",
--- a/tests/integration/agent/executor/executor_table_test.go
+++ b/tests/integration/agent/executor/executor_table_test.go
@@ -544,8 +544,8 @@ func TestDispatcher_AssembleResponse_Binary(t *testing.T) {
 		if len(content.Parts) != 2 {
 			t.Fatalf("Got %d parts, want 2", len(content.Parts))
 		}
-		assertFunctionResponse(t, content.Parts[0], "Here is your image")
-		assertInlineData(t, content.Parts[1], "image/png", []byte("blob"))
+		assertInlineData(t, content.Parts[0], "image/png", []byte("blob"))
+		assertFunctionResponse(t, content.Parts[1], "Here is your image")
 	})
 
 	t.Run("Multiple Binary Parts", func(t *testing.T) {
@@ -561,8 +561,8 @@ func TestDispatcher_AssembleResponse_Binary(t *testing.T) {
 		if len(content.Parts) != 3 {
 			t.Fatalf("Got %d parts, want 3", len(content.Parts))
 		}
-		assertInlineData(t, content.Parts[1], "application/pdf", []byte{1, 2, 3})
-		assertInlineData(t, content.Parts[2], "text/plain", []byte{4, 5, 6})
+		assertInlineData(t, content.Parts[0], "application/pdf", []byte{1, 2, 3})
+		assertInlineData(t, content.Parts[1], "text/plain", []byte{4, 5, 6})
 	})
 
 	t.Run("Multi-blob Interleaving and Ordering", func(t *testing.T) {
@@ -576,9 +576,9 @@ func TestDispatcher_AssembleResponse_Binary(t *testing.T) {
 			},
 		}}
 		content := e.AssembleResponse(calls, results)
-		assertFunctionResponse(t, content.Parts[0], "Captured 2 photos")
-		assertInlineData(t, content.Parts[1], "image/jpeg", []byte("photo1"))
-		assertInlineData(t, content.Parts[2], "image/jpeg", []byte("photo2"))
+		assertInlineData(t, content.Parts[0], "image/jpeg", []byte("photo1"))
+		assertInlineData(t, content.Parts[1], "image/jpeg", []byte("photo2"))
+		assertFunctionResponse(t, content.Parts[2], "Captured 2 photos")
 	})
 
 	t.Run("Binary Data with No Text", func(t *testing.T) {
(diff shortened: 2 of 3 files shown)


gosharplite:
## Architecture Review — PR #1442 (fix #1441)

**Verdict: directionally correct and safe to merge.** Verified on `4e7df468` (3 files, 58+/40−): the reorder reaches the Gemini wire as intended and is harmless to the OpenAI and Anthropic transports. The open items are completeness (parallel multi-tool batches) and enforcement-layer durability (persisted history), not correctness of the single-tool path.

### Verified safe — no regression risk

| Claim | Result |
|---|---|
| Reorder reaches the Gemini wire | ✅ `gemini/adapter.go` `toSDKContent`/`toSDKPart` map parts in order → `[inlineData, functionResponse]` on the wire |
| OpenAI transport unaffected | ✅ Order-independent: `partitionParts` emits FunctionResponse parts as separate `tool`-role messages; media+text become a separate `user` message with its own documented **media-first** ordering (`buildMessageContent`) — consistent with this PR's direction |
| Anthropic transport unaffected | ✅ `partToContentBlock` has **no `InlineData` case** — media is silently dropped there (pre-existing gap, unchanged by this PR) |
| `Content.Validate()` imposes no ordering | ✅ Only strips nil parts |
| No other mixed-media assembly site | ✅ `middleware.go` (loop feedback) and `context/transformers.go` (HistoryRepairer) build FR-only / `tool`-role contents; `AssembleResponse` is the only path mixing `InlineData` + `FunctionResponse` |
| Changed tests pass | ✅ `TestDispatcher_AssembleResponse_Binary` (4/4), `TestAgent_MultiModalFlow` |
| CC policy (≤10) | ✅ Extracted helpers measure CC 4–6; `AssembleResponse` structurally unchanged |
| modelith-drift | ✅ `*_test.go` excluded from the drift diff scan — new helpers don't trip it |
| Regression mock refactor | ✅ Functionally required, not cosmetic: the old `Parts[0].FunctionResponse` check breaks under the new order (InlineData now at index 0); `hasFuncResp` scanning all parts is the correct adaptation — good catch |

### [REFACTOR] Incomplete for the parallel multi-tool batch case

The new loop groups per-result: `[blob₀, FR₀, blob₁, FR₁]`. The issue's own diagnosis is "a `user` turn where a **FunctionResponse precedes a media block**" — under that reading, `FR₀` still precedes `blob₁`, so a batch where two tools both return media (the executor's parallel-batch path, `maxConcurrentTools`) still triggers the quirk. No test covers two results each carrying `BinaryData`; the "Multi-blob Interleaving and Ordering" subtest is single-result.

Suggested fix — two-pass assembly guarantees no FR-before-InlineData adjacency under any reading of the quirk:

```go
for _, tr := range results { /* pass 1: all InlineData */ }
for i, tr := range results { /* pass 2: all FunctionResponse */ }
```

Or at minimum, add a multi-result mixed-media test and verify empirically against the live API.

### [TECHNICAL DEBT] Provider quirk enforced in the provider-agnostic executor, with no rationale comment

1. **No comment** at the reorder in `executor.go` explaining *why* InlineData must precede FunctionResponse (cite #1441). This repo's convention is to document non-obvious constraints at the site (cf. `buildMessageContent`'s "media-first ordering" comment, the architect-acceptance comments throughout). Without it, a future cleanup silently reverts and the table tests are the only guard.
2. **Layer placement**: `AssembleResponse` is shared across providers; a Gemini parser quirk is adapter knowledge. Per this repo's pattern (ADR-055/060/067 — adapter owns the wire protocol), the durable enforcement point is the Gemini transport (`toSDKContent`): normalize user turns there so InlineData precedes FunctionResponse. That also **self-heals persisted history**: the assembled user turn is written to `history.jsonl` and re-sent verbatim on later turns — sessions poisoned *before* this fix would still 400 after upgrading with the executor-only change, whereas transport-level normalization fixes both fresh and reloaded turns. Keep the executor change as the canonical default (it matches OpenAI's own media-first convention), but don't rely on it as the guarantee.

### [REFACTOR] Test completeness gaps

- `Multiple Binary Parts` subtest asserts only the two blob positions — it no longer asserts `Parts[2]` is the FunctionResponse (the other three subtests assert both ends).
- No Gemini-adapter-level test pins the wire order (`gemini/adapter_test.go`) — the quirk is only indirectly covered by the executor table tests.
- `TestAgent_MultiModalFlow`'s mock is now order-insensitive, so it no longer detects ordering regressions; the executor table tests are the sole guard.

### Process nits

- Issue #1441 says "The fix has been applied to the `dev` branch" — stale; this PR is open/unmerged, targeting `dev`.
- The PR description attributes the regression-test extraction to CC limits; the functional reason is the mock's `Parts[0]` assumption. The extraction is needed — for correctness, not complexity.

### Recommendation

Merge as-is (fixes the reported single-tool case, harms nothing), then follow up with: (1) two-pass assembly or a multi-result mixed-media live test; (2) a rationale comment at the reorder citing #1441; (3) ideally move authoritative normalization into the Gemini transport so persisted/reloaded history self-heals.

gosharplite:
Follow-up items consolidated into [#1443](https://github.com/gosharplite/tell-me-go/issues/1443) (multi-result batch ordering, transport-level enforcement/self-healing, and test gaps).

Issue #1448, opened 2026-08-28
feat(llm): enable vision (image input) for OpenAI gpt-5.x models via the Responses API — implementation spec (supersedes #1447)


gosharplite:
## Summary

Enable vision (image input) for OpenAI gpt-5.x models via the **Responses API**. `tell-me-go` currently drops image parts for every OpenAI-family provider: `ResolveCapabilities` never sets `SupportsVision` for GPT models, and the transport + Responses sink discard images. This issue is the **implementation spec** — it supersedes #1447 and incorporates the findings of the [Architect × Griller round](https://github.com/gosharplite/tell-me-go/issues/1447#issuecomment-5449459553) (10 questions, every claim verified against the live tree) plus the [owner ratification of D1/D2/D3](https://github.com/gosharplite/tell-me-go/issues/1447#issuecomment-5449555045).

Full grill transcript (public gist): https://gist.github.com/gosharplite/d316168a24740a29244ea8f76531d471

Precedent: ADR-070 (DeepSeek vision, #1429) — `docs/adr/2026-09-deepseek-vision-capability.md`.

## Locked decisions (binding — do not re-litigate)

| | Decision | Detail |
|---|---|---|
| **D1** | **Full-history sticky `hasImage` routing** (deviation from #1447's per-turn wording) | Once an image enters the retained history window, every subsequent turn routes to `/responses` (re-serializing prior `input_image` blocks) until summarisation evicts the image turns, then flips back. Rationale: both sinks re-serialize the *entire* retained history, so per-turn routing would make routing a function of the current message while serialization is a function of full history — the two-gate disagreement D1 exists to kill. ADR records the deviation; criterion reworded (see Acceptance Criteria). |
| **D2** | **All gpt-5+ via `isGpt5OrNewer`** (major ≥ 5) | gpt-5.0–5.3 → vision via existing Chat Completions `image_url` path; gpt-5.4+ → vision via Responses API. Anti-allowlist, consistent with ADR-070 Decision 1. Accepted: 5.0–5.3 tool+image on Chat Completions is untested — smoke test at implementation time. |
[...]

PR #1454, opened 2026-08-28
fix(analysis): protect errors.Wrapper.Unwrap from dead-code false positives

gosharplite:
## Summary

Fixes a dead-code analyzer false positive: `Unwrap() error` methods were flagged as DEAD because the method is consumed **structurally** by Go's error-unwrapping protocol (`errors.Is` / `errors.As` / `fmt.Errorf("%w")`) — the call sites live inside the stdlib, invisible to AST reference analysis.

`Unwrap` was missing from the analyzer's well-known-contract map (`internal/tools/analysis/wellknown.go`), which already protects the same class of protocol methods (`Error`, `String`, `MarshalJSON`, `Read`, `Write`, ...). This surfaced after #1448 landed `MediaSizeError.Unwrap` (`internal/domain/llm`):

```
$ make dead-code
[DEAD] internal/domain/llm.(MediaSizeError).Unwrap (Method) — No references found within the module
       (including interfaces/tests). [WARNING: Text search found potential cross-package usage.
       Verify this is not a false positive due to structural typing.]
```

The tool's own warning called it — it was a structural-typing false positive (the method is load-bearing: pinned by `errors.Is(err, llm.ErrTerminal)` tests and documented in ADR-073).

## Change

- `internal/tools/analysis/wellknown.go` — add `errors.Wrapper.Unwrap` (`func() error`) to `wellKnownContractMethods` with rationale comment; probe-reference comment updated.
- `internal/tools/analysis/dead_code_contracts_test.go` — positive row (`Unwrap()error` → true) + two wrong-signature negatives (`Unwrap()string`, `Unwrap()(error,error)` → false).

This protects **all** `Unwrap() error` methods (current: `MediaSizeError.Unwrap`, `agentError.Unwrap`; future error types) — a root-cause fix, not a per-type workaround.

## Verification

| Check | Result |
|---|---|
| `go test ./internal/tools/analysis/` | ✅ PASS |
| `make dead-code` | ✅ **"No dead code found."** (was: DEAD `MediaSizeError.Unwrap`) |
| `gofmt -l` | ✅ clean |
| `go vet ./internal/tools/analysis/` | ✅ clean |
| `make verify-nonfix-catalog` | ✅ PASS (arch-tagged) |

## Notes

- No feature-code changes; the analyzer's contract map is extended, matching its documented purpose ("stdlib single-method interfaces that may be satisfied structurally").
- No catalog edits (the dead-code gate is report-only; no `AcceptanceClass` fits a load-bearing protocol method).

### DIFF
--- a/internal/tools/analysis/dead_code_contracts_test.go
+++ b/internal/tools/analysis/dead_code_contracts_test.go
@@ -83,6 +83,8 @@ func TestIsWellKnownContract_Table(t *testing.T) {
 		// --- Positive: legacy contracts ---
 		{"Error()string", "Error", nil, []string{"string"}, true},
 		{"String()string", "String", nil, []string{"string"}, true},
+		// --- Positive: errors.Wrapper.Unwrap (errors.Is / errors.As / %w) ---
+		{"Unwrap()error", "Unwrap", nil, []string{"error"}, true},
 		// --- Positive: io contracts (basic-kind-only) ---
 		{"Read([]byte)(int,error)", "Read", []string{"[]byte"}, []string{"int", "error"}, true},
 		{"Write([]byte)(int,error)", "Write", []string{"[]byte"}, []string{"int", "error"}, true},
@@ -101,6 +103,8 @@ func TestIsWellKnownContract_Table(t *testing.T) {
 
 		// --- Negative: matching name, wrong signature ---
 		{"Write(string)", "Write", []string{"string"}, nil, false},
+		{"Unwrap()string", "Unwrap", nil, []string{"string"}, false},
+		{"Unwrap()(error,error)", "Unwrap", nil, []string{"error", "error"}, false},
 		{"Close()(error,error)", "Close", nil, []string{"error", "error"}, false},
 		{"Read()", "Read", nil, nil, false},
 		{"MarshalJSON([]byte)error", "MarshalJSON", []string{"[]byte"}, []string{"error"}, false},
--- a/internal/tools/analysis/wellknown.go
+++ b/internal/tools/analysis/wellknown.go
@@ -29,6 +29,7 @@ import (
 //	encoding.TextUnmarshaler.UnmarshalText params=[[]byte] results=[error]
 //	encoding.BinaryMarshaler.MarshalBinary     params=[] results=[[]byte error]
 //	encoding.BinaryUnmarshaler.UnmarshalBinary params=[[]byte] results=[error]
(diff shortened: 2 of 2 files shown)


gosharplite:
### Architectural & Code Review: PR #1454

**Role**: Principal Software Architect (Read-Only)  
**Target**: `fix/dead-code-unwrap-contract` → `dev`  
**Verdict**: **[APPROVE / LGTM]**

---

### 1. Architectural Assessment & Root-Cause Resolution

This PR resolves a static analysis false positive at the systemic tool level rather than applying ad-hoc workarounds:

1. **Systemic Contract Protection vs. Per-Type Workarounds**:
   - The Go standard library error unwrapping mechanism (`errors.Is`, `errors.As`, `errors.Unwrap`, and `fmt.Errorf("%w")`) relies on structural type assertions (`interface{ Unwrap() error }` / `errors.Wrapper`). Because invocation sites reside inside stdlib runtime calls rather than explicit intra-module references, AST call-graph traversers cannot resolve direct call edges.
   - Adding `"Unwrap": {{params: nil, results: []string{"error"}}}` to `wellKnownContractMethods` in `internal/tools/analysis/wellknown.go` extends the existing well-known protocol dictionary (alongside `Error() string`, `String() string`, `MarshalJSON() ([]byte, error)`, etc.) in a disciplined, declarative manner.
   - This prevents false-positive alerts across all existing and future domain error wrappers (e.g., `MediaSizeError`, `agentError`) without requiring artificial mock call sites or catalog pollution.

2. **Interface Contract Precision**:
   - Exact signature enforcement (`params: nil`, `results: []string{"error"}`) ensures that only genuine Go error unwrappers are exempted. Non-conforming signatures (e.g., `Unwrap() string` or `Unwrap() (error, error)`) continue to be analyzed and flagged if unreferenced.

---

### 2. Test Coverage & Negative Boundary Verification

- **Positive Case**: Added `{"Unwrap()error", "Unwrap", nil, []string{"error"}, true}` to `TestIsWellKnownContract_Table`.
- **Negative Boundary Cases**: Added `{"Unwrap()string", ... false}` and `{"Unwrap()(error,error)", ... false}` to assert that signature validation is strictly enforced.
- **Verification**: `make dead-code` now exits cleanly with `"No dead code found."` and zero false positives.

---

### 3. Recommendation

Clean, surgical, and idiomatic fix that maintains analysis precision. Ready to merge into `dev`.

